PR #17, opened 2018-03-05
Update style server so tiles can be rendered from fs

karissa:
Then, you have to download the mapbox styles to `homedir/mapeo/style` 

Try doing this command:  
```
$ npm install -g tile-dl
$ tile-dl -t 'https://c.tiles.mapbox.com/v4/digitalglobe.0a8e44ba/{z}/{x}/{y}.png?access_token=pk.eyJ1IjoiZGlnaXRhbGdsb2JlIiwiYSI6ImNqOGRmNW9qZjBudmgzMnA1a294OGRtNm8ifQ.06mo-nDisy4KmqjYxEVwQw' --lon=-77.43049196 --lat=0.1236344282 --radius 0.1 -z 12 -o {z}/{y}/{x}.vector.pbf
```
which should download the area of Sinangoe, using @noffle's mapbox key. It will store pbf files in a directory called 12, which then should be moved to `~/mapeo/style/tiles/composite`on the filesystem.

Mapbox can interpret the `{host}` variable from style.json to find these tile files. It is written on the filesystem like this:
screen shot 2018-03-05 at 6 35 46 pm

It works by setting the tile server to `http://localhost:9081/`. This is how the JSON file is served to the client:
screen shot 2018-03-05 at 6 35 27 pm


staltz:
Looks like the correct approach @karissa ! :)
To Cindy and Helena: `homedir/mapeo/style` refers to the directory in the Android OS, and `homedir` is the app's directory, it's usually `/data/data/com.mapeomobile/files/` but the tricky thing is that you don't have direct access to this in the terminal (unless the phone is rooted). So the way you can get the files into that folder is by putting them all in `src/rnnodeapp` (you probably want to git ignore them), then they'll all be bundled together with the background process. That should be enough as a quick hack for the demo.

Another option, which makes it less hacky and more production ready is to run `tile-dl` not in the terminal, but in Node.js. This way, when the background process starts, it will start downloading the tiles from Mapbox, and it'll keep it later for offline use.

Another option, similar to the above, is to share tile images as a Dat! Then the tiles can be shared through LAN networks, for instance, without the internet.

But for now, bundling it as statics will do.

src/rnnodeapp/serve_styles.js
```diff
@@ -1,42 +1,86 @@
...
+    const files = fs.readdirSync(root);
+    res.statusCode = 200;
+    const string = files.map(f => JSON.stringify(f)).join(',');
+    console.log(root, files, string);
+    res.write(string);
+    res.end();
+    return null;
+  };
+
+  const tiles = (req, res) => {
```
karissa: Curious why this part is necessary and the ecstatic call at the bottom didn't work?

PR #20, opened 2018-04-15
Feature/april w3 helena

howdyitshelena:
Editor UI MVP


src/components/Views/Categories/Categories.js
```diff
@@ -26,7 +32,7 @@ export type StateProps = {
 
 export type DispatchProps = {
   listCategories: () => void,
-  updateObservation: (obs: Observation) => void,
+  updateObservation: (obs: Object) => void,
```
czengg: We might want to update the observation type if that's the issue

PR #22, opened 2018-04-17
Observation Manager UI MVP

howdyitshelena:
- updated category names to be singular
- fixed creating an observation from the camera
- fixed error when rendering detail on new observations


src/components/Views/ObservationsView/ObservationsView.js
```diff
@@ -73,80 +136,141 @@ I18n.translations = {
...
+                <Text
+                  style={{ fontWeight: '700', fontSize: 12, color: 'black' }}
+                >
+                  {I18n.t('observations.all_observations')}
+                </Text>
+                <Text style={{ fontSize: 12 }}>
+                  {I18n.t('observations.view_by')}
+                </Text>
+              </View>
+              {this.state.showSyncTip ? (
```
czengg: maybe just `!!this.state.showSyncTip && [component]`?

howdyitshelena: oooh good call! shall change that

PR #25, opened 2018-04-30
Remove mocked observations, fix ordering for ObservationsView

howdyitshelena:
ObservationsView doesn't render again when the drawer opens, so that's why the ordering wasn't getting sorted


src/components/Views/ObservationsView/ObservationsView.js
```diff
@@ -19,6 +20,7 @@ import ObservationCell from './ObservationCell';
 import ObservationHeader from './ObservationHeader';
 
 export type StateProps = {
+  drawerIsOpen: boolean,
```
czengg: let's called this `drawerOpened`?

src/components/Views/TabBarNavigation/index.js
```diff
@@ -3,9 +3,15 @@ import { connect } from 'react-redux';
...
 
 const mapStateToProps = state => ({
-    showSavedModal: state.app.modals.saved
-  });
+  showSavedModal: state.app.modals.saved
+});
 
-export default connect(mapStateToProps)(TabBarNavigation);
+const mapDispatchToProps = dispatch => ({
+  onDrawerClose: () => dispatch(drawerClose('opened')),
```
czengg: let's call this 'observations' instead of 'opened'

src/components/Views/ObservationsView/ObservationsView.js
howdyitshelena: 👍

src/components/Views/TabBarNavigation/index.js
howdyitshelena: 👍

PR #27, opened 2018-05-01
Run unit tests on Travis

howdyitshelena:
Have unit tests pass on Travis


.travis.yml
```diff
@@ -4,14 +4,6 @@ sudo: required
 dist: trusty
 group: edge
 
-env:
-  global:
-    - ANDROID_HOME=/usr/local/android-sdk-26.0.1
```
czengg: We don’t need these constants?

PR #32, opened 2018-05-05
Update observation detail with editing button

howdyitshelena:


src/components/Views/CameraMainView/CameraMainView.js
```diff
@@ -80,6 +82,27 @@ class CameraMainView extends React.Component<
     loading: false
   };
 
+  componentDidMount() {
```
czengg: probably want to do
`async componentDidMount() { await PermissionsAndroid.request(...) }`

src/components/Views/CameraMainView/CameraMainView.js
```diff
@@ -107,8 +130,8 @@ class CameraMainView extends React.Component<
         this.setState({ loading: true });
 
         const data = await this.camera.takePictureAsync(options);
-
         this.setState({ loading: false });
+        CameraRoll.saveToCameraRoll(data.uri);
```
czengg: Hmmm we might want to put this in a util function called `saveToCameraRoll` that requests permissions if not granted, saves to camera roll, and deletes the photo from app cache. This would prevent dupes

src/components/Views/CameraView/CameraView.js
```diff
@@ -100,6 +101,7 @@ class CameraView extends React.Component<
         this.setState({ loading: true });
         const data = await this.camera.takePictureAsync(options);
         this.setState({ loading: false });
+        CameraRoll.saveToCameraRoll(data.uri);
```
czengg: same comment as above (re: util function)

czengg:
We might also want to put this async call into an epic. It'll take care of async stuff for us with Observables. So workflow would be dispatch(saveToCameraRoll(uri)) action, goes through reducer (presumably does nothing), goes through epic that calls util. This would also allow us to eventually move the loading/saving into state so we could have loading indicators and failed to save indicators.

PR #33, opened 2018-05-05
Model Defaults and Bugfix

czengg:


src/types/observation.js
```diff
@@ -19,5 +19,4 @@ export interface Observation {
   observedBy: string;
   media: ObservationMedia[];
   icon: any;
-  categoryId?: string;
```
howdyitshelena: Think we actually need this for the Category name in ObservationCell and ObservationDetailView

howdyitshelena:
Nice job :D

PR #37, opened 2018-05-15
Feature/may w3 helena

howdyitshelena:
Update Observation Details with fields


src/components/Views/ObservationDetailView/ObservationDetailView.js
```diff
@@ -236,28 +250,57 @@ class ObservationDetailView extends React.Component<
       goBack
     } = this.props;
     const keyExtractor = item => item.source.toString();
+    let mediaTitle = null;
     let mediaText = '';
     const thereIsMedia =
```
czengg: Maybe you just want to do numOfMedia and have the length then calculate from there so you don't have to keep checking for existence of selectedObservation and media

howdyitshelena: Will do!

PR #39, opened 2018-06-01
Bugs/may w3 helena

howdyitshelena:


src/components/Views/ObservationEditor/ObservationEditor.js
```diff
@@ -492,7 +490,7 @@ class ObservationEditor extends React.Component<
             backgroundColor: VERY_LIGHT_BLUE
           }}
         >
-          <TouchableOpacity onPress={goBack}>
+          <TouchableOpacity onPress={goToCategories}>
```
czengg: This will mess up the navigation stack. We should be using back. Or else we'll have a bunch of unnecessary mounted components.

howdyitshelena: Ahyaa just if a user navigates to the editor from the details view, then it'll just go back to the details view

czengg: We should reset the navigation stack after navigating back to the editor view?

czengg: Maybe Detail view should be a drawer that slides in or a subview instead?

PR #40, opened 2018-06-04
Remove CameraMainView and other unneeded views

howdyitshelena:
- To fix the camera issue, workaround was to combine CameraMainView into CameraView since it seems react-native-camera can only have one instance of the camera
- TabBarNavigation renamed -> MapView since it's not a TabNavigator anymore
- MapView renamed -> Map, and under new MapView
- Moved SavedModal under Base
- Removed Position (we're not using this anymore)


src/components/Views/CameraView/CameraView.js
```diff
@@ -87,12 +109,18 @@ class CameraView extends React.Component<
   }
 
   camera: RNCamera;
+  rightDrawer: Drawer;
 
   takePicture = async () => {
     const {
-      updateObservation,
+      createObservation,
+      fromEditor,
```
czengg: I think you can just check if a `selectedObservation` exists, if it does then create the observation, otherwise update the selectedObservation

src/components/Views/CameraView/CameraView.js
```diff
@@ -101,87 +129,239 @@ class CameraView extends React.Component<
...
+        } else {
+          goToCategories();
+          const initialObservation = applyObservationDefaults({
+            id: size(observations) + 1,
+            createdFrom: 'camera'
           });
-          goToObservationEditor();
+          createObservation(initialObservation);
+
+          navigator.geolocation.getCurrentPosition(
```
czengg: We shouldn't have to get current position. This is set in the initial create flow through ducks.

src/components/Views/CameraView/CameraView.js
```diff
@@ -101,87 +129,239 @@ class CameraView extends React.Component<
...
+  openRightDrawer = () => {
+    this.rightDrawer.open();
+  };
+
+  handleRightDrawerRef = (ref: Drawer) => {
+    this.rightDrawer = ref;
+  };
+
+  goToMapView = () => {
+    this.props.navigation.navigate({ routeName: 'MapView' });
```
czengg: let's do this through a dispatchProp. I'd prefer if this goes through our actions process so it's synchronous.

src/components/Views/CameraView/CameraView.js
```diff
@@ -101,87 +129,239 @@ class CameraView extends React.Component<
...
+              >
+                <ActivityIndicator />
+                <Text>{I18n.t('saving_image')}</Text>
+              </View>
+            </View>
+          )}
+        </View>
+      );
+    }
+    return (
```
czengg: let's put a new line above this

src/components/Views/CameraView/CameraView.js
```diff
@@ -101,87 +129,239 @@ class CameraView extends React.Component<
...
+            position: 'absolute',
+            top: 0,
+            left: 0,
+            right: 0,
+            zIndex: 5
+          }}
+        />
+        {showSavedModal && <SavedModal />}
+        <View style={{ flex: 1, flexDirection: 'column' }}>
+          <RNCamera
```
czengg: Seems like some of this code is redundant. You might want to put it in a separate component or function `renderCamera`

src/components/Views/CameraView/index.js
```diff
@@ -2,22 +2,38 @@
...
+import { drawerClose, drawerOpen } from '../../../ducks/drawers';
 
-function mapStateToProps(state: StoreState): StateProps {
+function mapStateToProps(state: StoreState, ownProps: Props): StateProps {
   return {
-    selectedObservation: state.app.selectedObservation
+    observations: state.app.observations,
+    selectedObservation: state.app.selectedObservation,
+    showSavedModal: state.app.modals.saved,
+    fromEditor:
```
czengg: Probably don't need this, see above.

src/types/observation.js
```diff
@@ -23,4 +23,5 @@ export interface Observation {
   icon: any;
   categoryId: string;
   fields: Field[];
+  createdFrom: string;
```
czengg: It may be worth it to put this just in the AppState instead of on the observation model and then we can set it as undefined when we set selectedObservation as undefined. That way we keep the observation model pure and reflective of the API.

src/components/Views/CameraView/CameraView.js
howdyitshelena: 👍

src/components/Views/CameraView/CameraView.js
howdyitshelena: 👍

src/components/Views/CameraView/CameraView.js
howdyitshelena: Ya will refactor a bit

src/types/observation.js
howdyitshelena: Ahh true

PR #59, opened 2018-06-25
Remove NavigationService

howdyitshelena:


src/components/Views/ManualGPS/ManualGPS.js
```diff
@@ -761,4 +769,4 @@ class ManualGPS extends React.PureComponent<StateProps & DispatchProps, State> {
   }
 }
 
-export default ManualGPS;
+export default withNavigation(ManualGPS);
```
czengg: Not sure we even need this anymore! The new upgrade of react-navigation auto includes navigation as a prop so we don't need to do withNavigation!

src/components/Views/MapView/Map/Map.js
```diff
@@ -37,15 +37,14 @@ export type DispatchProps = {
...
-  goToCategories: () => void,
-  goToObservationDetail: () => void,
   selectObservation: (observation: Observation) => void,
   updateObservationSource: () => void,
   listStyles: () => void
 };
 
 type Props = {
-  isFocused: boolean
+  isFocused: boolean,
```
czengg: Do you mind switching this out too? New react-navigation update actually includes `isFocused` as a function on the `navigation` prop. So checking for isFocused would be `navigation.isFocused()` which returns a bool. More info on it here: https://reactnavigation.org/docs/en/navigation-prop.html

src/components/Views/ManualGPS/ManualGPS.js
howdyitshelena: ooo fancy, okay!

src/components/Views/MapView/Map/Map.js
howdyitshelena: 👍 alrighty

PR #62, opened 2018-06-25
Fix saving from editor

howdyitshelena:


src/components/Views/ObservationEditor/ObservationEditor.js
```diff
@@ -379,6 +379,10 @@ class ObservationEditor extends React.Component<
         navigation.navigate({
```
czengg: instead of all this code, since it seems like it's doing the same thing, you could probably just call `this.handleSave`

howdyitshelena: oooh smaht!

src/components/Views/SyncView/SyncView.js
```diff
@@ -134,25 +134,53 @@ class SyncView extends React.Component<
...
-      toggleDeviceSelect(item);
+      const syncInProgress =
+        item.syncStatus === 'requested' || item.syncStatus === 'syncing';
+      const startSync = () => {
+        updateDeviceSync({
+          ...item,
+          selected: true,
+          syncStatus: 'requested'
+        });
+        setTimeout(() => {
```
czengg: We don't want these timeouts

czengg: We should actually hook into the progress from mapeo-server

PR #64, opened 2018-06-26
Add sync api

howdyitshelena:


src/components/Views/SyncView/SyncView.js
```diff
@@ -142,20 +145,20 @@ class SyncView extends React.Component<
         updateDeviceSync({
           ...item,
           selected: true,
-          syncStatus: 'requested'
+          syncStatus: 'replication-started'
         });
         setTimeout(() => {
```
czengg: we probably want to remove these timeouts

src/ducks/devices.js
```diff
@@ -4,6 +4,20 @@ import update from 'immutability-helper';
 import { keyBy } from 'lodash';
 import { create } from '../lib/redux';
 
+export const {
```
czengg: Let's put this in `ducks/sync` and make the actions `SYNC_ANNOUNCE`

src/ducks/devices.js
```diff
@@ -4,6 +4,20 @@ import update from 'immutability-helper';
...
+  type: ANNOUNCE_SYNC,
+  action: announceSync,
+  reducer: announceSyncReducer
+} = create('ANNOUNCE_SYNC', {
+  success: (state, action) => {
+    const newState = update(state, {
+      $toggle: ['syncAnnounced']
```
czengg: might not want to put it in the state tree. I think we want to announce every time the view is in focus so we should do it when the isFocused function goes from false to true

src/components/Views/SyncView/SyncView.js
```diff
@@ -57,8 +59,9 @@ class SyncView extends React.Component<
   state = { wifi: false };
 
   componentDidMount() {
-    const { listDevices } = this.props;
+    const { announceSync, listDevices } = this.props;
 
+    announceSync();
     listDevices();
```
howdyitshelena: I'm guessing we want to listDevices() after announceSync() is successful ?

src/epics/devices.js
```diff
@@ -47,26 +54,35 @@ const initialDevices = [
 
 // const initialDevices = [];
 
+export const announceSyncEpic = (action$: ActionsObservable<any>) =>
+  action$
+    .ofType(ANNOUNCE_SYNC)
+    .filter(action => action.status === 'Start')
+    .flatMap(() => Sync.announce());
```
czengg: hmmm so this will break cause you're not passing back an action. You have a success reducer for announceSynce so you might want to just do `Sync.announce().map((resp) => syncAnnounce(action.meta, resp))`

PR #65, opened 2018-06-27
Bugs/june w4 helena

howdyitshelena:
- Only show cancel modal when changes have been made in editor
- Don't allow deletion of existing photos in editor
- Fix camera navigation when creating an observation from camera


src/components/Views/ObservationEditor/ObservationEditor.js
```diff
@@ -496,6 +516,76 @@ class ObservationEditor extends React.Component<
     }
   };
 
+  observationExists = () => {
```
czengg: bleh, since we're doing this might be easier to just pass in the observations as is and then turn it into an array in the component. This seems expensive

src/components/Views/ObservationEditor/ObservationEditor.js
```diff
@@ -496,6 +516,76 @@ class ObservationEditor extends React.Component<
...
+
+    return false;
+  };
+
+  handleCancel = () => {
+    const { navigation, observationSource, showCancelModal } = this.props;
+    if (this.hasEdits()) {
+      showCancelModal();
+    } else if (observationSource === 'map') {
+      navigation.navigate({
```
czengg: Won't this mess up the back button on the android?

src/components/Views/CameraView/CameraView.js
```diff
@@ -204,7 +204,11 @@ class CameraView extends React.Component<
         {fromEditor && (
           <TouchableOpacity
             style={styles.cancelButton}
-            onPress={() => this.props.navigation.goBack()}
+            onPress={() =>
```
czengg: Won't this mess up the back button/stack on the navigation?

howdyitshelena: For some reason, it's not going back to the editor from goBack hmm

czengg: What happens when you goBack?

src/components/Views/ObservationEditor/ObservationEditor.js
howdyitshelena: Hmm how so?

czengg: Oh never mind, I thought this was the cancel to go back button. This should be fine

src/components/Views/CameraView/CameraView.js
howdyitshelena: It ends up going to the camera view. So the flow here is:
1. create observation from camera
2. select category
3. in editor, open camera
4. hit cancel
5. currently, it takes you back to the camera view instead of to the editor

src/components/Views/ObservationEditor/ObservationEditor.js
howdyitshelena: 👍

PR #72, opened 2018-07-01
fix paths for making side-loading styles easier

karissa:


src/rnnodeapp/index.js
```diff
@@ -28,10 +28,16 @@ const styles = require('mapeo-styles');
...
+// NOTE: in the future, we might want separate private keys
+// for each user, so let's just have a 'default' user for now so it'll be
+// easier to migrate later
+const USER_ID = 'default' 
+const USER_PATH = path.join(os.homedir(), 'mapeo', USER_ID);
 const DB_PATH = path.join(USER_PATH, 'db');
 const MEDIA_PATH = path.join(USER_PATH, 'media');
-const STATIC_PATH = path.join(USER_PATH, 'static');
+
+const STATIC_PATH = path.join('/sdcard/Android/data/com.mapeomobile/static');
```
karissa: @staltz is this consistent across android devices?

staltz: Oh I don't believe it is, some devices might not even have an sdcard. The right way of doing this is fetching the path using Java APIs meant for this, passing the path string to the (frontend) JS, then passing that here (the backend). The package `react-native-fs` should help for doing the first part.

karissa: ping @gmaclennan

gmaclennan:
@karissa the Android filesystem is really confusing, and we should be using the Java APIs described here to get the folder: https://developer.android.com/training/data-storage/files#WriteExternalStorage

We define these paths in the `rnnode` process, not react-native, so we can't use `react-native-fs` unfortunately.

Where we want to store the files is called "external storage" by Android, although it is normally actually storage on the device see https://developer.android.com/guide/topics/data/data-storage#filesExternal

The external storage is, I think, always mounted at `/sdcard` even though this is almost never an SD card (yes, super confusing). All the devices I have tested use this path for mounting the "external" shared internal storage.

I think we need to pin a "TODO" here to actually use the Java APIs to get these paths, but I think for a short-term fix using the absolute paths as you have done is OK for now.

staltz:
Correct assessment, Gregor, and to clarify about how to get that path, there isn't a Java-to-`rnnode` communication channel, so two ways I see how to do it: (1) Java passes the path as an environment variable to the node process at startup time, (2) use react-native-fs to communicate Java-to-RN-to-`rnnode`, although that's some extra hoops. The first option is clean enough but requires modifying react-native-node source code. The second option means we don't need to modify any library code but it is a more indirect solution, also adds a lot more code from react-native-fs.

PR #78, opened 2018-07-13
[WIP] Fix GPS location updates

gmaclennan:
Working to fix the way GPS is handled by the app.

- Fix GPS state being over-written by redux-persist rehydrate
- Detect when location services are turned on and off whilst the app is open
- Avoid timeout errors and keep polling GPS until accuracy is good enough


gmaclennan:
Moved todos to #81 and #82, since those are new features. Leaving this PR as a bug-fix for current GPS behaviour. @cindy I think this is ready to merge.

PR #79, opened 2018-07-14
Sync fixes

karissa:
* show hostname instead of ip in list
* show completed when sync complete
* show 'progress' (just that data is being received)
* don't duplicate devices if a device disconnects and then reconnects

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,3 +60,8 @@ Low-level library tests run a benchmarks-only background Node.js process in the 
 ```sh
 npm run benchmark
 ```
+
+## Helpful Resources
+
+- Enzyme Example: https://github.com/airbnb/enzyme#shallow-rendering
+- Jest Snapshot Testing: https://facebook.github.io/jest/docs/en/snapshot-testing.html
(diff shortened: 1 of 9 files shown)


karissa:
@czengg forgot -- this PR should also make sure that the sync screen is re-requesting and refreshing the device list every second, then progress will be visible. It won't be hitting i/o, it's all memory, so should be efficient even though it's polling

karissa:
@czengg also, why is the build failing now? I am not exactly sure why the test won't properly show the device name

czengg:
@karissa snapshots are based off mocks in the mock folder. Likely that's why it's showing the wrong device name..

karissa:
@czengg hmm it says 'name' and i'm using 'device.name' in the view -- and nothing else should have changed, so it's confusing

karissa:
@czengg what are the snap files? do we have to re-generate them?

czengg:
@karissa our tests are snapshot tests so it’s testing that the component is outputting the same thing given mocks. Changing to device.name would cause a mismatch in output. So tests are failing correctly. You should just update it using yarn test -u

czengg:
@karissa yeah...I'm looking into just creating basic ReadMe for setup and tests... Nope, you should be able to run `yarn test` on src directory. You may have been in the `src/rnnodeapp` folder which doesn't have tests?

karissa:
No I was in the root folder. Should I have been in the 'src' folder?

PR #83, opened 2018-07-16
fixes for media sync

karissa:
* implement ._list for fs-blob-store which is required
* add media options for only pushing media from the phone


karissa:
depends upon https://github.com/digidem/mapeo-server/pull/20/

PR #89, opened 2018-07-17
Update observations to go under properties

czengg:
Quick hack fix, observations update endpoint only allows updates to lat, lon, and properties keys, so client will transform observation types when interacting with server


gmaclennan:
See https://github.com/digidem/mapeo-mobile/issues/90

karissa:
Looks like we are going to not do this because we're going to stick with osm models. @czengg can you see if this pull request fixes your problem with attachments? https://github.com/digidem/mapeo-server/pull/21

PR #91, opened 2018-07-18
bugfix: change method for checking app is ready

gmaclennan:


gmaclennan:
@czengg I haven't check this yet, but wanted to push my edits before dinner

PR #96, opened 2018-07-23
Replace react-native-node with nodejs-mobile-react-native

staltz:
The important commit to review is this one cf2f87e27e4faf44a5710db5c82b7c228cfcef29

Otherwise not much has changed in the architecture, perhaps except for two things.

### cwd hack

The backend project now has a loader file which hacks os.homedir and process.cwd:

```js
os.homedir = () => __dirname;
process.cwd = () => __dirname;
```

The cwd affects things like relative paths, if those happen to be used in our stack. Read this issue in noderify: https://github.com/dominictarr/noderify/issues/20 After discussions with Dominic and also the nodejs-mobile folks, the hack above seems the least problematic solution so far. I have tried to convince nodejs-mobile to instead set up the current working dir (`cwd`) in their library, but they want to keep it "honest", as in there is no actual current working directory for the Android app, and they don't want to emulate it because node.js doesn't run in a separate process, it runs in a thread (**unlike** react-native-node).

### Dropped support for x86 Android devices

This is probably just a "nominal" drop of support, because I believe x86 Android devices probably could not have run this app, because the native libraries used were not compiled for that, only for `arm`. Anyway, just to let you know, this was one step needed to avoid the nodejs-mobile builds of native node_modules from erroring when attempting to compile for that architecture.


staltz:
PS: I noticed the merge conflict. I noticed that's because Gregor made a fix for the retry. I managed to easily fix this in src/api/base.js just by using `Observable.defer`. The rationale is that `Observable.from(fetch(url))` can never ever be retried, because it doesn't re-execute `fetch(url)` again, because promises are eager, not lazy. So we need to wrap it with a promise factory, that's the use case for defer, `Observable.defer(() => Observable.from(fetch(url)))`, then the `retryWhen` will work.

gmaclennan:
Closing since these changes are integrated in another branch

PR #97, opened 2018-07-26
[WIP] Allow selection of presets

czengg:


karissa:
Are we going to have to retire this in place of Projects? How does this work relate? @gmaclennan @aldopuicon

gmaclennan:
closing since this is no longer needed, functionality will be implemented by projects

PR #98, opened 2018-07-26
Use 127.0.0.1 to be able to reach server offline

noffle:
127.0.0.1 seems to be able to hit mapeo-server even when wifi and cell data are
not active.


karissa:
i belive you

PR #100, opened 2018-07-27
update observations

czengg:


src/models/observations.js
```diff
@@ -1,19 +1,59 @@
...
-  link: 'link',
   created: new Date(),
   name: '',
   notes: '',
   observedBy: 'Tú',
   attachments: [],
-  icon: null,
-  categoryId: '0',
-  fields: [],
+  categoryId: '',
```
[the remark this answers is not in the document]
czengg: They are stored under the tags field

src/models/observations.js
```diff
@@ -1,19 +1,59 @@
...
+  const keep = { created: true };
+  Object.keys(tags).forEach(property => {
+    if (!keep[property] && tags[property] === defaultObservation[property]) {
+      delete tags[property];
+    }
+  });
+
+  // remove topLevel properties
+  const topLevel = ['lat', 'lon', 'attachments', 'id'];
+  topLevel.forEach(p => delete tags[p]);
```
czengg: This is where we separate the toplevel properties from those we're keeping in tags. The property in tags are essentially everything that isn't top level and in the observation model, ie: created, categoryId, fields, etc.

czengg:
@gmaclennan this should address bringing the answers for presets up to observation.tags

src/models/observations.js
```diff
@@ -1,19 +1,72 @@
...
+
+  return {
+    lat: lat || 0,
+    lon: lon || 0,
+    version,
+    attachments: (attachments || []).map(id => ({ id })),
+    type: 'observation',
+    device_id: '1',
```
karissa: What is device_id?

PR #106, opened 2018-08-17
unnannounce device sync when sync view closes

karissa:
depends upon mapeo-server bump


src/components/Views/SyncView/SyncView.js
```diff
@@ -84,10 +85,14 @@ class SyncView extends React.Component<
   }
 
   componentWillUnmount() {
+    const { unannounceSync } = nextProps;
+
     if (this.interval) {
       clearInterval(this.interval);
     }
 
+    unannounceSync();
```
noffle: Is it guaranteed that `componentWillUnmount` will be called when the view closes?

noffle:
Hey @czengg does this look good to you? I've already merged the server dependency for it at https://github.com/digidem/mapeo-server/pull/25.

src/components/Views/SyncView/SyncView.js
czengg: No, it is not guaranteed that the view will unmount. You might need to hook into the willBlur listener on react-navigation though: https://reactnavigation.org/docs/en/navigation-prop.html#addlistener-subscribe-to-updates-to-navigation-lifecycle

czengg:
@noffle one quick comment otherwise looks good

src/components/Views/SyncView/SyncView.js
noffle: @czengg If we are then announcing on `componentDidMount` but unannouncing on `willBlur`, could we get into a state where we can "unblur" without re-mounting and have our announcing be turned off?

I'm feeling a little out of my depth here, and karissa is taking some time off. Looking to follow your lead on this. :)

noffle:
@czengg I can't find your comment. Did you see mine on the unmount/blur stuff?

src/components/Views/SyncView/SyncView.js
czengg: @noffle Sorry just saw this. I can take over this branch if you'd like. We will need to check willFocus as well to trigger an announce. We shouldn't rely on the component lifecycle for this

noffle:
@czengg yes please! I'm out of my depth here :)

czengg:
@gmaclennan This branch is just about ready. It requires the update-observation branch to be merged however, because of the mapeo-server upgrade

PR #110, opened 2018-09-19
Syncfile

karissa:
Initial pass of implementing the 'Sync to USB' button. This definitely needs a close look -- I am a bit out of my depth with the existing ducks/epics model.

Depends upon merging and publishing this PR: https://github.com/digidem/mapeo-sync/pull/4 in a new version of mapeo-server.

### DIFF
--- a/src/api/sync.js
+++ b/src/api/sync.js
@@ -21,10 +21,10 @@ class Sync {
       route: `/sync/start?host=${device.host}&port=${device.port}`
     });
 
-  static toFile = (filename: string) =>
+  static dataset = (dataset: string) =>
     blankRequest({
       method: 'GET',
-      route: `/sync/start?filename=${filename}`
+      route: `/sync/start?dataset=${dataset}`
     });
 }
 
--- a/src/api/sync.js
+++ b/src/api/sync.js
@@ -20,6 +20,12 @@ class Sync {
       method: 'GET',
       route: `/sync/start?host=${device.host}&port=${device.port}`
     });
+
+  static toFile = (filename: string) =>
+    blankRequest({
+      method: 'GET',
+      route: `/sync/start?filename=${filename}`
+    });
 }
 
 export default Sync;
--- a/src/components/Views/SyncView/SyncView.js
+++ b/src/components/Views/SyncView/SyncView.js
@@ -11,6 +11,7 @@ import {
   NetInfo
 } from 'react-native';
 import type { NavigationScreenProp } from 'react-navigation';
+import SyncIcon from 'react-native-vector-icons/MaterialIcons';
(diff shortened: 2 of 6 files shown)


src/components/Views/SyncView/SyncView.js
```diff
@@ -155,6 +156,10 @@ class SyncView extends React.Component<
 
     const keyExtractor = (item, index) => item.id;
 
+    const handleUSBPress = () => {
+      startSync({filename: '/sdcard/Android/data/com.mapeomobile/static/data.syncfile'})
```
czengg: Is this path consistent for all Android phones?

src/epics/devices.js
```diff
@@ -65,11 +65,12 @@ export const syncStartEpic = (action$: ActionsObservable<any>) =>
   action$
     .ofType(SYNC_START)
     .filter(action => action.status === 'Start')
-    .flatMap(action =>
+    .flatMap(action => {
+      if (action.meta.filename) return Sync.toFile(action.meta.filename)
```
czengg: this should be 

```
const syncing = action.meta.filename ? Sync.toFile(action.meta.filename) : Sync.start(action.meta);
return syncing.map(resp => syncStart(meta, resp);
```

src/components/Views/SyncView/SyncView.js
karissa: no :P we definitely want to change this

noffle:
You can use mapeo-sync@3.1.0 now!

src/rnnodeapp/index.js
```diff
@@ -62,7 +62,8 @@ function start() {
   const route = Router(db, media, {
     host: HOST,
     media: { mode: 'push' },
-    staticRoot: STATIC_PATH
+    staticRoot: STATIC_PATH,
+    writeFormat: 'osm-p2p-syncfile'
```
noffle: could we do `sync: { writeFormat: 'osm-p2p-syncfile' }` instead? same on your mapeo-desktop PR -- same way we're namespacing the `media` opts

karissa: sure sounds good!

noffle:
nevermind, I see that we're passing in opts directly to mapeo-sync

karissa:
ok cool @noffle .

src/api/sync.js
```diff
@@ -20,6 +20,12 @@ class Sync {
       method: 'GET',
       route: `/sync/start?host=${device.host}&port=${device.port}`
     });
+
+  static dataset = (dataset: string) =>
+    blankRequest({
+      method: 'GET',
+      route: `/sync/start?dataset=${dataset}`
```
karissa: this should add the `filepath` for the USB location

PR #112, opened 2018-10-05
fix: migrate media files from fs-blob-store to safe-fs-blob-store 

karissa:
This little PR migrates media blob stores from fs-blob-store to safe-fs-blob-store

If media is not migrated, then people using older versions of Mapeo Mobile (e.g., the version ECA has) who upgrade to the newest version will not be able to see or sync the images they've taken previously.

### DIFF
--- a/src/rnnodeapp/index.js
+++ b/src/rnnodeapp/index.js
@@ -51,13 +51,13 @@ console.log('2: dirs created');
 
 // Migrate media from old fs-blob-store, if needed
 migrateMedia(MEDIA_PATH, function (err) {
-  if (err) throw err;
+  if (err) console.error(err);
 })
 
 // Unpack styles and presets, if needed
 styles.unpackIfNew(STATIC_PATH, function(err, didWrite) {
   console.log('3: unpacked', err, didWrite);
-  if (err) throw err;
+  if (err) console.error(err);
 
   start();
 });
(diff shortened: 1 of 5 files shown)


karissa:
@noffle for some reason there seems to be a race condition with writing and reading the files from the blob stores, can you help me figure out what is going on there? The tests don't pass consistently

karissa:
screenshot from 2018-10-05 17-13-23
@noffle after running tests, it errors saying file doesn't exist, but when i ls the dir, I see the file there.

src/rnnodeapp/index.js
```diff
@@ -47,13 +49,20 @@ mkdirp.sync(MEDIA_PATH);
 mkdirp.sync(STATIC_PATH);
 console.log('2: dirs created');
 
+// Migrate media from old fs-blob-store, if needed
+migrateMedia(MEDIA_PATH, function (err) {
+  if (err) console.error(err);
```
noffle: Won't this potentially start the app before migration is complete?

src/rnnodeapp/migrateMedia.js
```diff
@@ -0,0 +1,41 @@
...
+      })
+    })
+    parallel(tasks, cb)
+  })
+
+  function migrateFile (key, cb) {
+    var filename = path.join(mediaPath, key)
+    fs.stat(filename, function (err, stat) {
+      if (err) return cb(err)
+      if (stat.isFile()) {
```
noffle: `cb` is never called if `isFile()` isn't truthy. will `parallel` hang?

karissa: addressed! nice catch

noffle:
Awesome! All that's left is testing it on a real device /w legacy data and then seeing if the new version of the app can see media. (*groan*)

cc @gmaclennan

PR #114, opened 2019-02-06
Improve speed of ObservationView and MapView

karissa:
# Final list of changes 
* Use a feature collection to speed up map rendering
* Don't re-render source with every render of map component
* Remove react-native-drawer from camera and observations list view
* Remove observations from state and do not rely on redux in listing observations
* Don't iterate over all observations to create a new observation

### initial exploration of performance
Testing with about 500 points, found two main issues:

1. Found that MapboxGL becomes really slow, and sped up immediately after removing the code that renders circles for each observation — was able to move the map freely afterward. 

To fix this, I used a FeatureCollection to render all of the points at once, which seems to have done the trick to speed it up nicely. **I noticed that nothing happens when a point is clicked** but this could be a separate non-related bug?

2. Separately, the react native buttons are non-responsive, even though the Mapbox map is moving freely after removing the above code. 

I pulled out the code that stores observations in redux and replaced it with an api call that happens every time the map is loaded and every time the drawer is loaded. These are fast. I also made the ObservationCell a React `PureComponent` which was suggested by the error messages to make rendering faster.

### Results

Here's a demo of the app now sped up after these changes: https://datbase.org/dat://03d4d3c64d8c4788daa420417a06bfebb38f4af6ffe073fb2119dba570e4a914/contents/2019-02-05-195834.webm

The app is now responsive but is still slow for the first few seconds, the drawer opens and closes with about a 2 second delay. To fix this, I think we need to implement virtual scrolling in the drawer to only load the points in view:

I implemented a way to pass `limit, `gt` and `lt` (among the other possible variables) to the observation list route, to fetch a limited number of observations, which is available in mapeo-server 9.0.1

TODO: we would then use `onEndReached` (https://facebook.github.io/react-native/docs/flatlist#onendreached) to fetch the next batch (e.g., 20 at a time) and append them on to the end of the observation list.


src/components/Views/MapView/Map/Map.js
```diff
@@ -217,30 +217,6 @@ class Map extends React.Component<Props & StateProps & DispatchProps> {
                 : undefined
             }
           >
-            {!!observations && !isEmpty(observations)
```
karissa: We can put this back in the code, or find some other way to address how it is currently slowing down the app. Thoughts welcome

src/components/Views/ObservationsView/ObservationsView.js
```diff
@@ -46,14 +46,39 @@ I18n.translations = {
...
+  constructor (props) {
+    super(props)
+    this.state = {
+      observations: []
+    }
+  }
+
+  componentDidMount () {
+    console.log('component mounted')
+    var observable = Observation.list({ limit: 20 })
```
karissa: This is not really happening yet until we get the outstanding PRs merged in mapeo-core and hyperkv.

src/epics/observations.js
```diff
@@ -26,20 +26,6 @@ import type { StoreState } from '../types/redux';
 import Observation from '../api/observations';
 import { parseObservationRequest } from '../models/observations';
 
-export const observationListEpic = (
```
karissa: fewer redux things for great justice

src/components/Views/MapView/Map/Map.js
```diff
@@ -115,10 +116,24 @@ class Map extends React.Component<Props & StateProps & DispatchProps> {
...
     MapboxGL.setAccessToken(env.accessToken);
   }
 
+  componentDidMount () {
+    var observable = Observation.list({ limit: 20 })
+    var onSuccess = (observations) => {
+      this.setState({observations})
+    }
+    var onError = (err) => {
+      console.log('eff', err)
```
karissa: TODO: handle errors?

src/components/Views/MapView/Map/Map.js
```diff
@@ -191,7 +206,28 @@ class Map extends React.Component<Props & StateProps & DispatchProps> {
...
+          geometry: {
+            type: 'Point',
+            coordinates: [o.lon, o.lat]
+          },
+          properties: {
+            categoryId: o.categoryId
+          }
+        }
+      }).reduce((acc, feature) => (
+        acc.concat(feature)
```
karissa: would this be any faster if we do it on the node side?

src/components/Views/MapView/Map/Map.js
```diff
@@ -217,30 +253,22 @@ class Map extends React.Component<Props & StateProps & DispatchProps> {
...
-                    <MapboxGL.CircleLayer
-                      id={`circles-${o.id || ''}`}
-                      style={mapboxStyles.observation}
-                    />
-                  </MapboxGL.ShapeSource>
-                ))
-              : null}
+          {featureCollection ?
+            <MapboxGL.ShapeSource
+              onPress={(item) => console.log(item)}
```
karissa: onPress doesn't seem to do anything

src/components/Views/MapView/Map/Map.js
```diff
@@ -217,30 +253,22 @@ class Map extends React.Component<Props & StateProps & DispatchProps> {
...
+              key={`observations-source`}
+              id={`observations-source`}
+              shape={{
+                type: 'FeatureCollection',
+                features: featureCollection
+              }}
+            >
+              <MapboxGL.CircleLayer
+                id={`circles`}
```
karissa: TODO: better names for ids here

src/components/Views/ObservationsView/ObservationsView.js
```diff
@@ -46,14 +46,39 @@ I18n.translations = {
...
+    var onSuccess = (observations) => {
+      this.setState({observations})
+    }
+    var onError = (err) => {
+      console.log('eff', err)
```
karissa: TODO: actually handle errors

src/components/Views/ObservationsView/ObservationsView.js
```diff
@@ -77,6 +102,7 @@ class ObservationsView extends React.Component<
           }}
         >
           <FlatList
+            initialNumToRender={20}
```
karissa: TODO: use a constant (should be same as our limit -- sort of a PAGE_SIZE)

src/api/observations.js
```diff
@@ -9,10 +10,10 @@ import type {
 } from '../types/observation';
 
 class Observation {
-  static list = () =>
+  static list = (opts) =>
     jsonRequest({
       method: 'GET',
-      route: '/observations'
+      route: '/observations?' + querystring.stringify(opts)
```
karissa: this will pass the proper limits if mapeo-server in rnnodeapp is set to `9.0.1`

gmaclennan:
^^ Just to clarify @karissa I think in this PR we should raise observation state to the App component, and memoize the GeoJSON and wrap Source + Style in a Pure component. We can look at the drawer and virtualized list in a separate PR.

src/components/Views/MapView/Map/Map.js
```diff
@@ -110,13 +110,64 @@ const mapboxStyles = MapboxGL.StyleSheet.create({
...
+}
+
 class Map extends React.Component<Props & StateProps & DispatchProps> {
   map: any;
 
   constructor() {
     super();
-
     MapboxGL.setAccessToken(env.accessToken);
+    this.handleObservationPress = this.handleObservationPress.bind(this)
```
gmaclennan: The fancy syntax `handleObservationPress = (e) => {` is shorthand for this (as opposed to normal class methods written as `handleObservationPress (e) {`) so this is not needed.

src/components/Views/ObservationsView/ObservationsView.js
```diff
@@ -98,7 +102,7 @@ class ObservationsView extends React.Component<
                 icons={icons}
               />
             )}
-            data={observations}
+            data={Object.keys(observations).map((key) => observations[key])}
```
gmaclennan: Probably best memoize this

karissa: will memoizing cause issues when observations change?

src/components/Views/MapView/Map/Map.js
karissa: nice catch!

src/components/Views/ObservationsView/ObservationsView.js
karissa: nevermind, figured it out

karissa: this was helpful: https://medium.com/ovrsea/react-re-rendering-purecomponents-and-memoization-553a2cc863c3

src/lib/media.js
```diff
@@ -26,3 +26,7 @@ export const getMediaUrl = (id: string, thumbnail?: boolean) => {
 
   return `http://127.0.0.1:9080/media/original/${id}`;
 };
+
+export const getSvgUri = (svg: string) => {
+  return `data:image/svg+xml;utf8,${svg}`.replace(new RegExp('#', 'g'), '%23')
```
gmaclennan: Good find! That must have been frustrating. Let's put switching from SVGs to PNGs next on the todo list. I think this is a performance bottleneck in the app and subject to bugs because not all SVGs are equal and some designs can format incorrectly or cause errors.

gmaclennan:
Great work on this @karissa. Before merging I'd like to hoist the observation state into the parent component, so we don't have two separate observation states in the app, which I think could be a source of bugs in the future. I might have some time this afternoon to look at this, check in with me on Slack if you start working on this.

noffle:
So excited for PNG icons & the speed increase!

PR #120, opened 2019-05-06
fix(migration): use osm-p2p

okdistribute:
The directory structure for the mobile app was different than the migration scripts and desktop. The `osm-p2p` module encapsulates the logic around where the internal leveldb for indexing and for kappa-core are stored on the filesystem. By using the same logic for all instances (migration, desktop, and mobile) it is more straightforward to copy data directly between these deployments without modifying the directory structure.

### DIFF
--- a/src/backend/package.json
+++ b/src/backend/package.json
@@ -9,9 +9,7 @@
   "dependencies": {
     "bindings-noderify-nodejs-mobile": "^10.3.0",
     "debug": "^4.1.1",
-    "kappa-core": "^2.6.2",
-    "kappa-osm": "^3.0.1",
-    "level": "^4.0.0",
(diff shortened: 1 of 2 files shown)


gmaclennan:
Unfortunately I think at this stage this might be a breaking change, which means we would have to have a migration script for the Sinangoe phones that are arriving on Thursday. I'm not sure this is possible at this stage. We could revisit this later? It's not actually possible to access the data folder of a non-debug version of the app (it's completely sandboxed on Android unless the phone is rooted) so this may be a mute point - e.g. it is not possible to copy data folders from phones so the structure doesn't matter.

karissa:
ah ok great. it wasn't clear to me if there was going to be sideloading of data on to the phones

PR #121, opened 2019-05-09
feat(sync): set device name with random string at the end

karissa:


src/frontend/screens/SyncModal/index.js
```diff
@@ -41,10 +42,17 @@ class SyncModal extends React.Component<Props, State> {
   state = { serverPeers: [], syncErrors: new Map(), ssid: null };
   _opened: number;
   _subscriptions: Array<{ remove: () => void }> = [];
+  _deviceName: string;
+
+  constructor (props) {
+    super(props)
+    // recreate this as little as possible
+    this._deviceName = this._deviceName || "Android " + crypto.randomBytes(2).toString('hex').slice(0, 3);
```
karissa: Ideally the device name wouldn't change every time you open/close sync

src/frontend/screens/SyncModal/SyncView.js
```diff
@@ -24,12 +24,12 @@ const Header = ({ onClosePress }: HeaderProps) => (
...
 );
 
-const WifiBar = ({ onPress, ssid }) => (
+const WifiBar = ({ onPress, ssid, deviceName }) => (
   <TouchableNativeFeedback onPress={onPress}>
     <View style={styles.wifiBar}>
       <WifiIcon />
       <Text style={styles.wifiBarText}>
-        <Text style={styles.bold}>Wi-Fi:</Text> {ssid}
+        <Text style={styles.bold}>Wi-Fi:</Text> {ssid}, {deviceName}
```
karissa: how do we want to display the device name?

src/frontend/screens/SyncModal/index.js
gmaclennan: I switched to generate this once during the lifecycle of the app
FYI whenever the constructor is called, this._deviceName would always be undefined, since it's only called when creating a new instance.

src/frontend/screens/SyncModal/SyncView.js
gmaclennan: I think this is a good place to put it. Just tweaked formatting

PR #154, opened 2019-06-19
[FEATURE] Observation edit UX re-designs

ChrisC:
for mockups, see https://www.notion.so/MM-Designs-for-Milestone-2-f449d87e98924a45b2a5e9bdc15af2a2#0264acec2d674acd9373a5cb554a91bc

this doesn't include the designs for the new save-state screens, since that seems like a separate task unto itself (I added that into the milestone spreadsheet)...

### DIFF
--- a/src/frontend/screens/ObservationEdit/ObservationEditView.js
+++ b/src/frontend/screens/ObservationEdit/ObservationEditView.js
@@ -2,6 +2,7 @@
 import React from "react";
 import { ScrollView, View, Text, StyleSheet, TextInput } from "react-native";
 import { withNavigationFocus } from "react-navigation";
+import MaterialIcons from "react-native-vector-icons/MaterialIcons";
 
 import LocationField from "../../sharedComponents/LocationField";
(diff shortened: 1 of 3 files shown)


src/frontend/screens/ObservationEdit/ObservationEditView.js
```diff
@@ -24,7 +26,17 @@ const LocationView = ({ longitude, latitude, accuracy }) => (
...
+          color="orange"
+          style={{ marginRight: 5 }}
+        />
+        <FormattedCoords
+          style={styles.locationText}
+          lat={latitude}
+          lon={longitude}
+        />
         {accuracy && (
           <Text style={styles.accuracy}>{"±" + accuracy.toFixed(2) + "m"}</Text>
```
gmaclennan:
```suggestion
          <Text style={styles.accuracy}>{" ±" + accuracy.toFixed(2) + "m"}</Text>
```

This needs an extra space in the new design so it isn't too close to the coordinates

PR #317, opened 2020-02-18
[WIP] feat: Add setting to share Mapeo APK

gmaclennan:
Add a new option on the setting screen to share the Mapeo APK installer so that it can be used to install Mapeo on another device.

Currently the option brings up a share sheet which shows apps that can be used for sending a file, but it does not show an option for sharing the APK installer to disk.

TODO: Add a way to save the Mapeo APK installer to the Downloads folder, so that the user can put it on a USB drive and move it to another device manually.


gmaclennan:
This is superseeded by the p2p update work

PR #53, opened 2018-06-16
Feature/june w2 helena

howdyitshelena:
- Add cancel modal to editor
- Fix arrow size in ObservationDetailView
- Make sure app is always in portrait mode

PR #388, opened 2020-04-23
fix(SyncView): Fix peers flickering or appearing twice after error

gmaclennan:
Trying to use `peer.name` to de-dupe sync peers, because an errored peer can appear twice when it re-connects. Still needs work to handle:

- Some peers can have the same name even if they are different (desktop devices use the hostname which is not unique).
- When the peer is in "completed" state it is not returned in the peer list from the server, so the errored peer re-appears until the peer reconnects and the errored peer is deduped again. See https://github.com/digidem/mapeo-core/issues/84

### DIFF
--- a/src/frontend/screens/SyncModal/index.js
+++ b/src/frontend/screens/SyncModal/index.js
@@ -77,6 +77,10 @@ const deviceName: string =
     .slice(0, 4)
     .toUpperCase();
 
+// Used to track the order that peers appear, so that new peers do not appear
+// above existing peers and cause the display to jump around and flicker
+const firstSeen = new Map<string, number>();
+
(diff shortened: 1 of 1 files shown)


okdistribute:
Will this also need to be updated in Desktop's sync frontend? Any ideas on how we can share this sync state code?

gmaclennan:
yes, I think it will be needed for Desktop too. I think all of this state management needs to be moved into mapeo-core. I was thinking along the lines of #52 and #53, but it would be worth re-visiting those thoughts with some of the ideas around index state: https://gist.github.com/gmaclennan/c13cf0b63d08cdefc50223af830aa1e9

okdistribute:
This is now fixed in v5.0.0 branch, so this can be closed

PR #448, opened 2020-08-07
feat: Add empty offline map so that points still show if no custom map

### LINKED ISSUE #379: Include a default offline map in app
Currently if Mapeo is offline and the user has not installed any offline map tiles, no map shows up on the phone.

We should include a basic offline map of the world, which shows countries and major cities, that will display as a fallback when there are no offline map tiles, or if the user is outside the area of the offline map tiles.

gmaclennan:
Previously, if Mapeo was offline and the user had not installed an offline map, then no map would show and no points would show.

This PR adds a default offline map with country boundaries, land and ocean, rivers and lakes, and a graticule grid of 0.1°. Currently the offline map data is shipped as geojson via the package [`mapeo-offline-map`](https://github.com/digidem/mapeo-offline-map), which means that all the data is loaded into memory, so we need to keep it small. The current offline map uses simplified data from https://www.naturalearthdata.com. The graticule is included because when the user is zoomed in, no other features are present, so the graticule overlay helps give the user the sense that pan and zoom interactions are working.

mapeo-offline-map-example

Download a [QA build of this PR#448](https://app.bitrise.io/artifact/46204805/p/94ab64a32fe07fab1f8fe190e952f27e)

**TODO:**

- [x] Inset maps in ObservationView do not work

Fixes #379


src/frontend/sharedComponents/MapView.js
```diff
@@ -237,11 +257,31 @@ class MapView extends React.Component<Props, State> {
...
+    // don't setState, because it would be too slow, so we use this.zoomRef to
+    // track the current map zoom. This means both this.zoomRef and setState
+    // need to be called to update the zoom of the map. This should be fixed
+    // with a better approach, because it can lead to bugs. For now, this code
+    // just ensures that the map zooms in when the user clicks the "locate me"
+    // button.
+    this.setState(state => {
+      const newZoom = (this.zoomRef =
+        styleURL === "error"
+          ? this.zoomRef < 4
```
okdistribute: Math.max(4, this.zoomRef) might be more readable

src/frontend/sharedComponents/MapView.js
```diff
@@ -279,17 +315,23 @@ class MapView extends React.Component<Props, State> {
...
             }
-            onDidFinishRenderingMapFully={() =>
-              bugsnag.leaveBreadcrumb("onDidFinishRenderingMapFully")
-            }
+            onDidFinishRenderingMapFully={() => {
+              if (styleURL !== "error") {
+                // For the default offline map (that does not contain much
+                // detail) we stay at zoom 4, but if we do load a style then we
+                // zoom in to zoom 12 once the map loads
+                this.setState({ zoom: 12 });
```
okdistribute: if 12 and 4 are special, can we make them constants at the top of the file?

src/frontend/sharedComponents/MapView.js
gmaclennan: Thanks @okdistribute, problem with open source coding late on Friday, everyone sees your dumb code :P

PR #458, opened 2020-09-08
feat: Translate observation categories & details

### LINKED ISSUE #455: Displays "value", not "label"  under details of observation
Form displays labels correctly
Screen Shot 2020-09-02 at 3 05 48 PM

Screenshot_20200902-102326_Mapeo
Screen Shot 2020-09-02 at 3 05 54 PM

Screenshot_20200902-102337_Mapeo

**But the observation screen shows the value**
Screenshot_20200902-111314_Mapeo

and so does the share.
Screen Shot 2020-09-02 at 3 04 17 PM

gmaclennan:
This PR adds functionality to Mapeo Mobile to display translated fields for presets. It depends on translations being built with [`mapeo-settings-builder`](https://github.com/digidem/mapeo-settings-builder) v3.1.0 or later (presets built with earlier versions should still work, they will just not be translated).

`mapeo-settings-builder@^3.1.0` includes scripts to extract messages from preset and field definitions for translation, and also a build script to include translated messages into the config package.

This PR should also fix the display of select fields, so that the label rather than the value is shown.

**TODO:**

- [x] Translate fields in share message

This PR is built on #448

Download a QA version of Mapeo for this PR: https://app.bitrise.io/artifact/48898947/p/6fe9bad13146a7fda75afc887d5a191a

Closes #455


okdistribute:
Does this also need to be ported to Desktop? We didn't talk about that extra work today, not sure if it's needed?

gmaclennan:
> Does this also need to be ported to Desktop? We didn't talk about that extra work today, not sure if it's needed?

It does, eventually. I do not _think_ it is an immediate priority for @arky to also have this on Desktop. However it does also fix the display of select fields. As part of this PR I am abstracting the code to do this into re-usable components, which could be shared between Mobile and Desktop.

src/frontend/sharedComponents/FormattedData.js
```diff
@@ -0,0 +1,154 @@
+// @flow
```
okdistribute, line 1: This seems useful! Can we abstract this out as part of this PR, perhaps to @mapeo/react-components? It would be great to re-use this in Md (and get the formatted preset names, too) for the Report. Come to think of it, the Md Report and the Mm Share screen ought to probably share a lot of code... :)

src/frontend/lib/utils.js
```diff
@@ -142,6 +143,35 @@ export function formatCoords({
...
+  return typeof tagValue === "undefined" ? defaultValue : tagValue;
+}
+
+/**
+ * Convert a select option which could either be a string or an object with
+ * label and value props, to an object with label and value props TODO: Show
+ * meaningful translated values for null and boolean, but these types are not
+ * used widely in presets yet
+ */
+export function convertSelectOptionsToLabeled(
```
okdistribute, line 159: This also seems like it will be important to re-use this same rendering logic in other UIs, but it is not a react component. If the only argument is a SelectOptions object, I wonder how we can make it more tightly coupled, so it isn't in a 'utils' file?

src/frontend/sharedComponents/FormattedData.js
```diff
@@ -0,0 +1,154 @@
...
+// anywhere yet)
+function getValueLabel(
+  value: null | boolean | number | string,
+  field: Field
+): string {
+  if (field.type === "select_one" || field.type === "select_multiple") {
+    // Look up label from field options. This is not necessary for presets
+    // created with mapeo-settings-builder@^3.1.0, which will have these options
+    // in the translation file, but is needed for older versions of presets
+    const options = convertSelectOptionsToLabeled(field.options);
```
okdistribute, line 141: is there ever a time someone would use field.options in a Mapeo React component in the presentation/view layer without having it already be parsed by the 'convertSelectOptionsToLabeled' function? (not counting mapfilter hooks or custom functions that require the original field value)

src/frontend/sharedComponents/FormattedData.js
okdistribute: I started this here. https://github.com/digidem/mapeo-components/pull/1

In the desktop app I copy/pasted the relevant two components (FormattedFieldProp & FormattedValue) to get the Report PDF branch ready to launch. It would be an extra piece of work (1-2 weeks) to refactor these components into a shared library.

gmaclennan: Yes, wrote this in this way as a first step towards extracting into a shared library, see https://github.com/digidem/mapeo-mobile/issues/460 however this depends on #459 which I would rather not do as part of this PR.

src/frontend/lib/utils.js
gmaclennan: Yeah I have been wondering about that. This is used by:

1. Select One data editing component
2. Select Multiple data editing component
3. Select one and multiple data view component

On desktop the editing component will likely be different (due to different UI patterns on Desktop) but will also need this function. As such I think it needs to be a helper function shared by these components. I think we can drop support for this in a later version of Mapeo Presets, or perhaps do this transform on the server. This is necessary because we originally supported options to be an array of strings, but later supported labels and values. Once we have a preset editor, these simplicity of just adding an array of strings is no longer so important.

src/frontend/sharedComponents/FormattedData.js
gmaclennan: I don't think so, so perhaps we can transform options on the server or at the client api level, and treat `Array<string>` select options as a deprecated option that we change before the front-end receives it.

PR #512, opened 2021-02-08
fix: Fix import config button crash

gmaclennan:
The `<Button>` component currently expects children to be a string, so it doesn't wrap `<FormattedMessage>` in a `<Text>` component. We need to fix `<Button>` but this is a short-term fix.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "mapeo-mobile",
-  "version": "5.0.3",
+  "version": "5.1.0",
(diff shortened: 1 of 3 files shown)


luandro:
Was about to send a fix for this, didn't notice you had one already @gmaclennan. Also added the pt translations so we don't need to wait for Crowdin translation updates. Looks good to me, will merge once checks pass.

gmaclennan:
💯

luandro:
Should I delete the branch after merge?

Issue #513, opened 2021-02-10
Write preset `tags` to observations


gmaclennan:
**Expected behaviour**
When a user creates a new observation and selects a preset, the `tags` field in the preset definition should be written to the newly created observation. E.g. if the user creates a new observation and selects a "River" preset, and the River preset is defined by this JSON:

```json
{
  "icon": "animal",
  "fields": ["animal-type"],
  "geometry": ["point", "area"],
  "tags": { "type": "animal" },
  "terms": [],
  "name": "Animal"
}
```

Then the newly created observation should have `"type": "animal"` as one of its tags.

**Current behaviour**
The `tags` property of the preset definition is ignored, and instead a field `categoryId` is written to the newly created observation's tags, with the value being the `id` of the preset. This code is here:

https://github.com/digidem/mapeo-mobile/blob/53c5655bf934b272d3fa6e5e6f3407ebd9fd70c1/src/frontend/screens/CategoryChooser.js#L81-L89

**Implementation ideas**
- Modify the [category chooser code](https://github.com/digidem/mapeo-mobile/blob/53c5655bf934b272d3fa6e5e6f3407ebd9fd70c1/src/frontend/screens/CategoryChooser.js#L81-L89) to merge the value of `preset.tags` into `draftValue.tags` when updating the draft.
- We need to handle the user changing the preset/category for an observation after they have previously selected a category. Changing a category should:
    1. Remove tags that were applied by the previously matched category (e.g. if the user had previously selected the preset "animal" for the observation, which writes the tag `"type": "animal"` to the observation, then this tag should be removed when the user changes the category.
    2. Do not remove any other tags (e.g. the user could have filled in details or notes, which are also stored as tags on the observation).
    3. The special `categoryId` tag should change to the ID of the newly selected category (note: this already happens in existing code).
    4. Tags should only be removed if there is an exact match. E.g. if the observation was previously categorized as "animal", which wrote the tag `"type": "animal"`, but if the user had changed the tag to be `"type": "monkey"`, then changing the category should not delete that tag, but the tag may be over written if the newly selected category/preset contains a new value for that tag.
e.g.
```js
const oldPreset = presets[draftValue.tags.categoryId]
const newPreset = presets[newCategoryId]
const newDraftValue = {
  ...draftValue,
  tags: { categoryId: newCategoryId }
}
// Copy across user-edited tags
Object.keys(draftValue.tags).forEach(key => {
  // Skip copying tags that come from the previously selected preset
[...]

PR #514, opened 2021-02-10
fix: Write preset tags to observations

luandro:
Should solve #513

Add logic to include tags from category and clear tags from old category.

### DIFF
--- a/src/frontend/screens/CategoryChooser.js
+++ b/src/frontend/screens/CategoryChooser.js
@@ -71,55 +71,41 @@ const Item = React.memo(
 const CategoryChooser = ({ navigation }: { navigation: NavigationProp }) => {
   const [{ presets }] = useContext(ConfigContext);
   const [{ value: draftValue }, { updateDraft }] = useDraftObservation();
-
   const presetsList = Array.from(presets.values())
     // Sort presets by sort property and then by name, then filter only point presets
     .sort(presetCompare)
     // Only show presets where the geometry property includes "point"
     .filter(p => p.geometry.includes("point"));
   const handleSelectPreset = (selectedPreset: Preset) => {
-    if (draftValue && Object.keys(draftValue.tags).length > 0) {
-      // We're updating an observation
-      // Tags from current preset
-      const currentDraftTags = draftValue.tags;
-      // Tags from previous preset
-      const prevPresetTags =
-        (Object.fromEntries(presets)[currentDraftTags.categoryId] || {}).tags ||
-        {};
-      // Create object with new tags only
-      const draftTags = Object.keys(currentDraftTags).reduce(
-        (previous, current) => {
-          // Check if tag belongs to previous preset
-          const tagIsFromPrevPreset =
-            typeof currentDraftTags[current] !== "undefined" &&
-            currentDraftTags[current] === prevPresetTags[current];
-          // If belongs to previous preset, ignore it
-          if (tagIsFromPrevPreset) return;
-          // Else, include in new object
-          return {
-            ...previous,
-            [current]: currentDraftTags[current],
-          };
-        },
-        {}
-      );
-      updateDraft({
-        tags: {
-          ...selectedPreset.tags,
-          ...draftTags,
-          categoryId: selectedPreset.id,
-        },
-      });
-    } else {
-      // Creating a new observation
-      updateDraft({
-        tags: {
-          ...selectedPreset.tags,
-          ...(draftValue || {}).tags,
-          categoryId: selectedPreset.id,
-        },
-      });
-    }
+    // Tags from current preset
+    const currentDraftTags = (draftValue || {}).tags || {};
+    // Tags from previous preset
+    const prevPresetTags =
+      (presets.get(currentDraftTags.categoryId) || {}).tags || {};
+    // Create object with new tags only
+    const draftTags = Object.keys(currentDraftTags).reduce(
+      (previous, current) => {
+        // Check if tag belongs to previous preset
+        const tagIsFromPrevPreset =
+          typeof currentDraftTags[current] !== "undefined" &&
+          currentDraftTags[current] === prevPresetTags[current];
+        // If belongs to previous preset, ignore it
+        if (tagIsFromPrevPreset) return previous;
+        // Else, include in new object
+        return {
+          ...previous,
+          [current]: currentDraftTags[current],
+        };
+      },
+      {}
+    );
+    updateDraft({
+      tags: {
+        ...selectedPreset.tags,
+        ...(draftTags || (draftValue || {}).tags),
+        categoryId: selectedPreset.id,
+      },
+    });
     navigation.navigate("ObservationEdit");
   };
 
--- a/src/frontend/screens/CategoryChooser.js
+++ b/src/frontend/screens/CategoryChooser.js
@@ -101,8 +101,8 @@ const CategoryChooser = ({ navigation }: { navigation: NavigationProp }) => {
     );
     updateDraft({
       tags: {
+        ...draftTags,
         ...selectedPreset.tags,
-        ...(draftTags || (draftValue || {}).tags),
         categoryId: selectedPreset.id,
       },
     });
--- a/src/frontend/screens/CategoryChooser.js
+++ b/src/frontend/screens/CategoryChooser.js
@@ -77,14 +77,49 @@ const CategoryChooser = ({ navigation }: { navigation: NavigationProp }) => {
     .sort(presetCompare)
     // Only show presets where the geometry property includes "point"
     .filter(p => p.geometry.includes("point"));
-
-  const handleSelectPreset = (preset: Preset) => {
-    updateDraft({
-      tags: {
-        ...(draftValue || {}).tags,
-        categoryId: preset.id,
-      },
-    });
+  const handleSelectPreset = (selectedPreset: Preset) => {
+    if (draftValue && Object.keys(draftValue.tags).length > 0) {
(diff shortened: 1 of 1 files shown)


src/frontend/screens/CategoryChooser.js
```diff
@@ -71,20 +71,52 @@ const Item = React.memo(
...
-      },
-    });
+    const newTags = preset.tags;
+    if (Object.keys(draftValue.tags).length > 0) {
+      // We're updating an observation
+      // Get tags from old category
+      const oldTags = draftValue.tags;
+      const oldCategoryTags = presetsList.filter(
+        p => p.id === oldTags.categoryId
+      )[0].tags;
```
gmaclennan:
```suggestion
      const oldCategoryTags = (presets[oldTags.categoryId] || {}).tags || {}
```
Much simpler :), but it's important to handle the case where there is not a preset matching the categoryId, or where categoryId is undefined (this suggestion will do that - are you using Flow to check this? I helps catch stuff like this)

src/frontend/screens/CategoryChooser.js
```diff
@@ -71,20 +71,52 @@ const Item = React.memo(
...
-        ...(draftValue || {}).tags,
-        categoryId: preset.id,
-      },
-    });
+    const newTags = preset.tags;
+    if (Object.keys(draftValue.tags).length > 0) {
+      // We're updating an observation
+      // Get tags from old category
+      const oldTags = draftValue.tags;
+      const oldCategoryTags = presetsList.filter(
```
gmaclennan: I know we've been using "category" as user-facing terminology, but let's keep to "preset" in the code terminology, since otherwise this can be confusing. E.g. call this `prevPresetTags`.

src/frontend/screens/CategoryChooser.js
```diff
@@ -71,20 +71,52 @@ const Item = React.memo(
...
+        p => p.id === oldTags.categoryId
+      )[0].tags;
+      // Get list of new tags
+      const updatedTags = Object.keys(oldTags).reduce((previous, current) => {
+        // Check if tag belongs to old category
+        const hasOldInfo =
+          Object.entries(oldCategoryTags).filter(([key, value]) => {
+            return key === current && value === oldTags[current];
+          }).length > 0;
```
gmaclennan:
```suggestion
        const tagIsFromPreset = typeof oldCategoryTags[current] !== 'undefined' && oldCategoryTags[current] === oldTags[current]
```
I think this is easier to read and understand what we are checking.

src/frontend/screens/CategoryChooser.js
```diff
@@ -71,20 +71,52 @@ const Item = React.memo(
...
-
   const handleSelectPreset = (preset: Preset) => {
-    updateDraft({
-      tags: {
-        ...(draftValue || {}).tags,
-        categoryId: preset.id,
-      },
-    });
+    const newTags = preset.tags;
+    if (Object.keys(draftValue.tags).length > 0) {
```
gmaclennan: `draftValue` can be `undefined`

src/frontend/screens/CategoryChooser.js
```diff
@@ -71,20 +71,52 @@ const Item = React.memo(
...
     .filter(p => p.geometry.includes("point"));
-
   const handleSelectPreset = (preset: Preset) => {
-    updateDraft({
-      tags: {
-        ...(draftValue || {}).tags,
-        categoryId: preset.id,
-      },
-    });
+    const newTags = preset.tags;
```
gmaclennan: As a matter of code style, I think it's better to just use `preset.tags` to refer to this throughout, since assigning it to a new name (`newTags`) just means anyone reading the code has to go back through the code to check what it refers to (here, `newTags` and `preset.tags` are referring to the same value).

gmaclennan (changes requested):
Great stuff Luandro. I've made some suggestions that simplify the code and catch some edge-cases (undefined values).

Since this is quite a complex bit of code now, I think it's good to be really clear with naming variables to help future readers. Here's some ideas:

- `selectedPreset` = The preset selected by the user (currently called `preset`)
- `currentDraftTags` = The tags currently defined on `draftValue` (currently called `oldTags`)
- `prevPresetTags` = The tags from the previously selected preset (if there was one)
- `updatedTags` = `draftTags` with tags from the previously selected preset removed

Worth putting some comments in the code to clarify this.

Maybe avoid the `newTags` terminology, and instead just keep this as `selectedPreset.tags`, which is explicit.

Can you also check that these tags that come from the preset are not displayed in the `ObservationView` component? I don't think they are, since I think we only display tags defined referenced by `fields` in the preset, but worth checking. We will need to fix this on Desktop, because the Desktop code also shows tags on any observation that are not referenced by `fields`.

luandro:
Thanks for thoughtful review @gmaclennan. Pushed new changes hopefully taking every comment into account.

messages/es.json
```diff
@@ -300,6 +300,10 @@
     "description": "Label for name & version of current configuration",
     "message": "Configuración actual"
   },
+  "screens.Settings.importConfig": {
+    "description": "Button to import Mapeo config file",
+    "message": "Importar configuración"
+  },
```
gmaclennan: Please remove from commit, this should be in the develop branch already.

src/frontend/screens/CategoryChooser.js
```diff
@@ -71,20 +71,54 @@ const Item = React.memo(
...
-      },
-    });
+  const handleSelectPreset = (selectedPreset: Preset) => {
+    if (draftValue && Object.keys(draftValue.tags).length > 0) {
+      // We're updating an observation
+      // Tags from current preset
+      const currentDraftTags = draftValue.tags;
+      // Tags from previous preset
+      const prevPresetTags =
+        (Object.fromEntries(presets)[currentDraftTags.categoryId] || {}).tags ||
```
gmaclennan:
```suggestion
        (presets.get(currentDraftTags.categoryId) || {}).tags ||
```
My suggestion probably was confusing, I thought this was an object, not a Map. Since it's a Map, we can just get the preset directly.

src/frontend/screens/CategoryChooser.js
```diff
@@ -71,20 +71,54 @@ const Item = React.memo(
...
-      tags: {
-        ...(draftValue || {}).tags,
-        categoryId: preset.id,
-      },
-    });
+  const handleSelectPreset = (selectedPreset: Preset) => {
+    if (draftValue && Object.keys(draftValue.tags).length > 0) {
+      // We're updating an observation
+      // Tags from current preset
+      const currentDraftTags = draftValue.tags;
```
gmaclennan:
```suggestion
      const currentDraftTags = (draftValue | {}).tags || {};
```
If you change this line to this, then you can get rid of the if statement and the else below.

gmaclennan (changes requested):
Great, just a couple more things, cleaning up the code.

Looks like the translations are in this branch - can you maybe rebase onto `develop`?

luandro:
Looking much better now @gmaclennan, great tips :+1:

src/frontend/screens/CategoryChooser.js
```diff
@@ -71,18 +71,39 @@ const Item = React.memo(
...
+        };
+      },
+      {}
+    );
     updateDraft({
       tags: {
-        ...(draftValue || {}).tags,
-        categoryId: preset.id,
+        ...selectedPreset.tags,
+        ...(draftTags || (draftValue || {}).tags),
```
gmaclennan:
```suggestion
        ...draftTags,
```
`draftTags` will always be at least `{}`, so the other side of this `||` is never reached, and anyway, if `draftValue.tags` did exist, we've already copied them to `draftTags`.

src/frontend/screens/CategoryChooser.js
```diff
@@ -71,18 +71,39 @@ const Item = React.memo(
...
+          [current]: currentDraftTags[current],
+        };
+      },
+      {}
+    );
     updateDraft({
       tags: {
-        ...(draftValue || {}).tags,
-        categoryId: preset.id,
+        ...selectedPreset.tags,
```
gmaclennan, line 105: I _think_ this needs to come last (but a bit tired so will need to re-visit in the morning). If this is first then it could be overwritten by existing tags on an existing draft.

gmaclennan (changes requested):
Getting there :)
One change, and a question.
Also, I need to get CI working again so that the tests run.

src/frontend/screens/CategoryChooser.js
luandro: lol true

src/frontend/screens/CategoryChooser.js
luandro: True, there might be some case that draft may want to carry an object with the same key.

gmaclennan:
Great, I'm waiting for Github actions to run on https://github.com/digidem/mapeo-mobile/pull/515 so that I can test install of the compatible version of NDK

gmaclennan:
@luandro I managed to fix the CI build, for more details of the error see #517. I'm wondering what the testing plan is for this to check it works? At a minimum, I would like to do a manual test to ensure that this works as intended, e.g. by using presets that define `tags` and checking that the created observations have those tags, and also checking changing the preset for an existing observation to ensure that tags are removed as planned.

We are also in dire need of strong unit tests for things like this, and that's certainly possible for this, using https://testing-library.com - if you feel up to it, it would be a great place to start getting unit tests in place.

gmaclennan (approved):

PR #515, opened 2021-02-11
chore: Update NDK version [WIP]

gmaclennan:
Looks like Github has updated their images to NDK 21.4.7075529. We could install our version as part of the build process, but testing to see if builds work in this newer NDK version.


gmaclennan:
This was part of an update to MacOS 11 and MacOS 10.15 images 3 days ago. They updated NDK from r21e to r21f. Unfortunately NDK r21f causes errors in Mapeo, so need to figure out how to get the Github action to install the compatible NDK version.

gmaclennan:
Close in favour of #516

PR #519, opened 2021-02-17
Add react-testing-library and Jest suite

luandro:
Using this command to test: `npx jest src/frontend/__tests__`

Getting error:

```
  ● Test suite failed to run

    TypeError: globals.Headers is not a constructor

      122 |   const startupTime = Date.now();
      123 | 
    > 124 |   const req = ky.extend({

```


gmaclennan:
oi @luandro, great to see progress on this. The specific issue here is because we are running tests in a Node environment, but the module `ky` that we use (for making http requests) is designed to run in the browser, and expects certain global variables to exist (e.g. the `Headers` constructor). The solution to this issue is to use [`ky-universal`](https://github.com/sindresorhus/ky-universal#faq), however I do not think this is the right solution in our case.

`ky` is being used in [`api.js`](https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/api.js) to communicate with our backend (Mapeo Core) which runs in a separate process on mobile. One solution would be to run the whole backend for tests, so we don't really mock anything. However, http (via `ky`) is not the only way that the frontend communicates with the backend: we also make IPC calls using built-in methods from `nodejs-mobile-react-native`. This will not be possible to get working in a testing environment, so I don't think it will be possible to run a backend that the frontend can communicate with in a testing env.

I think the right place to mock in our case is the `api.js` file. I think we should mock this file, starting with the methods that are used the CategoryChooser screen, which are these calls from the Presets Context: https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/context/ConfigContext.js#L138-L141 We can mock these to return fake data for the category screen test (we should use the default presets as the "fake" data).

Does this make sense? I can help more in delving into how to mock things. With Jest I find the docs on this hard to understand: https://jestjs.io/docs/en/manual-mocks

gmaclennan:
Fixed the Jest integration so it is now working. There was a lot of stuff that needed to be setup and fixed, some of it due to our app structure was created with an older version of `react-native` and some subtle things have changed.

- The Jest config option `"transform"` initially used here [does not work in recent versions of react-native]()
    ```json
    "transform": {
       "^.+\\.js$": "<rootDir>/node_modules/react-native/jest/preprocessor.js"
    }
    ````
- Most frustratingly, and probably a big cause of issues including mocking: the Jest `react-native` preset for some reason does not pick up `.babelrc`, instead it only looks for `babel.config.js`. Found this out by going through differences with a bare `npx react-native init` app.
- Our app has several context providers. If we are testing a component that uses any of these contexts then we need to mock them all in our custom render function in `lib/test-utils`.
- By default Jest does not do any code transforms on modules in `node_modules`. However, for react-native many modules are published without transforms, e.g. raw Typescript or ESM code. All of these modules need to be excluded from the `transformIgnorePatterns` option.

gmaclennan:
**This does not actually run any meaningful Component tests yet**. The PR in the current state just ensures that the Jest config is setup correctly and a has one test that checks a the `CategoryChooser` screen renders in the test environment.

@ErikSin I would like to get this merged sooner rather than later, so maybe let's just put one simple unit test in here then merge.

src/frontend/api.spec.js
```diff
@@ -4,7 +4,7 @@ import nodejs from "nodejs-mobile-react-native";
 import RNFS from "react-native-fs";
```
ErikSin: @gmaclennan I'm not sure what happened, but it seems like these tests are now failing and timing out? I just uncommented them out.

gmaclennan:
@ErikSin added a few fixes here:

- Removed the haste map errors by ignoring folders. You only see these on first run, because Jest caches the hast map. To clear it: `npm test -- --clearCache`
- The warning on the `Settings` test was due to the Config screen being part of the import, and that screen imports expo-document-picker. I add a mock for this module in `/__mocks__`, which fixes the warning.
- `getByText()` was not working, I debugged and found this is because `<FormattedMessage>` renders text in a `<Fragment>` by default, and getByText() can only read direct children of a `<Text>` element. It [looks like this might be fixed](https://github.com/callstack/react-native-testing-library/pull/554) one day, but for now I have changed the default wrapper (only in tests) to be `<Text>` rather than `<Fragment>`. I remember I first did this in the main app, but it caused issues with formatting and layout of text, hence I kept it as `<Fragment>` 
- The failure of api.spec.js was due to a breaking change in the implementation of `fakeTimers()` in jest@27. I had to manually use "legacy" timers, however we may need to use this fix in the future in order to use the modern fake timers: https://github.com/callstack/react-native-testing-library#custom-jest-preset
- Ideally we follow [this guide](https://callstack.github.io/react-native-testing-library/docs/how-should-i-query) and use our default intl text for querying.
- Regarding Navigation, I'm not sure it makes sense to test navigation within unit tests, because we need to render the whole app to do that. Perhaps we implement navigation tests in the e2e tests - we already have a few. We can do some basic mocking like I've done in the settings test - checking that pressing a button calls `navigate()` with the given argument.

ErikSin:
@gmaclennan I agree with the not testing the navigation. I just mocked out the navigation hook because it was throwing an error otherwise.

PR #520, opened 2021-02-17
Sharing observations working on Telegram

### LINKED ISSUE #380: Sharing an observation does not work in all apps
The share button for an observation should allow the observation details and photos to be shared with different apps installed on the phone.

Currently it does not work consistently across different apps. From @aliya-ryan:

> - I tried to share with gmail on a phone where it didn't already have my account set up. Once I realised and try to go back this didn't work, and I had to quit app and reopen to get back to Mapeo.
> - Sharing on a phone where I did have gmail set up worked as I would expect.
> - Sharing via bluetooth to my computer from both phones was unsuccessful.
> - Sharing via whatsapp works, but if there is more than one photo then it sends one message with all info for each photo.
> - I shared with google maps - it only seems to recognise the photo info and you have to them tell it where to place it (on google maps). I don't know if it worked (photo is not there but perhaps it gets reviewed).

luandro:
Fixes #380 

- Update react-native-share to latest


luandro:
Not sure what else to do, created [issue](https://github.com/react-native-share/react-native-share/issues/966) on `react-native-share` repo.

gmaclennan:
I have a feeling we need to specify different options for each app that we share with. It would be worth testing different share methods with Telegram.

The Java code in `react-native-share` for creating a ShareIntent for the share is here: https://github.com/react-native-share/react-native-share/blob/0686c48cdc8041895995a89435e529ba3bbeea92/android/src/main/java/cl/json/social/ShareIntent.java

The code in Telegram for accepting a share I _think_ is here: https://github.com/DrKLO/Telegram/blob/master/TMessagesProj/src/main/java/org/telegram/ui/LaunchActivity.java#L1269

I've only briefly scanned it, but it looks like Telegram does have logic for a message with both text and an image. It would be worth figuring out what properties of a ShareIntent react-native-share sets (maybe with some debugging statements), and check the code for Telegram to see how they would be handled.

There could be an issue with sharing images via http URLs, vs. sharing file:// urls, which it seems like Telegram handles differently (my guess, based on a rapid read of the code and not knowing any Java). It is possible that we could add an API method to mapeo-core to return a direct file reference for an image, or, perhaps more robust going forwards (since we might not in the future store images as files on disk) we could copy the file to temp file storage as part of the share step.

gmaclennan:
See also: https://gist.github.com/mahdimortazavi/e54bc60b4ba58e69a2d8cc827260bb63#file-gistfile1-txt-L3

luandro:
Also found this: https://stackoverflow.com/questions/35205119/sharing-image-and-text-together-is-not-working-for-telegram-on-android

Looking at the Telegram code I wasn't able to find any pointer of how image+text is handled.

I tried mimicking some details that I thought might be relevant from the gist and the stackoverflow question, but they didn't do the trick. But both seem to create a new file before sharing, which I started trying to do, but realized it would be a rabbit-hole of Java modules to import and understand.

Before going down that rabbit hole I thought it would be best to wait for some input, as I may have missed something fundamental.

*update*: The intent we're sending looks ok:
```
#Intent;action=android.intent.action.SEND;type=image/*;launchFlags=0x1;S.android.intent.extra.SUBJECT=Alerta%20Mapeo%20%E2%80%94%20_*Animal*_%20%E2%80%94%2011%20de%20fev.%20de%202021%2010%3A26;S.android.intent.extra.TEXT=Alerta%20Mapeo%20%E2%80%94%20_*Animal*_%20%20%0A11%20de%20fev.%20de%202021%2010%3A26%20%20%0A%5C-14.0645622%2C%20-47.456437%0A%0AUm%20macaco%0A%0A%E2%80%94%20Enviado%20pelo%20Mapeo%20%E2%80%94;end
```

achou11:
closing due to staleness. think there's wider work around improving share functionality to interface better with apps like Signal and Telegram

PR #523, opened 2021-03-05
Add arrow indicating user direction

luandro:
Simply enables `showsUserHeadingIndicator` on [MapboxGL.UserLocation](https://github.com/react-native-mapbox-gl/maps/blob/master/docs/UserLocation.md) component which shows a small arrow which indicates direction the device is pointing relative to north.


ErikSin:
@luandro It seems to be working when I refresh the page. It will show the initial direction of the phone, but if I change direction, I have to refresh the app for the change in direction to be reflected. Does the map, or this component, have a refresh rate that we could increase?

ErikSin:
Closing because another PR was created for this.

PR #524, opened 2021-03-08
Peer-to-peer upgrades

hackergrrl:
Work-in-progress. The backend (`src/backend/`) is mostly finished & is ready for review.


android/app/src/main/java/com/mapeo/MainApplication.java
```diff
@@ -46,16 +46,12 @@ public boolean getUseDeveloperSupport() {
...
+      // example:
       // packages.add(new MyReactNativePackage());
-
-      // Add unimodules
-      List<ReactPackage> unimodules = Arrays.<ReactPackage>asList(
-        new ModuleRegistryAdapter(mModuleRegistryProvider)
-      );
-      packages.addAll(unimodules);
-
+      packages.add(new ModuleRegistryAdapter(mModuleRegistryProvider));
```
gmaclennan, line 53: Does this actually work? My understanding was that this is a list of modules. These "unimodules" are the `expo-` modules that we use for:

- Camera
- Geolocation (GPS)
- Detecting device rotation for landscape photos

Can we check all of this still works in a build with this option?

android/app/src/main/java/com/mapeo/MainApplication.java
```diff
@@ -75,37 +71,5 @@ public void onCreate() {
     super.onCreate();
     BugsnagReactNative.start(this);
     SoLoader.init(this, /* native exopackage */ false);
-    initializeFlipper(this, getReactNativeHost().getReactInstanceManager());
```
gmaclennan: Any particular reason why we're removing Flipper here?

src/backend/lib/constants.js
```diff
@@ -0,0 +1,36 @@
...
+    Error: 3,
+  },
+
+  Server: {
+    Idle: 1,
+    Sharing: 2,
+    Draining: 3,
+    Error: 4,
+  },
+};
```
gmaclennan: This is a style preference, but I prefer strings for state, because it makes debugging easier. If I log the state value then I know what it is, but when it's a number I need to look it up.

gmaclennan: PS. When I've been writing with Typescript/Flow, then I have dropped using constants, and instead just typed the state variable. Without Typescript I appreciate that using a constant avoids nasty typo errors!

src/backend/lib/local-upgrade-info.js
```diff
@@ -0,0 +1,102 @@
+const path = require("path");
+const fs = require("fs");
+const RWLock = require("rwlock");
+
+const UPGRADE_INFO_FILENAME = require("./constants").UPGRADE_INFO_FILENAME;
```
gmaclennan:
```suggestion
const { UPGRADE_INFO_FILENAME } = require("./constants");
```

Really not bothered about this, but I think most of the code in Mapeo uses destructuring for this kind of import.

src/backend/lib/upgrade-manager.js
```diff
@@ -0,0 +1,147 @@
+const UpgradeStorage = require("./upgrade-storage");
+const UpgradeServer = require("./upgrade-server");
+const UpgradeDownload = require("./upgrade-download");
+const semver = require("semver");
+const log = require("debug")("p2p-upgrades:manager");
+const validate = require("./validate-upgrade-state");
+
+class UpgradeManager {
+  constructor(dir, port, currentVersion, emitFn, listenFn, removeFn) {
```
gmaclennan:
```suggestion
  constructor({ dir, port, currentVersion, emitFn, listenFn, removeFn}) {
```

Code style pref: With this many arguments it is easy for a mistake to creep in when calling the constructor by missing an argument or getting them in the wrong order (I've spent some time debugging issues like this). Passing arguments as properties on an `options` object has less chance of error, and is easy with destructuring like this.

src/backend/package.json
```diff
@@ -9,29 +9,42 @@
...
   "license": "MIT",
+  "scripts": {
+    "lint": "eslint *.js \"lib/*.js\" \"test/*.js\"",
+    "test": "tape \"test/*.js\""
+  },
   "dependencies": {
     "@bugsnag/js": "^6.3.1",
     "bindings-noderify-nodejs-mobile": "^10.3.0",
-    "debug": "^4.1.1",
-    "discovery-swarm": "6.0.0",
```
gmaclennan, line 19: I remember this was in here to lock the version of discovery-swarm used by mapeo-core, since we were seeing errors in ^6.0.0. Do you know if the original issue was resolved? I only have vague memories of this.

src/backend/server.js
```diff
@@ -131,10 +142,61 @@ function createServer({ privateStorage, sharedStorage, flavor }) {
...
+      const listenFn = rnBridge.channel.on.bind(rnBridge.channel);
+      const removeFn = rnBridge.channel.removeListener.bind(rnBridge.channel);
+      upgradeManager = new UpgradeManager(
+        upgradePath,
+        port,
+        version,
+        emitFn,
+        listenFn,
+        removeFn
+      );
```
gmaclennan: This architecture makes me nervous... I came here looking for the boundary between UpgradeManager and the `backend/server.js` code that hooks up the front-end. Giving UpgradeManager responsibility for managing listeners on the IPC channel feels prone to bugs, and figuring out exactly when and how events are emitted to the front-end is harder to follow. Is there a different way of implementing this?

src/frontend/api.js
```diff
@@ -6,13 +6,18 @@ import nodejs from "nodejs-mobile-react-native";
...
+import AppInfo from "./lib/AppInfo";
 
 import type {
   Preset,
   Field,
   Metadata,
   Messages,
 } from "./context/ConfigContext";
+
+import { version as APP_VERSION } from "../../package.json";
```
gmaclennan, line 19: The actual version of Mapeo Mobile (returned by `react-native-device-info`) includes a build number, e.g. `2.4.1.839`. There are scenarios where we might have a broken build and release a new build that only increments the build number. Also, we _may_ change the build process where the version in `package.json` is not kept in sync, so I think better to read the version from `DeviceInfo` which will always be set.

src/frontend/api.js
```diff
@@ -137,6 +142,19 @@ export function Api({
...
+    const apkName = "mapeo.apk";
+    const tmpApkPath = `${tmpDir}/${apkName}`;
+    await RNFS.mkdir(tmpDir);
+    log("+++ 6");
+    await RNFS.copyFile(AppInfo.sourceDir, tmpApkPath);
+    log("+++ 7");
+    // await RNFS.unlink(tmpDir);
+    nodejs.channel.post("apk-ready", tmpApkPath, APP_VERSION);
+  });
+
```
gmaclennan: Why do the file copy in front-end code rather than nodejs-mobile? This feels like it belongs in the backend.

gmaclennan (commented):
Wow @noffle this is a massive amount of work! I only had 30 mins just now to do a very quick read through and leave initial comments. I was actually here reading through to answer https://github.com/digidem/mapeo-mobile/issues/528 and https://github.com/digidem/mapeo-mobile/issues/529, but it may be until next week that I can do a more thorough review of this, so I thought I would leave the thoughts/comments that came to me on the first read here as a starter. I'll answer your #528 #529 directly now that I have some understanding of the code.

android/app/src/main/java/com/mapeo/MainApplication.java
hackergrrl: `git blame` shows that this was code you wrote (maybe for the `share-apk` branch?). I don't know what any of this does :D

Results
- Camera: yes
- GPS: yes
- Orientation: yes

android/app/src/main/java/com/mapeo/MainApplication.java
hackergrrl: I think these java changes are all because of the drift from the original `share-apk` branch that this branch is based on, and `develop`. I'm going to to restore this code.

src/backend/lib/constants.js
hackergrrl: Good idea! I'll change it over.

src/backend/server.js
hackergrrl: What bugs is this prone to?

src/frontend/api.js
hackergrrl: This is how it was done in the `share-apk` branch you wrote, so I kept it here, since it worked fine.

hackergrrl (commented):
Looking over the full diff, I think there's a bunch of junk code from the `share-apk` branch I based this off of, so I appreciate the "why is this even here?" style comments :)

gmaclennan:
@noffle, sorry my review of this is coming in drips and drabs, but I wanted to get initial thoughts to you rather than delaying to get you everything at once.

As I mentioned on our call right now, I really like how the code is structured here, and that everything is essentially behind the `upgrade-manager.js`, i.e. there is a clear boundary between the upgrade code and the rest of the app. However, with the current patterns of passing `emitFn, listenFn, removeFn` to `UpgradeManager`, the API surface is not very clear.

I think it would be easier for others who develop the app if there is a clear public API to `UpgradeManager`, and we do the wiring up with the nodejs RPC outside it. That way also the implementation of `UpgradeManager` is not tied to how we have implemented communication between frontend-backend.

As I understand it, `UpgradeManager` provides 3 public methods:

1. `getState()`
2. `startServices()`
3. `stopServices()`

And it is an event emitter that emits two types of event:

1. `state`
2. `error`

I think we should write it as such, and then, outside of `upgrade-manager.js` in `src/backend/server.js` write the wiring up code, e.g. 

```js
rnBridge.channel.on('p2p-upgrade::get-state', () => {
  rnBridge.channel.post('p2p-upgrade::state', upgradeManager.getState())
})
upgradeManager.on('state', state => {
  rnBridge.channel.post('p2p-upgrade::state', state)
}
```

That way also the code in `server.js` is responsible for cleaning up these listeners on close (which it already does for other RPC listeners)

gmaclennan:
The other point I mentioned is the copying of the APK. This was only necessary for the testing that I did sharing the APK via the Android Share dialog - I needed to move the APK into a different part of Android storage so that I could add fileshare permissions (that is what the [`filepaths.xml`](https://github.com/digidem/mapeo-mobile/pull/524/files#diff-a029272cc71f7b06d507bc9b3507570edd16da59c9a85c0cac4fba592864b30c) code is about - it is letting Android know that we want the app to be able to share files in these folders with other apps via the Android Share screen - again this is not necessary now)

The nodejs process should be able to read the APK in its original location, so no need for the copy step. Only the frontend knows where this file is, but rather than have a back-and-forth when you start the upgrade manager, I suggest that we pass the APK path as part of the `config` rnBridge call:

https://github.com/digidem/mapeo-mobile/blob/share-apk-p2p/src/backend/index.js#L75
https://github.com/digidem/mapeo-mobile/blob/share-apk-p2p/src/frontend/api.js#L239

That way you can pass the APK path as a parameter when you first create a new `UpgradeManager` instance. I think you could remove the `apkReady` function in `server.js` if you did this.

hackergrrl:
This is great news -- it will simplify things considerably.

hackergrrl:
Agreed, I like this approach.

gmaclennan:
> Agreed, I like this approach.

Ok, great. In terms of the throttle of state events, I think then the place to put it is to throttle the function passed to the event handler `upgradeManager.on('state', throttledFunction)` -- I can do this once you change the API.

gmaclennan:
For discussion: How to handle different version numbers?

Mapeo version names are set at build time, and there are 3 types of releases we make:

1. **"Production" releases**: These have the format `{major}.{minor}.{patch}.{buildNumber}` where `buildNumber` is an integer that is guaranteed to increment for every build (including builds of the other "non-production" releases)
2. **Release Candidates**: Format `{major}.{minor}.{patch}-RC.{buildNumber}`. The `{major}.{minor}.{patch}` portion is the target production release version. Each subsequent build of an RC (e.g. for fixing bugs identified in QA) will keep the same `{major}.{minor}.{patch}` but `buildNumber` will increment.
3. **Internal test builds**: These are built for every PR against the `develop` branch, and they have the version format: `internal.{buildNumber}`. These are used to test a specific feature before a PR is merged.

I think the upgrade logic that makes most sense to me is:

1. Production releases only update to newer production releases. We could use _only_ the build number for this (since it is guaranteed to increment for every new production release), but in the future we might want to use the major/minor/patch in order to have different behaviour based on that (e.g. we might treat a major update differently from a minor in the UX)
2. Release candidate releases should update to newer RCs _for the same `{major}.{minor}.{patch}`_, but if the major/minor/patch changes they should be ignored.
3. Internal test builds should update to a newer internal test build from the same PR, based on the incrementing build number.

Since these version numbers are set as part of our CI build, when testing local builds, the version name will be `0.0.0` by default. In order to build locally with a different version name, set the `ANDROID_VERSION_NAME` env variable, e.g. `ANDROID_VERSION_NAME=1.1.0.0 npm run android` will build an APK where `DeviceInfo.version()` returns `1.1.0.0`.

I think the current version naming scheme is confusing and hard to parse (due to not being clearly defined). I propose that we switch to semver compatible versions, see https://github.com/digidem/mapeo-mobile/issues/554

hackergrrl:
Sounds good to me!

gmaclennan:
@noffle Can we remove this commit from here: https://github.com/digidem/mapeo-mobile/commit/28271b1a0166228bc05847208247bde55d3ccec7 This was just an experiment and the code is unfinished (I was trying to learn how to check signatures on the APK, but installing an incorrectly signed APK will not work anyway).

hackergrrl:
@gmaclennan sure! pushed a revert commit!

src/backend/lib/upgrade-server.js
```diff
@@ -0,0 +1,271 @@
...
+        }
+      }
+
+      function stop() {
+        log("drain: stopping..");
+
+        clearInterval(this.announceInterval);
+        this.announceInterval = null;
+
+        // XXX: 'unannounce' seems to always return an error here; ignoring.
```
gmaclennan, line 125: Yes, I checked the dns-discovery code, and there is a bug where this will always return an error (because `success` is never set to true on the unannounce path when no servers are defined) and will never emit `unannounce`. https://github.com/mafintosh/dns-discovery/blob/master/index.js#L425-L482

src/backend/lib/upgrade-server.js
```diff
@@ -0,0 +1,271 @@
...
+        this.discovery.unannounce(DISCOVERY_KEY, this.port, _ => {
+          log("drain: ..discovery unannounced");
+          this.discovery.destroy(err => {
+            if (err) {
```
gmaclennan, line 129: The destroy callback never has any arguments (`err` will always be undefined) see https://github.com/mafintosh/dns-discovery/blob/master/index.js#L605-L625

src/backend/lib/upgrade-download.js
```diff
@@ -0,0 +1,303 @@
...
+          next(null, chunk);
+        });
+        const ws = this.storage.createApkWriteStream(
+          filename,
+          option.version,
+          option.hash,
+          err => {
+            if (err) this.setState(UpgradeState.Download.Error, err);
+            else this.setState(UpgradeState.Download.Downloaded, null);
+          }
```
gmaclennan: It doesn't seem to make sense to mix a stream and a callback, because we end up with two ways this completes: when the `pump` pipeline completes, and then when this callback is called, which happen at different times. I feel like `createApkWriteStream` should return a writable stream which will `finish` when the APK is written.

gmaclennan (commented):
I've had a chance to more thoroughly review the code in this PR. It took me a while because it is so large and complex.

### State modelling

I think some states are not being modelled, which is a problem because the behaviour in these states is hard to understand from the code and can result in unexpected behavior when things happen suring this missing states.

Specifically: I can see 4 states in the `Search` module:

1. `IDLE`
2. `STARTING`
3. `SEARCHING`
4. `STOPPING`

(and maybe `ERROR` if it is possible for an un-recoverable error to happen)

Currently the code works around not being explicit about these states by using the `rwlock` module, however this makes it unclear about what will happen as a result of any action in these states. E.g. what does the `onPeer` event do if the state is in `STARTING` or `STOPPING`. Right now that requires understanding how the `rwlock` module works and the rules that it follows (parallel reads, but only after writes are complete) and requires knowing the exact order of the history e.g. what happens when onPeer is called after reset is called but before it has started?

Similarly the state model for the `Server` class is missing the `STARTING` state. Again the work-around using `rwlock` makes it hard to reason about what is happening and catch bugs.

I think using `rwlock` as a way to update state makes it hard to reason with the code and introduces the possibility of bugs, due to the difficulty to model what happens in the transition states. It essentially queues up state updates and reads, but in a way that is not that obvious and might not be what we actually want.

Can we explicitly model these transition states, and then we can be explicit in the other methods (e.g. `onPeer` or the server methods) about what happens in these states.

### Version comparison

There are a few places with logic to compare versions: in `upgrade-download.js`, `upgrade-storage.js` and in `upgrade-manager.js` (although this is just a semver compare here). Can we use a single function for comparing versions everywhere, otherwise I see this as being easy to miss if we want to change the comparison logic in the future.

Also the current logic in the `Check.check()` method seems to just randomly choose an available option from storage, rather than the most recent (it filters the available upgrades then picks the one at index `0`).

### Storage and checks

It might be useful to make the `Storage` class stateful, so that it can maintain a state about the available upgrades to avoid the need for the `Check` class.

E.g. if `Storage` had a `availableUpgrades` state, that was emitted every time that it changes e.g.

- After initial startup and read of the disk
- When `createApkWriteStream` successfully completes

This way the upgrade manager would only need to listen for changes in the storage state to know when an upgrade is available, and it would not need the coordination between the `Check` module and the `Storage` module.

gmaclennan:
Superseded by #595

PR #526, opened 2021-03-16
feat: Setting to choose coordinate display format

luandro:
This adds the ability to change the coordinate system, saving to AsyncStorage the selected option.

- [x] Add coordinate system page to settings with ability to change between LatLon, UTM, DMS, DD
- [x] Change coords on observation screens based on setting
- [x] Change coords on GPS screen based on setting


hackergrrl:
I don't feel super qualified to review this, but it looks good to me! If it all seems to be working in your manual tests, I think it's ok to merge now.

cc @gmaclennan to take a look over when he's back.

gmaclennan:
Hi! I'm back! Great work on this @luandro! I haven't installed this or checked the code yet, but based purely on the screenshots a couple of comments:

- "Latitude and Longitude" and "Decimal Degrees" are the same thing. DMS is also Latitude and Longitude. Basically DMS and Decimal Degrees are two ways or writing lat & lon.
- I think we should round decimal degrees to 6 decimal places (see https://wiki.openstreetmap.org/wiki/Precision_of_coordinates). That is about 10cm precision at the equator, which is way more than our GPS precision anyway.
- Just to make sure that we show decimals on DMS coordinates, e.g. 0°00′00.004″. We should show 3 decimal places here to match the same precision as DD.
- There are lots of ways to format / write DMS (see https://github.com/gmaclennan/parse-dms/blob/master/test/index.js) but I think `DD°MM′SS.sss″ N DD°MM′SS.sss″ W` makes the most sense. Latitude is first when writing DMS (most of the time!)
- There is controversy about whether to write decimal degrees as lon/lat or lat/lon. The "traditionalists" say lat/lon, which is "y-axis" first, but most computer code expects Longitude first (e.g. x-axis first). I would lean towards lon first. Not sure what the best way is to format this so that it is clear to the user, could you do some research into ways of formatting / writing decimal degrees?
-

gmaclennan:
Also we should put this data (selected coordinate system) in React Context, and the hook should read from there. I can talk through the reasons for this and how to do this on our next check-in.

luandro:
> "Latitude and Longitude" and "Decimal Degrees" are the same thing. DMS is also Latitude and Longitude. Basically DMS and Decimal Degrees are two ways or writing lat & lon.

lol thanks, was really confused that both looked exactly the same.

> I think we should round decimal degrees to 6 decimal places 

Makes a lot of sense.

> Just to make sure that we show decimals on DMS coordinates, e.g. 0°00′00.004″. We should show 3 decimal places here to match the same precision as DD.

:+1:

 > could you do some research into ways of formatting / writing decimal degrees?

In relation to what comes first (lat or long):
- [answers.com](https://www.answers.com/Q/What_goes_first_Longitude_or_Latitude)
- [gis.stackexchange.com](https://gis.stackexchange.com/questions/7379/why-latitude-longitude-and-not-the-other-way-round)
- [quora.com](https://www.quora.com/Which-came-first-latitude-or-longitude)
- [stackoverflow.com](https://stackoverflow.com/questions/7309121/preferred-order-of-writing-latitude-longitude-tuples-in-gis-services)
- [docs.geotools.org](http://docs.geotools.org/latest/userguide/library/referencing/order.html)

My interpratation:

Lat/Long is commonly used because Latitute was discovered first, as has been used first since.

Long/Lat is also commonly used because of  conventional math (ie, f(x ,y, z)).

I think lat/long will make more sense for the user, while long/lat will make more sense for a machine. Thoughts @gmaclennan, as you have more experience? @rudokemper might also have some thought on this.

rudokemper:
In my experience, latitude is generally always written out first, conforming to your observations on user expectations @luandro. But that might be specific to the contexts where I have done GIS work (Suriname, Brazil, Colombia, and now Kenya). But even experienced GIS specialists get lat & long confused all the time, and have to look up which is which, haha..

src/frontend/lib/utils.js
```diff
@@ -143,6 +158,38 @@ export function formatCoords({
...
+}
+
+export function formatCoords({
+  lon,
+  lat,
+  format = "utm",
+}: {
+  lon: number,
+  lat: number,
+  format?: "utm",
```
gmaclennan:
```suggestion
  format?: "utm" | "latlon" | "dms",
```

luandro:
Todo:

- [x] Add settings context, and have the coordinate config live there
- [x] Rename LatLog to Decimal Degrees
- [x] Pass lat lon as object to coordinate format functions

Forgetting anything @gmaclennan?

src/frontend/lib/utils.js
luandro: :+1:

luandro:
Had a bit of a hard time updating my knowledge on React hooks, but all worked out. Ready for another review @gmaclennan.

src/frontend/context/SettingsContext.js
```diff
@@ -0,0 +1,84 @@
...
+const saveData = async value => {
+  try {
+    const stringified = JSON.stringify(value);
+    await AsyncStorage.setItem(STORE_KEY, stringified);
+    return value;
+  } catch (e) {
+    console.log("Failed to save the data to the storage");
+  }
+};
+
```
gmaclennan: There is already a `usePersistedState` hook, so best use that rather than have two different implementations of this.

src/frontend/context/SettingsContext.js
```diff
@@ -0,0 +1,84 @@
...
+    if (didMountRef.current) {
+      saveData(contextValue[0]);
+    } else {
+      didMountRef.current = true;
+      getData(dispatch);
+    }
+  }, [contextValue]);
+  return (
+    <SettingsContext.Provider
+      value={{ settings: contextValue[0], dispatch: contextValue[1] }}
```
gmaclennan: This creates a new object on every render

src/frontend/lib/utils.js
```diff
@@ -121,15 +121,30 @@ export function getLastPhotoAttachment(
...
+function toDegreesMinutesAndSeconds(coordinate) {
+  const absolute = Math.abs(coordinate);
+  const degrees = Math.floor(absolute);
+  const minutesNotTruncated = (absolute - degrees) * 60;
+  const minutes = Math.floor(minutesNotTruncated);
+  const seconds = (minutesNotTruncated - minutes) * 60;
+  const decimals = 3;
+  const formattedSeconds = Number(
+    Math.round(seconds + "e" + decimals) + "e-" + decimals
+  );
```
gmaclennan: Why not `seconds.toFixed(3)`?

src/frontend/sharedComponents/LocationField.js
```diff
@@ -55,6 +58,7 @@ const LocationField = ({ children }: Props) => {
   return children({
     longitude: value.lon,
     latitude: value.lat,
+    system: coordinateSystem,
```
gmaclennan: This doesn't really belong here, because this component has nothing to do with the display of coordinates. To separate concerns, I think it's better to keep this in the `<LocationView>` component

gmaclennan (commented):
Adding my feedback here that I shared over Slack, so there's a record of it. I've fixed this up myself and I'm just testing locally before pushing.

PR #545, opened 2021-04-15
Add additional config available for UpgradeManager instance

gmaclennan:
I also did a few cleanup tasks here.

This makes the following variables available for the creation of the UpgradeManager instance:

- `apkPath`: Path to currently running APK (file can be read from nodejs)
- `minSdkVersion`: Min Android SDK version supported by currently running APK
- `version`: Version name of running APK (semver if release, but not if testing)
- `buildNumber`: Increments every time CI build happens - must always increase
- `bundleId`: Different variants have different bundleIds
    - com.mapeo
    - com.mapeo.qa
    - com.mapeo.debug
    - com.mapeo.icca
  Different bundle ids install as separate apps

### DIFF
--- a/android/app/src/main/java/com/mapeo/AppInfoModule.java
+++ b/android/app/src/main/java/com/mapeo/AppInfoModule.java
@@ -14,10 +14,6 @@ import java.util.HashMap;
 public class AppInfoModule extends ReactContextBaseJavaModule {
   private static ReactApplicationContext reactContext;
 
-  static final String MODULE_NAME = "AppInfo";
-
-  private static final String sourceDir = "sourceDir";
-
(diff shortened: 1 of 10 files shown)


hackergrrl:
Could you rebase? A bunch of my edits are mixed into the diff.

gmaclennan:
I never managed to rebase this because when I tried yesterday, lots of the code in here is "kind of" in your branch. If needed I can do this work again on top of the current HEAD of `share-apk-p2p`, but let me know when is a good moment so that we don't end up stepping on each others feet again.

hackergrrl:
This should now be superceded by https://github.com/digidem/mapeo-mobile/pull/566

PR #555, opened 2021-04-20
feat: Show alert dialog when Mapeo is updated

### LINKED ISSUE #548: (Front) Display dialogue if user upgraded already
**Description :**
Display a dialogue at start-up if the user has successfully updated the app. 
Have a standard text dialogue as the quick fix. (then do the nicer version later.)

Copy in the text dialogue : TBD

**Acceptance Criteria :**
It is done when a dialog is displayed on first start (but not subsequent starts of the same version) after a new version is installed.

**Technical Notes:**
See #555

Points:
3pts

gmaclennan:
Show an alert dialog when Mapeo is successfully updated. Fixes #548 

The first release that includes this PR will not show anything (since without previously installed version info, it is impossible to tell the difference between an update and the first install, and we do not want to show this alert when the user first installs Mapeo).

Updates after this PR lands will show a dialog to the user letting them know that Mapeo has successfully updated. Requesting review of the language (see image below) from @jencastrodoesstuff @rudokemper @marcartro.

update-notifier

This PR stores a complete history of Mapeo installs, including the time that each update was installed. This information is currently unused, but could be useful for debugging purposes in the future.

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -19,6 +19,18 @@
     "description": "Title of dialog that shows when cancelling a new observation",
     "message": "Discard observation?"
   },
+  "hooks.updateNotifier.dialogMessage": {
+    "description": "Message shown in dialog shown to the user when Mapeo is updated",
+    "message": "Mapeo was successfully updated from version {previousVersion} to {currentVersion} on {updateDateTime}"
+  },
+  "hooks.updateNotifier.dialogOK": {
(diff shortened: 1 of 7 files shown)


jencastrodoesstuff:
As long as translations work for this (for current Spanish and Portuguese speaking partners) I think it is fine until it gets design love down the road.

Issue #549, opened 2021-04-19
(Front) Feature off by default


ximenabb:
**Description :** As part of a soft launch, the team has decided to release the feature in stages. First stage is to set the feature by default and give the option to turn it on.

- Set the feature off by default
- Display a screen explaining the feature + CTA to turn on

**Acceptance Criteria :**
It is done when... - TO ADD

**Technical Notes:**
Notes added by the tech team - TO ADD

Points:
5pts

luandro:
Isn't this the same as #541?

ximenabb:
> Isn't this the same as #541?
you are correct!

luandro:
![toggle_p2p_updates](https://user-images.githubusercontent.com/693728/115451101-24b68b80-a1f3-11eb-8e29-a63b009114e8.gif)

Is this what we're aiming for?

PR #561, opened 2021-04-21
Opt-in setting for p2p upgrades

luandro:
Solves #549 

- [x] Cleanup old UI on settings page
- [x] Create "Experiments" page and add route
- [x] Create SelectMultiple component based on `src/screens/ObservationDetails/SelectMultiple.js`
- [x] Update settingsContext state on select change
- [x] Update logic to only start and show p2p-upgrades when it's enabled

### DIFF
--- a/src/frontend/AppStack.js
+++ b/src/frontend/AppStack.js
@@ -22,6 +22,7 @@ import ProjectConfig from "./screens/Settings/ProjectConfig";
 import AboutMapeo from "./screens/Settings/AboutMapeo";
 import LanguageSettings from "./screens/Settings/LanguageSettings";
 import CoordinateFormat from "./screens/Settings/CoordinateFormat";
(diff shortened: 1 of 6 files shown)


luandro:
experimenta_features

gmaclennan (approved):
Good work on this @luandro. I rebased on the latest base branch, and I added my "comments" as commits to fix a few small things. (see Git comments for details).

The integration to turn this on/off is a little fragile I think (I see you needed to modify a few different useEffect hooks, and that kind of code is vulnerable to bugs), and that to me points to a need to adjust how this p2p upgrade component is integrated. I am going to address that in a different PR, which will simplify turning this experiment on/off.

I say we should merge this now, and then refactor the p2p upgrade component in a separate PR.

hackergrrl:
Deferring to @gmaclennan on this PR. :+1:

PR #566, opened 2021-04-21
Add additional config available for UpgradeManager instance

hackergrrl:
Recreated version of https://github.com/digidem/mapeo-mobile/pull/545, fixed for compatibility with `share-apk-p2p`.

---

I also did a few cleanup tasks here.

This makes the following variables available for the creation of the UpgradeManager instance:

- `apkPath`: Path to currently running APK (file can be read from nodejs)
- `minSdkVersion`: Min Android SDK version supported by currently running APK
- `version`: Version name of running APK (semver if release, but not if testing)
- `buildNumber`: Increments every time CI build happens - must always increase
- `bundleId`: Different variants have different bundleIds
    - com.mapeo
    - com.mapeo.qa
    - com.mapeo.debug
    - com.mapeo.icca
  Different bundle ids install as separate apps

### DIFF
--- a/src/backend/index.js
+++ b/src/backend/index.js
@@ -36,7 +36,6 @@ const releaseStage = prereleaseComponents
 const log = debug("mapeo-core:index");
 const PORT = 9081;
 const status = new ServerStatus();
-let storagePath;
 let server;
 
 // This is nastily circular: we need an instance of status for the constructor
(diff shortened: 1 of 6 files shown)


hackergrrl:
@gmaclennan Could you give this a look-over to make sure it's accurate?

src/backend/index.js
```diff
@@ -68,26 +67,27 @@ status.startHeartbeat();
...
+    status.setState(constants.ERROR, { error, context: "config" });
+    return;
+  }
   try {
     server = createServer({
       privateStorage: rnBridge.app.datadir(),
-      sharedStorage: storagePath,
       apkPath: config.apkPath,
       apkVersion: config.apkVersion,
+      sharedStorage: config.storagePath,
```
gmaclennan:
```suggestion
```

These are not named the same now, and easier if we only maintain this in createServer() and `src/frontend/api.js` rather than having to keep change this also if the config shape changes

gmaclennan (changes requested):
👍 All looks good, one fix needed.

PR #573, opened 2021-04-22
[DRAFT] Error handling

hackergrrl:
Hey @gmaclennan, this is a discussion / notes PR (not to be merged!) for the various places errors rise up out of the p2p upgrades backend code. See the diff for notes!

```
// DEFINITIONS of errors, for the context of this diff
//
// Fatal: upgrade feature won't work anymore
// Warning: worth showing a dialog to inform the user
// Ignorable: not even worth telling the user about
```

### DIFF
--- a/src/backend/lib/upgrade-download.js
+++ b/src/backend/lib/upgrade-download.js
@@ -164,6 +164,7 @@ class Search extends EventEmitter {
 
       this.discovery.destroy(err => {
         if (err) {
+          // Ignorable error
           this.setState(UpgradeState.Search.Error, err);
           searchLog("..stopped, with error:", err);
         } else {
@@ -236,6 +237,11 @@ class Download extends EventEmitter {
           option.version,
           option.hash,
           err => {
+            // Warning: worth showing a dialog box
+            // THOUGHT: it would be nice if there was a way for this component to, like, reset itself?
+            //   some options:
+            //     1. set to state Error, like we do here, then, after a timeout, set to Idle again
+            //     2. expose a 'reset()' function that the frontend can manually call that resets the entire upgrade system
             if (err) this.setState(UpgradeState.Download.Error, err);
             else this.setState(UpgradeState.Download.Downloaded, null);
           }
@@ -276,6 +282,8 @@ class Check extends EventEmitter {
 
     checkLog("checking for an upgrade..");
     this.storage.getAvailableUpgrades((err, options) => {
+      // Seriously fatal error, if upgrade storage ever becomes corrupted
+      // in the future this might warrant a way for the frontend to purge storage to reset it?
       if (err) return this.setState(UpgradeState.Check.Error, err);
       options = options.filter(o => this.isUpgradeCandidate(o));
       if (options.length > 0) {
--- a/src/backend/lib/upgrade-manager.js
+++ b/src/backend/lib/upgrade-manager.js
@@ -1,3 +1,9 @@
+// DEFINITIONS of errors, for the context of this diff
+//
+// Fatal: upgrade feature won't work anymore
+// Warning: worth showing a dialog to inform the user
+// Ignorable: not even worth telling the user able
+
 const UpgradeStorage = require("./upgrade-storage");
 const UpgradeServer = require("./upgrade-server");
 const UpgradeDownload = require("./upgrade-download");
@@ -108,6 +114,8 @@ class UpgradeManager {
     const result = validate(this.state);
     if (result) {
       const msg = result.join(",");
+      // Validation error: this indicates an internal error that should never happen.
+      // If it does: definitely a fatal error.
       this.emit("p2p-upgrade::error", { message: msg });
     } else {
       this.sanitizeState();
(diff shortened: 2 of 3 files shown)


src/backend/lib/upgrade-download.js
```diff
@@ -236,6 +237,11 @@ class Download extends EventEmitter {
           option.version,
           option.hash,
           err => {
+            // Warning: worth showing a dialog box
+            // THOUGHT: it would be nice if there was a way for this component to, like, reset itself?
+            //   some options:
+            //     1. set to state Error, like we do here, then, after a timeout, set to Idle again
+            //     2. expose a 'reset()' function that the frontend can manually call that resets the entire upgrade system
```
gmaclennan, line 244: I think this is an error "event" rather than an error state. As I understand it, if there is an error when downloading, then the state of the downloader returns to `IDLE`, ready to download again? e.g. and error event is what causes the state to transition from `DOWNLOADING` to `IDLE`. Or should an error event trigger a retry?

If this is implemented as an event, then the front-end can decide whether to retain a UI state for a period of time after the error occurs. I think the simplest initial implementation is to show a dialog for the error, which is displayed until the user dismisses it.

My suggestion: Do not capture this as an error state, but emit an error event when this happens.

src/backend/lib/upgrade-download.js
```diff
@@ -276,6 +282,8 @@ class Check extends EventEmitter {
 
     checkLog("checking for an upgrade..");
     this.storage.getAvailableUpgrades((err, options) => {
+      // Seriously fatal error, if upgrade storage ever becomes corrupted
+      // in the future this might warrant a way for the frontend to purge storage to reset it?
```
gmaclennan, line 286: Ok, so if the updater is in this state, then we consider the updater to be broken, until we have a way for the user to resolve this.

src/backend/lib/upgrade-manager.js
```diff
@@ -108,6 +114,8 @@ class UpgradeManager {
     const result = validate(this.state);
     if (result) {
       const msg = result.join(",");
+      // Validation error: this indicates an internal error that should never happen.
+      // If it does: definitely a fatal error.
```
gmaclennan, line 118: Ok so I think this should be considered an error state then. The UpgradeManager is in an error state once this happens. However, I feel like we are doing runtime validation of something that should be a test or a type check. Does any input of `this.state` come from userland (e.g. read from somewhere external) or is state always generated by the code here, e.g. if this doesn't validate is it because there is a bug in this code, or could it be bad data getting into here somehow?

src/backend/lib/upgrade-server.js
```diff
@@ -83,6 +83,7 @@ class UpgradeServer extends EventEmitter {
     this.stateLock.writeLock(release => {
       const self = this;
       function done(err) {
+        // Ignorable. this only fires if discovery.destroy() errors, which is fine
```
gmaclennan, line 86: If this error does not mean that the server is no longer is either the `IDLE` or `SEARCHING` state, then I don't think we should change the state here. What is the consequence of this error?

src/backend/lib/upgrade-download.js
```diff
@@ -164,6 +164,7 @@ class Search extends EventEmitter {
 
       this.discovery.destroy(err => {
         if (err) {
+          // Ignorable error
```
gmaclennan, line 167: I think this means that the Search state is in `IDLE` right, but there was an error event?

gmaclennan (commented):
Thanks for this @noffe. I think several of these errors are error events that trigger a transition to a different state, but do not result in an error state. I think some of these only need to be logged for debugging, but the download fail event should be shown to the user (e.g. by emitting an event that the front-end can handle in its own way).

src/backend/lib/upgrade-download.js
hackergrrl: > As I understand it, if there is an error when downloading, then the state of the downloader returns to IDLE, ready to download again?

Currently an error in a component will put in eternally in the `"Error"` state. For the case of a failed download, we could emit an `"error"` event instead though and transition back to `"Idle"`.

src/backend/lib/upgrade-manager.js
hackergrrl: Each component (`Server, Search, Download, Check`) emits its own `{state, context}`, which parent components (`UpgradeDownload, UpgradeManager`) compose into a larger state object.

hackergrrl: > if this doesn't validate is it because there is a bug in this code, or could it be bad data getting into here somehow?

Only code bugs should be able to cause an error here.

src/backend/lib/upgrade-server.js
hackergrrl: > If this error does not mean that the server is no longer is either the IDLE or SEARCHING state, then I don't think we should change the state here.

I don't understand what you mean by this.

> What is the consequence of this error?

Unknown. Worst case, it's possible that the server is still advertising itself & responding to mDNS requests.

src/backend/lib/upgrade-download.js
hackergrrl: Right now the components goes into its permanent `"Error"` state. There are no `"Error"` events in the upgrade system.

PR #574, opened 2021-04-26
P2pUpdate component refactor

gmaclennan:
Depends on: #561 

Attempt to fix a few issues in the current implementation of the p2p upgrade frontend code

**Encapsulation of state logic**
Work on #561 highlighted that the way the p2p upgrades is integrated into the existing code makes it difficult to turn on and off. The other consequence of the existing implementation is that any change in p2p upgrade state also causes an unnecessary re-render of the sync screen.

This PR moves all of the p2p upgrade logic into a the `UpgradeBar` component, so that everything is encapsulated into one file/component. The code in Mapeo Mobile did initially separated "state" logic from "view" components, however I've found this approach looses the advantage of encapsulating logic into components, and ends up with logic "leaking" into multiple places in the code, so now I prefer to encapsulate logic and view into a single component.

**`fakeTrigger` Hack**
React will update when state updates, so this should not be needed, and if it is needed it suggests a bug.

**Avoid derived state from another state**
The function `getFrontendStateFromUpgradeState` will update the `upgradeInfo` state in the UpgradeBar when the `backendState` updates. Creating one state derived from another can lead to bugs (see https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-unconditionally-copying-props-to-state for a description of a similar problem).

Currently `upgradeInfo` will be updated with a derived value from `backendState` every time `backendState` changes, but we also try to `setUpdateInfo()`, so it's not clear which state would "win".

**DRY styled components**
The same styles for the upgrade bar are copied into all the versions for different states. This leads to a lot of repeated code, and can be simplified.

**Avoid creating component at render**
Passing a IIFE wrapped around the switch statement creates a new React component on every render, and will break React's diffing, resulting in poor render performance.

**Move backend interaction into `api.js`**
All interaction with the backend is currently in api.js. This gives us a single layer of communication with the backend, so if we change how that communication happens, we only change it in one place. It also makes it easier to mock `api.js` for testing.

**Simplify startup back-and-forth**
There is no need for the "front-end-ready", "back-end-ready" to-and-fro, this can be simplified.

- [x] Move `UpgradeBar` into its own file
- [x] Move p2p Upgrade state logic into `UpgradeBar` component
- [x] Simplify "experiments" code to not render `UpdateBar` when experiment is off
- [x] Remove fakeTrigger hack
- [x] Avoid using "derived state"
- [ ] Re-use styles for upgrade bar
- [ ] Avoid IIFE in switch
- [x] Move backend interaction into `api.js`
- [x] Simplify startup back-and-forth


src/frontend/screens/SyncModal/UpgradeBar.js
```diff
@@ -0,0 +1,458 @@
...
+  }
+  if (state.downloader.check.state === BackendUpgradeState.Check.Error) {
+    return {
+      state: "GenericError",
+      context: state.downloader.check.context.message,
+    };
+  }
+
+  // Edge case I've (kira) seen once that shouldn't happen.
+  if (state.downloader.search.state === BackendUpgradeState.Search.Idle) {
```
hackergrrl, line 363: I've since realized that this IS normal (I see it often), and we don't need to treat this as an error.

hackergrrl (commented):
What happens when the user hits the Back button when in the sync view? It unmounts all components, which calls the cleanup function for each effect? Is this also what happens when the user backgrounds the app?

gmaclennan:
> What happens when the user hits the Back button when in the sync view? It unmounts all components, which calls the cleanup function for each effect? Is this also what happens when the user backgrounds the app?

When the user hits the back button, yes, it unmounts all components, calling the cleanup function for each effect. Unfortunately when the user backgrounds the app, it does not unmount any components. For the sync logic, the backend listens for the background event, and closes the server. We could listen for the background event on the front-end and try unmounting when it happens (so that cleanup happens) but I will need to test that the unmount actually happens (not sure if UI updates happen in the background).

src/frontend/screens/SyncModal/UpgradeBar.js
hackergrrl: I've fixed this on my latest PR fyi #580

Issue #578, opened 2021-04-28
(back) Determine update candidate based on new version naming scheme


gmaclennan:
Currently we have 3 types of APK releases, each with their own version name format:

1. **Production Releases**: Standard semver e.g. `{major}.{minor}.{patch}`.
2. **Release Candidates**: `{major}.{minor}.{patch}-RC+{sha7}`. Here `{major}.{minor}.{patch}` refers to the _pending_ release. `sha7` is the Git commit sha used for the build, shortened to 7 characters.
3. **Internal test builds**: `{major}.{minor}.{patch}-PR{pullRequestNumber}+{sha7}` where `pullRequestNumber` is the number of the PR that the build is built from, `sha7` as above. Here `{major}.{minor}.{patch}` would be most recently published production version that has been merged back into the `develop` branch.

The p2p updater should consider an APK an "update" if:

1. **Production Releases**: Should update to the next semver production release, but not a release candidate or internal test build. If an update with the same version is available, but a higher build number, then we should update. This is an edge case that can happen as a result of a failed build, but is unlikely to happen in the wild (we normally catch this early before the app is distributed widely) so we can skip implementing this for now.
2. **Release Candidates**: Should update to the next RC that shares the same `{major}.{minor}.{patch}`, and any production release that is semver => `{major}.{minor}.{patch}` of the RC. The versionCode / build number should be used to determine updates if the two RC releases have the same version (but different `+{sha7}`). The higher build number "wins".
3. **Internal test builds**: Should update to anything that matches `{major}.{minor}.{patch}-PR{pullRequestNumber}`, again using the versionCode / build number.

### Examples:

1. `1.0.0` should update to `1.0.1` and `1.2.0` but not to `1.0.0-RC`
2. `1.1.0-RC+9a6731e` should update to `1.1.0` and to `1.1.0-RC+19d986e` _if_ the build number of the latter is higher.
3. `1.0.0-PR571+eac5a0f` should update to `1.0.0-PR571+afc1a1b` _if_ the build number of the latter is higher. It should not update to `1.0.0-PR572+4667afc` or any other version that does not match `/$1\.0\.0-PR571/`.

See https://github.com/digidem/mapeo-mobile/issues/554#issuecomment-823649801 for an example implementation.

PR #580, opened 2021-04-30
Support downloading subsequent newer versions of the app

### LINKED ISSUE #540: (Front) UI State when new dl begins
**Description :**
have the ui state reflect when a new dl begins, after a previous upgrade has been downloaded

**Acceptance Criteria :**
TO ADD

**Technical Notes:**
TO ADD

**Points:**
3pts

hackergrrl:
Fixes #540.

This branch adds support for downloading multiple upgrades in a row. Previously the upgrade state machine would download at most **one** upgrade before locking itself into a state where it would download no more unless the app was restarted.

### DIFF
--- a/src/backend/lib/upgrade-download.js
+++ b/src/backend/lib/upgrade-download.js
@@ -220,7 +220,7 @@ class Download extends EventEmitter {
 
   reset() {
     if (this.state !== UpgradeState.Download.Downloaded) return;
-    this.setState(UpgradeState.Download.Idle);
+    this.setState(UpgradeState.Download.Idle, null);
   }
 
   download(option) {
@@ -258,7 +258,6 @@ class Download extends EventEmitter {
         pump(res, progress, ws, err => {
           if (err) downloadLog("download pipeline error", err);
           else downloadLog("download pipeline ended ok");
-          this.setState(UpgradeState.Download.Idle, null);
         });
       })
       .once("error", err => {
--- a/src/backend/lib/upgrade-download.js
+++ b/src/backend/lib/upgrade-download.js
@@ -48,7 +48,6 @@ class UpgradeDownloader extends EventEmitter {
       // Check for a new version if the downloader gets set back to Idle.
       if (state === UpgradeState.Download.Idle) {
         this._check.check();
-        this._search.reset();
       }
     });
     this._check.on("state", (state, context) => {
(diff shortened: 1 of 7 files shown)


src/backend/lib/upgrade-download.js
```diff
@@ -181,6 +183,12 @@ class Search extends EventEmitter {
     this.emit("state", state, context);
   }
 
+  reset() {
+    if (this.state !== UpgradeState.Search.Searching) return;
```
gmaclennan: My assumption would be that if the Search is in the `IDLE` state, then `reset()` would start the search (e.g. transition the Search state to `SEARCHING`), but `reset()` will only work if it is already `SEARCHING`. Is this how we want it to work? Should be documented for any future maintainers.

gmaclennan (changes requested):
I think the original issue is because `DOWNLOADED` is not a state, it is an event. As far as I can see, we don't read or use this state anywhere. I think there are two possible states of the `Download` class:

1. `IDLE`
2. `DOWNLOADING`

The `download()` method transitions from `IDLE` --> `DOWNLOADING` and the events `error` or `download-complete` transition from `DOWNLOADING` --> `IDLE`. (I know we don't model those events as EventEmitter events, I'm referring them to abstract events caused by either the download failing or the download completing successfully).

In the future we might add a `cancelDownload()` method which would similarly transition `DOWNLOADING` --> `IDLE`.

I think the need for this reset method comes from modelling `DOWNLOADED` as a state, and what this PR does is essentially ensure that that state only exists for a single tick, because it immediately transitions that state to `IDLE`, so do we need it in the first place? It seems to add code complexity.

What does the search reset do? If the search is in the `SEARCHING` state does stopping and starting it force it to start finding things again?

hackergrrl:
>  I think there are two possible states of the Download class:   IDLE    DOWNLOADING

Ooh, I like this. I wrote this patch as a hack to try and fix a bug without *really* thinking through the bigger picture. I'll revise to include this idea.

> I think the need for this reset method comes from modelling DOWNLOADED as a state, and what this PR does is essentially ensure that that state only exists for a single tick, because it immediately transitions that state to IDLE, so do we need it in the first place? It seems to add code complexity.

The *actual* reason (which wasn't made obvious by just this diff) is that download options accumulate in the state's `context` object, which was confusing the logic in UpgradeManager, which decides which upgrade to download. I've revised how this works, so now the reset() logic isn't necessary.

The most recent changes also include a fix where, if there is a redownload (or multiple upgrade options in storage), the newest one wasn't being selected.

gmaclennan (approved):
Ok on a quick read-through this looks good, we can catch any issues we might have missed in QA on the share-apk-p2p branch. To address (not in this PR): put all the version compare logic in one place, to make it easier to implement https://github.com/digidem/mapeo-mobile/issues/578

PR #583, opened 2021-05-05
Enable Typescript on backend via JSDoc

hackergrrl:
As proposed in https://github.com/digidem/mapeo-desktop/issues/467

### DIFF
--- a/src/backend/tsconfig.json
+++ b/src/backend/tsconfig.json
@@ -1,5 +1,8 @@
 {
   "compilerOptions": {
+    "target": "es2019",
+    "lib": ["es2020"],
+    "resolveJsonModule": true,
     "allowJs": true,
     "checkJs": false,
     "noEmit": true,
--- a/src/backend/tsconfig.json
+++ b/src/backend/tsconfig.json
@@ -5,6 +5,6 @@
     "noEmit": true,
     "typeRoots": ["types", "node_modules/@types"]
   },
-  "include": ["*.js", "test/**/*"],
+  "include": ["**/*.js"],
   "exclude": ["node_modules/"]
 }
--- a/src/backend/tsconfig.json
+++ b/src/backend/tsconfig.json
@@ -3,6 +3,7 @@
     "target": "es2019",
     "lib": ["es2020"],
     "resolveJsonModule": true,
+    "moduleResolution": "node",
     "allowJs": true,
     "checkJs": false,
     "noEmit": true,
--- a/src/backend/tsconfig.json
+++ b/src/backend/tsconfig.json
@@ -1,0 +1,10 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": false,
+    "noEmit": true,
+    "typeRoots": ["types", "node_modules/@types"]
+  },
+  "include": ["*.js", "test/**/*"],
+  "exclude": ["node_modules/"]
+}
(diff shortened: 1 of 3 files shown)


src/backend/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "checkJs": false,
+    "noEmit": true,
+    "typeRoots": ["types", "node_modules/@types"]
```
gmaclennan, line 10:
```suggestion
    "target": "es2019",
    "lib": ["es2020"],
    "resolveJsonModule": true,
    "allowJs": true,
    "checkJs": false,
    "noEmit": true,
    "typeRoots": ["types", "node_modules/@types"]
```

`resolveJsonModule` ensures that requiring `.json` files still works (off by default)
`lib` tells Typescript what type definitions to include. We run Node v12.16.3 on mobile, which supports all ES2020 _modules_ (see https://stackoverflow.com/a/59787575/3071863)
`target` is not actually used, since we don't use Typescript to compile, but just adding it in case we turn that on. Node v12 only supports 100% es2019 syntax, not all es2020.

src/backend/tsconfig.json
```diff
@@ -0,0 +1,10 @@
...
+  },
+  "include": ["*.js", "test/**/*"],
```
gmaclennan:
```suggestion
  "include": ["**/*.js"],
```
I think we want this to include all js files in `src/backend` right? Especially since it's opt-in via `// @ts-check`. Can drop the `test` option, since that's included in the first now.

src/backend/tsconfig.json
```diff
@@ -0,0 +1,10 @@
...
+  "exclude": ["node_modules/"]
```
gmaclennan:
```suggestion
  "exclude": ["node_modules"]
```
I _think_ that's the correct syntax, at least that's what I've always used.

gmaclennan (approved):
That's great Kira, added some changes (with explanations) to the config, but this should work (I haven't tested it locally yet).

gmaclennan (changes requested):
`package-lock.json` is out-of-sync (it's not been updated with this typescript addition). Can you re-install and update package-lock?

hackergrrl:
Thanks for deciphering all of the tsconfig bits :)

gmaclennan:
> Thanks for deciphering all of the tsconfig bits :)

oh that was one of the hardest parts of starting to use Typescript! There's more to deal with when compiling `.ts` to `.js` too, but we don't have to worry about that when writing js-doc comments.

hackergrrl:
Merging.

@gmaclennan In the future can you either not review /w the "prevent merging" bit set, or unset it when I resolve the issues?

gmaclennan:
Ah yes, my apologies, I'm not sure how to do that unless you re-request review. I think the "request changes" is useful to indicate that something needs to change, so perhaps the best workflow is:

1. Author requests review
2. if in review it needs changes, reviewer marks as "Request changes"
3. Author requests review again
4. Reviewer approves changes

hackergrrl:
@gmaclennan If we don't use the review option that blocks merging, we can have a social contract that says we merge only once all items are resolved. That will let us skip steps 3 + 4, which, given our timezones, can easily add an extra day to getting PRs in.

PR #586, opened 2021-05-10
fix: Fix ApkInstaller for Android <7 (SDK version <24)

### LINKED ISSUE #582: (Android) APK install not working for Android <7
The Java [code we're using](https://github.com/digidem/mapeo-mobile/blob/share-apk-p2p/android/app/src/main/java/com/mapeo/ApkInstallerModule.java) for installing apks throws error on my Android 6 device:
```
W PackageInstaller: Parse error when parsing manifest. Discontinuing installation
```
This could be a permission problem for Android, as on Android 6 they are not asked for:
mapeo_permissions

But there's no way to give permission through the settings and it's confusing if the has the permissions or not.

### Paths foward:

- Check [original package](https://github.com/nodece/react-native-apk-installer-n/blob/master/android/src/main/java/com/nodece/apkinstallern/RNApkInstallerNModule.java) for any updates on the code
- Reactivate [similar issue](https://github.com/nodece/react-native-apk-installer-n/issues/3) on original repo
- Compare code from [react-native-apk-installer](https://github.com/null--/react-native-apk-installer/blob/master/android/src/main/java/com/cnull/apkinstaller/ApkInstallerModule.java) which I remember worked on my phone many years ago

Other ideas are welcome!

gmaclennan:
Fixes #582 

It turns out that for Android <7, the APK and any parent folders need to have their file permissions set be world-readable. E.g. we store the APK in `[INTERNAL_FILES_DIR]/installer/myApk.apk`. To be able to install this, `myApk.apk` needs to be world-readable and the `installer` folder needs to be world-executable.

We _could_ set these permissions when creating the `installer` folder, but since these permissions are only needed on Android <7 I think it's better to leave the folder with more restrictive permissions, and set the permissions in the Java code which only runs for Android <7.

Ideally the Java code that sets permissions would set all parent folders up to `[INTERNAL_FILES_DIR]` to be world-executable, but my Java knowledge it not good enough, so for now it assumes the APK is in a folder one-level deep, e.g. it only sets permissions on the parent folder.

@luandro can you test this on your phone and see if it now works as expected?

### DIFF
--- a/android/app/src/main/java/com/mapeo/ApkInstallerModule.java
+++ b/android/app/src/main/java/com/mapeo/ApkInstallerModule.java
@@ -1,10 +1,12 @@
 package com.mapeo;
 
 import android.content.Intent;
+import android.content.ActivityNotFoundException;
 import android.app.Activity;
 import android.net.Uri;
 import android.os.Build;
 import androidx.core.content.FileProvider;
+import android.util.Log;
 
 import com.facebook.react.bridge.ActivityEventListener;
 import com.facebook.react.bridge.BaseActivityEventListener;
@@ -21,8 +23,10 @@ import java.io.File;
 public class ApkInstallerModule extends ReactContextBaseJavaModule {
 
   private final ReactApplicationContext reactContext;
+  private static final String TAG = "ApkInstaller";
   private static final int INSTALL_REQUEST = 3033;
   private static final String E_ACTIVITY_DOES_NOT_EXIST = "E_ACTIVITY_DOES_NOT_EXIST";
+  private static final String E_FILE_DOES_NOT_EXIST = "E_FILE_DOES_NOT_EXIST";
   private static final String E_INSTALL_CANCELLED = "E_INSTALL_CANCELLED";
   private static final String E_INSTALL_FAILED = "E_INSTALL_FAILED";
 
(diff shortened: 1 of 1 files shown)


gmaclennan:
FYI the changes to the Java code here are mainly copied from F-Droid from:

- https://gitlab.com/cde/fdroidclient/-/blob/master/app/src/main/java/org/fdroid/fdroid/installer/ApkFileProvider.java
- https://gitlab.com/cde/fdroidclient/-/blob/master/app/src/main/java/org/fdroid/fdroid/installer/DefaultInstallerActivity.java

I also borrowed some code from https://github.com/mikehardy/react-native-update-apk

luandro:
Great job @gmaclennan, it's working now!

I was going to point out the code for installing apk from [react-native-send-intent](https://github.com/lucasferreira/react-native-send-intent/blob/master/android/src/main/java/com/burnweb/rnsendintent/RNSendIntentModule.java#L416), but you handled it.

I suppose we're just missing:
- [ ] cleaning up logs
- [ ] not changing permissions when Android > 6

gmaclennan:
rebased on `share-apk-p2p`.

> I suppose we're just missing:
> 
> * [ ]  cleaning up logs

@luandro not sure what you mean. I added logging intentionally for any errors. I think it should be left in.

> * [ ]  not changing permissions when Android > 6

This is already here - the permission change only happens if `Build.VERSION.SDK_INT < Build.VERSION_CODES.N`

I think this is good to merge, do you agree @luandro?

luandro:
The logging makes sense @gmaclennan. I was reading the `if` statement wrong, you're right. Ok all good to merge.

hackergrrl:
🚀

PR #591, opened 2021-05-11
fix: Fix self-discovery and repeat discovery

gmaclennan:
**Do not merge**: This PR is based on `p2p-update/merged` which is a temporary branch that merges `p2p-update/fix-progress-throttle` and `issue-536-2` into `share-apk-p2p`.

When I was testing all these fixes together and did some QA testing with a real device and the upgrade simulator, I noticed a few things in the logs:

1. Each device is repeatedly querying itself (e.g. `GET /list` for itself)
2. When two devices discover each other they repeatedly call `GET /list` on each other, every <100ms (although sometimes up to 2s)

I discovered that (1) is due to having two instances of `dns-discovery`. Each instance has logic to de-dupe announcements from itself, but since we have two instances (one in the downloader and one in the server) then it (correctly) identifies the announcements from the same device as coming from somewhere else, so the `onPeer` handler is fired.

I found that (2) is due to `discover.announce()` being repeatedly called on 2 second intervals. This appears to gradually increase the number of times that `onPeer` is called, and it seems like it is called more than every 2 seconds.

The commit in this PR fixes this with two changes:

1. Create a single instance of `dns-discovery` that is passed to the `UpgradeDownloader` and the `UpgradeServer`.
2. Only call `discovery.announce()` once, removing the `setInterval()` call.

I understand that the setInterval() might have existed to resolve a discovery bug, but that might have been a result of having two instances. In testing this PR I found that discovery works well with the single `announce()` even if the second device does not start until well after the first device calls `announce()`.

This all seems to greatly reduce the amount of HTTP requests and network traffic, and it still seems to work as expected.

Tests, however, are broken, because this changes some of the behaviour or `UpgradeDownloader` and `UpgradeServer`. Before changing the tests I wanted to get @noffle's opinion on whether this PR is a good idea and if you think this fixes things.

### DIFF
--- a/src/backend/lib/upgrade-download.js
+++ b/src/backend/lib/upgrade-download.js
@@ -2,7 +2,6 @@ const semver = require("semver");
 const http = require("http");
 const collect = require("collect-stream");
 const EventEmitter = require("events").EventEmitter;
-const dns = require("dns-discovery");
 const RWLock = require("rwlock");
 const pump = require("pump");
 const searchLog = require("debug")("p2p-upgrades:search");
@@ -20,10 +19,10 @@ const PROGRESS_THROTTLE = 400; // milliseconds
 // subcomponents.
 class UpgradeDownloader extends EventEmitter {
   // UpgradeStorage -> Void
-  constructor(storage) {
+  constructor({ storage, discovery }) {
     super();
 
-    this._search = new Search(storage);
+    this._search = new Search({ storage, discovery });
     this._check = new Check(storage);
     this._download = new Download(storage);
     this._storage = storage;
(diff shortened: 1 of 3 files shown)


hackergrrl:
Sorry -- I've known about this since I implemented it, & had been meaning to file an issue for a while now for "prevent p2p upgrades from discovering itself". Fortunately I don't think it adds enough overhead to block the initial feature release.

> 1. Create a single instance of dns-discovery that is passed to the UpgradeDownloader and the UpgradeServer.

Nice!

> Only call discovery.announce() once, removing the setInterval() call.

I don't recommend this: I added the interval to improve discovery. I think the issue was that `announce()` just sends a UDP message over multicast, which other devices can easily miss via packet loss. The interval was essential for reliable discovery.

Issue #576, opened 2021-04-28
(Back) Delete old upgrade binaries


hackergrrl:
Right now, after an APK is downloaded and installed, it remains in the `upgrades/` directory (see `backend/lib/upgrade-storage.js`) forever. Since our binaries are >60mb each, this could quickly add up.

The work here would be to have `UpgradeStorage` check for obsolete upgrades on start-up, and purge them. This could be done synchronously, or async with locks to avoid race conditions on the underlying storage.

luandro:
@noffle @gmaclennan quick input on the plan for doing this is welcome.

I think doing it synchronously won't be much of a bottleneck since we'll just be reading and deleting.

- [ ] Read `storageDir` and check if there are more then one binary files
- [ ] If there are, check for the latest (is there code ready and tested for this?
- [ ] Delete other files that aren't the latest apk file

gmaclennan:
@luandro I think it's important that this is async - we shouldn't underestimate how long it might take to delete 60Mb+ archives on an older phone, and we don't want to hold up the node process for that (it would block things like sync and indexing).

Other than that I think the plan looks good, although I think the delete logic should be to delete any APK that is <= the installed version. There is no need to keep around an APK that is already installed, even if it is the latest.

hackergrrl:
I think it would be simpler and safer for this to be synchronous:

1. no chance of race conditions against the upgrade metadata on disk
2. easier to implement
3. easier to reason about

If we don't make it synchronous, I think it would be best to wrap UpgradeStorage operations in a read-write lock to avoid race conditions.

hackergrrl:
> If there are, check for the latest (is there code ready and tested for this?)

You'll want to build a list of all downloaded upgrades that are older or equal to the current version (`this.version` within UpgradeStorage). You can compare versions using `semver.{gt,lt,gte,lte}()`. You can fetch what's on disk with `this.localUpgrades.get((err, options) => {}`.

luandro:
@hackergrrl 
It seems `getAvailableUpgrades.get` and `getAvailableUpgrades.set` will force us to have to use async, or create sync functions that do the same, which might ignore the locks?

I'm writing the cleanup function on Storage's `constructor`, as there are also `mkdirp.sync` and `rimraf.sync` there, is that the right place? If we really need async, should we run the cleanup from `UpgradeManager` instead?

hackergrrl:
[...]

PR #592, opened 2021-05-13
Delete old binaries

luandro:
Solves #576

- [x] Create init function on upgrade-storage that does cleanup
- [x] Add RWLocks to other methods in upgrade-storage
- [ ] Write test to verify cleanup is happening

### DIFF
--- a/src/backend/lib/upgrade-storage.js
+++ b/src/backend/lib/upgrade-storage.js
@@ -7,6 +7,7 @@ const through = require("through2");
 const readonly = require("read-only-stream");
 const pump = require("pump");
 const semver = require("semver");
+const RWLock = require("rwlock");
 const LocalUpgradeInfo = require("./local-upgrade-info");
 const log = require("debug")("p2p-upgrades:storage");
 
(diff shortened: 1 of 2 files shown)


src/backend/lib/upgrade-storage.js
```diff
@@ -42,13 +43,44 @@ class Storage {
     log("..done. recreating temp directory..");
     mkdirp.sync(this.tmpdir);
     log("..done");
-
+    this.lock = new RWLock();
+    log("deleting old apks");
+    this.lock.writeLock(release => {
+      function done(err) {
+        if (err) throw new Error(err);
```
hackergrrl, line 50: What happens to the Node process (and/or Android app) if we throw an error & it doesn't get caught?

cc @gmaclennan

src/backend/lib/upgrade-storage.js
```diff
@@ -42,13 +43,44 @@ class Storage {
...
+      );
+      const toKeep = options.filter(option =>
+        this.isNewerThanCurrentApk(option)
+      );
+      foreachAsync(
+        toDelete,
+        (option, cb) => {
+          rimraf(option.filename, cb);
+        },
+        _ => {
```
hackergrrl, line 77: @gmaclennan Thoughts on whether we should error if some of the deletions fail?

I'm thinking it might be safest to add the files that we failed to delete to the `toKeep` list, so at least they are tracked. Otherwise we could have failed-to-delete entries clogging up disk space that the upgrade system doesn't even have a reference to. So, my suggestion is: don't have `foreachAsync` fail on an error, but rather accumulate failures + report them back, and then add those failed deletions back to `toKeep` so they stay tracked.

src/backend/lib/upgrade-storage.js
```diff
@@ -65,36 +97,50 @@ class Storage {
...
+        done(null, results);
       });
-      log("getAvailableUpgrades results:", results);
-      cb(null, results);
     });
   }
 
   // String, String, Callback<Void> -> WritableStream
   createApkWriteStream(filename, version, hash, cb) {
+    // Why doesn't accept locks?
```
hackergrrl, line 141: What do you mean by this?

src/backend/test/upgrade-storage.js
```diff
@@ -220,6 +220,53 @@ test("leftover files in the upgrade temp dir get wiped on init", t => {
...
+    expected.hash,
+    err => {
+      t.error(err);
+      storage.getAvailableUpgrades((err, options) => {
+        t.error(err);
+        scrub(options);
+        t.equals(options.length, 1);
+        t.deepEquals(options[0], expected);
+        const newStorage = new Storage(dir, storageOptions);
+        fs.stat(path.join(dir, "foo.apk"), (err, stat) => {
```
hackergrrl, line 256: It's risky doing this check synchronously: the async `init()` code may not have had a chance to run yet. I've seen different behaviour around this kind of code even just between version of Node.

An idea: expose a `ready(cb)` function on `UpgradeStorage` that just does e.g.
```js
  ready (cb) {  
    this.readLock(release => { release(); cb() })
  }
```
Then you can know when it's safe to check that the file was deleted.

src/backend/lib/upgrade-storage.js
luandro: When I tried using locks with this function the program came to a hault with tests. Only investigated enough to find out the locks there were causing problems, will dig further for more detail.

gmaclennan:
Incorporated into #595

Issue #589, opened 2021-05-10
Write an APK parser that extracts the APK's bundleId and supported architectures


hackergrrl:
Split off from #585, which depends on this.

> There is an Android Java API to read this data, and that API can be called from React Native e.g. see https://github.com/staltz/dat-installer/blob/master/android/app/src/main/java/com/staltz/datinstaller/PackageInfoModule.java
>
> However, this information is needed by the backend nodejs process, so accessing it via the Android Java API would require a complex back-and-forth via the React Native process.
>
> It would be better to read this information directly from the APK in the nodejs-mobile process. An APK is a Zip archive, and the version and bundleId are stored in the file AndroidManifest.xml that is always present in the root of the Zip archive (see https://developer.android.com/guide/topics/manifest/manifest-intro). AndroidManifest.xml is encoded as binary XML.
>
> Fortunately there is a tool to read this information: app-info-parser. This tool also reads the resources.arsc file, which includes the app icons, which we don't need to it adds a small memory overhead, but if this is a performance issue in the future we could fork app-info-parser and make that parsing separate.
>
> Note that the package field of the APK almost always, but not actually always, the same as the applicationId see https://developer.android.com/guide/topics/manifest/manifest-element#package I think in our case we can trust that it is the same, but I don't quite understand this difference.
>
> Unfortunately there is no Java API for reading the supported architectures within an APK (at least I can't find one). I think the best way to read the supported architectures of an APK would be to read the contents of the lib folder in the APK. It will contain a folder for each supported architecture. I have started work to add a readdir() method to random-access-zip (also, the name is bad, it should be zip-fs since it replicates the fs methods for a zipfile).

Issue #585, opened 2021-05-07
feat(p2p-upgrade): Read APK version, bundleId and supported arch before installing


gmaclennan:
p2p upgrade currently trusts metadata from a peer to know the version of an APK.

We should check the APK to ensure that it really is the version the metadata says it is, and also check:

- `bundleId` (also called `applicationId`) e.g. `com.mapeo`
- Supported architectures e.g. `arm64-v8a`, `armabi-v7a`

There is an Android Java API to read this data, and that API can be called from React Native e.g. see https://github.com/staltz/dat-installer/blob/master/android/app/src/main/java/com/staltz/datinstaller/PackageInfoModule.java

However, this information is needed by the backend nodejs process, so accessing it via the Android Java API would require a complex back-and-forth via the React Native process.

It would be better to read this information directly from the APK in the nodejs-mobile process. An APK is a Zip archive, and the version and bundleId are stored in the file `AndroidManifest.xml` that is always present in the root of the Zip archive (see https://developer.android.com/guide/topics/manifest/manifest-intro). `AndroidManifest.xml` is encoded as binary XML.

Fortunately there is a tool to read this information: [`app-info-parser`](https://github.com/chenquincy/app-info-parser). This tool also reads the `resources.arsc` file, which includes the app icons, which we don't need to it adds a small memory overhead, but if this is a performance issue in the future we could fork `app-info-parser` and make that parsing separate.

Note that the `package` field of the APK almost always, but not actually always, the same as the `applicationId` see https://developer.android.com/guide/topics/manifest/manifest-element#package I think in our case we can trust that it is the same, but I don't quite understand this difference.

Unfortunately there is no Java API for reading the supported architectures within an APK (at least I can't find one). I think the best way to read the supported architectures of an APK would be to read the contents of the `lib` folder in the APK. It will contain a folder for each supported architecture. I have started work to add a [`readdir()` method to `random-access-zip`](https://github.com/gmaclennan/random-access-zip/pull/9) (also, the name is bad, it should be `zip-fs` since it replicates the `fs` methods for a zipfile).

Ping @noffle 

Depends on #589. Part of #559.

gmaclennan:
[...]

Issue #559, opened 2021-04-21
(back) Filter updates by bundleId


gmaclennan:
We release different "variants" of Mapeo, that are distinguished by different "bundle id"s or "application id"s. Each variant with a different bundle id installs as a separate app on the phone, so you can have two variants of Mapeo running alongside each other with their own data, e.g. the QA variant and the main variant.

If we do not check that the bundle id of the update matches the running application, then we could end up upgrading / installing a different app to the one that is currently running. E.g. if someone is using the QA variant, and they get an update from someone running the main variant, then the QA variant would install the update over the main variant on their phone, which is not what the user would be expecting (the QA variant that they are running would not update).

To avoid this, we need to only accept updates that share the same bundle id. I think that it makes sense to avoid discovery of different variants altogether, e.g. QA variants should only discover other QA variants for finding updates. Could we incorporate the bundleId into the discovery key?

Current variants we build:

- `com.mapeo`: the "main" variant of the app - blue icon
- `com.mapeo.qa`: used for QA testing without overwriting the main variant - red icon
- `com.mapeo.debug`: used for development when running locally - yellow icon
- `com.mapeo.icca`: The "Mapeo for ICCAs" variant created for WCMC - green icon

There are also `.debug` variants of the QA and ICCA builds. These are only built as internal builds used for debugging errors with release candidates.

---

Kira's work checklist:
- [ ] pass bundleId down into UpgradeStorage
- [ ] perform discovery by key `${bundleId}-upgrade`
- [ ] write APK parser to determine bundleId (#589)
- [ ] alongside hash checks, also verify the bundleId of the APK matches our bundleID (#585)

hackergrrl:
This makes sense -- I think it's a good solution. The alternative would be to store the `bundleId` in the upgrade metadata, but I don't see a benefit to doing it that way vs the discovery key. Right now we use the key `mapeo-upgrade`. We could instead use `mapeo-upgrade-${bundleId}`.

The main work is to pass the `bundleId` into the UpgradeManager, where it'd be passed down into the UpgradeDownloader and UpgradeServer so they can incorporate it into the discovery key they listen/search on.

gmaclennan:
> This makes sense -- I think it's a good solution. The alternative would be to store the `bundleId` in the upgrade metadata, but I don't see a benefit to doing it that way vs the discovery key. Right now we use the key `mapeo-upgrade`. We could instead use `mapeo-upgrade-${bundleId}`.

[...]

Issue #537, opened 2021-04-09
(Front) Recovering from transient errors


ximenabb:
**Description :**
Issue : When a download is interrupted (kill upgrade simulator) app shows an error that can't be dismissed

**Proposed solution**
Errors that do not represent a permanent state (e.g. something happened, but upgrade search/download can continue/restart) can be shown as a standard Android dialog, that is displayed until the user dismisses it.

**Acceptance Criteria**
TO ADD

**Technical Notes:**
TO ADD

**Points:**
3pts

**Dependencies**
@noffle I am not clear on the different types of error that can be part of the upgrader state. Could you document the type/shape of the upgrade manager state somewhere? I'm also wondering that if the error is transient (e.g. it is something that happens like "upgrade fail" rather than a permanent error state like "upgrader broken"), then maybe it does not belong in state and should be modelled as an event?

gmaclennan:
@noffle I think to move forwards on this I need to understand the difference between:

1. Error "events", e.g. a failed download, that results in the upgrader being in a new, non-error state, e.g. Searching for downloads -> downloading --(failed download)--> Searching for downloads
2.  vs. Error "state" - e.g. something that results in the the upgrader being in an error state until another event happens, or is unrecoverable.

Could you help me understand the different errors and states in the current implementation?

hackergrrl:
@gmaclennan See my notes in https://github.com/digidem/mapeo-mobile/pull/573/files -- let's discuss the error types there?

PR #595, opened 2021-05-17
feat: Add experimental support for peer-to-peer updates

gmaclennan:
This PR started as an attempt to add strict TypeScript type-checking to this code, but that work resulted in identifying several bugs and some issues with the existing architecture that created many opportunities for race conditions.

This is a really complicated piece of code, with some complicated logic for managing peers appearing and disappearing from a local network. It's important that we can know it works reliably, but also the code needs to be well documented and easy to maintain by the entire team. This PR attempts to address that.

In general I've found that writing this with Typescript and async / await has resulted in much faster development - the code here took about 25 hours to write (excluding thinking time away from the computer) and it should be easier and quicker to add features and fix bugs.

### Typescript

Everything is strictly typed now.

### State Management

We were using the module `rwlock` to manage async state management. The main async state that we need to maintain is that the service can start and stop many times (because the user can navigate to and from the sync screen, or close and re-open the app) and the start and stop functions can be async. The use of `rwlock` meant that the `starting` and `closing` states were hidden.

I created an abstraction, `AsyncService`, which manages start and stop of an async service according to tested rules. This means that we can test this complex code and simplify the code where the upgrade logic lies.

### async / await

Using callbacks with code that has so many async actions makes it hard to follow what is happening and catch race conditions. I have changed all the code to use async / await patterns, which I think makes the code easier to follow and maintain by avoiding nested callbacks and ensuring that the code flow is the same as the logic flow (top-to-bottom).

### State complexity

The shape of the `state` emitted by the UpgradeManager was complex, which resulted in a lot of code to decide what state to show to the user in the UX. The main things that the UX needs to know are:

1. Is the p2p upgrade service running (without error)?
2. Progress of any downloads
3. Progress of any uploads
4. Is there an update downloaded and available to install?

This PR simplifies the state to just capture that information, it is documented in `lib/types.d.js` as `UpgradeState`

### Code documentation

It's really important that this code is easy for others on the team to understand and maintain. It took me quite a while to parse the original code, so I have written many comments to explain why the code is written the way it is. I have also tried to be really clear about what are public methods and what methods are private (e.g. internal to the implementation and not something to be tested or used outside each class), with explanations of what each class and method does. I have used JSDoc format for comments, extended with Typescript annotations.

### Server

- No need for drain logic, since server.close() will stop accepting new connections, then wait for existing connections to close before closing.
- Include url property in installer schema returned by server (url is for downloading the APK)
- Fix start/stop logic in server to avoid race conditions
- Validate list installers response body matches schema
- Use fastify for better error handling and simpler API
- Simplified emitted state (only need a list of active uploads from server)

### Storage

- Simplify state logic and async to remove race conditions
- Read info about APKs from files themselves
- Cache installer info rather than read it each time
- Don't read files into memory for generating hash
- Use async methods for disk operations
- Turn into EventEmitter so can listen to available installers
- Async initialization to allow async cleanup of files (and reading info about current installers)
- createWriteStream() will only finish (e.g. emit "finish" event) when the installer is completely written to storage. This was tricky code, so I abstracted it as a util `startFinishStream()` and thoroughly tested it.

### Discovery

- Ensure onPeer event is removed when discovery is stopped, to avoid the event being added multiple times
- Add TTL & timeout to discovered installers from peers, so that installers from peers that go offline are removed from state
- Validate lists of installers from peers and protect against prototype pollution
- Simplified state that emits a list of available installers
- No logic about evaluating installers - will return all discovered installers no matter platform, version etc.
- Check port is available on each start (since it could become unavailable between stop() and start())
- createReadStream() for downloading (but abstracts how this is done) and ensures that an installer is removed from "available" if download fails (to avoid repeat attempts at download - will become available again if re-discovered by discovery.lookup())

### Evaluating potential upgrades

We had several places for comparing and evaluating upgrades. I tried to move all this logic into testable utility functions, with two separate tests:

1. Checking if a given installer is compatible with the current device
2. Getting the most recent of compatible installers

This should make it easier to add additional logic for selecting a suitable upgrade candidate.

### Next Steps

I really hope that these changes make this code more reliable and easier to maintain in the future.

There are a few tasks remaining (with estimated time):

- [x] Tests for upgrade-discovery (3h)
- [x] Tests for upgrade-server (2h)
- [x] Tests for upgrade-manager (4h)
- [x] Integration of new API with `server.js` (1h)
- [x] Update frontend code for new state shape (2h)
- [ ] Cleanup AsyncService code https://github.com/digidem/mapeo-mobile/pull/595#discussion_r635513490 (1h) - non-essential chore
- [x] Change all `_name` to `#name` https://github.com/digidem/mapeo-mobile/pull/595#discussion_r636315925 (1h) - non-essential chore
- [x] Evaluate how long to wait for initial download (1h)
- [x] Accidentally fixed beforeAfterStream tests for node@16, breaking node@12. Revert (1h)

When complete this code should close these issues: #589, #585, #576, #562, #559, #537


hackergrrl:
I have yet to read it all through, but this looks like it's going to be a really valuable set of changes re: the reliability & maintainability of the backend codebase.

Honestly, it kinda makes me wanna just flip the switch and go full typescript on the MM backend >__>;

hackergrrl:
@gmaclennan Can you rebase around this branch so that the diff only includes your changes & not also a bunch of merged changes from develop? This will make the diff a lot easier to read imo. Feel free to merge develop back into share-apk-p2p to make this happen!

gmaclennan:
@hackergrrl sure thing, now rebased.

src/backend/lib/async-service.js
```diff
@@ -0,0 +1,199 @@
...
+   * @returns {Promise<void>} Resolves when service is started
+   */
+  async start(...args) {
+    switch (this.getState().value) {
+      case "starting":
+      case "started":
+        // Will still resolve once has started
+        return this.#started.promise;
+      case "error":
+        this.#started.reject(this.getState().error);
```
hackergrrl: It seems like this code path could be hit multiple times once the service enters the "error" state: what happens if `reject()` gets called multiple times on the `#started` promise?

src/backend/lib/async-service.js
```diff
@@ -0,0 +1,199 @@
...
+        break;
+      default:
+        return this.#started.reject(new Error("Should not happen"));
+    }
+    // We've begun starting, so stopping promise needs to be unresolved until
+    // we stop again
+    this.#stopped = pDefer();
+    try {
+      this.setState({ value: "starting" });
+      await this._start.apply(this, args);
```
hackergrrl, line 211: Shouldn't `#started` go back into a pending state before calling `_start()`? (Same Q for `stop` below)

src/backend/lib/upgrade-discovery.js
```diff
@@ -0,0 +1,211 @@
...
+   * Create a readstream for an available installer from another peer
+   *
+   * @param {string} hash
+   * @returns {import('stream').Readable}
+   */
+  createReadStream(hash) {
+    const installer = this.#availableInstallers.get(hash);
+    if (installer) {
+      const stream = got.stream(installer.installer.url);
+      stream.on("error", e => {
```
hackergrrl:
```suggestion
      stream.once("error", e => {
```

src/backend/test/upgrade-discovery.test.js
```diff
@@ -0,0 +1,268 @@
...
+) {
+  const filepaths = apkFilenames.map(filename =>
+    path.join(validApksFolder, filename)
+  );
+  const { storage, expected, cleanup } = await setupStorage(
+    filepaths,
+    currentApkInfo
+  );
+  const discovery = new Discovery({
+    discoveryKey: "com.mapeo",
```
hackergrrl: TODO: use bundleId to determine key

.nvmrc
```diff
@@ -1 +1 @@
-10.16
+12.16.3
```
hackergrrl, line 1: You mentioned needing node > 12. Is this the new minimum?

hackergrrl (commented):
Initial comments. 13/46 files reviewed 💦

src/backend/test/upgrade-discovery.test.js
gmaclennan: This is just in the test (hard-coded so that test instances discovery each other). In the code it used bundleId: https://github.com/digidem/mapeo-mobile/blob/p2p-update/typescript/src/backend/lib/upgrade-manager.js#L103
However, we should also test to make sure changing this means discovery does not happen.

.nvmrc
gmaclennan: The code uses a few JavaScript features that are only available in Node v12. I'm not sure how many are in Node >v12.0.0, but since the production code is running on Node v12.6.3 (due to the version of nodejs-mobile we use), I thought it best to target this specific version for development and testing.

src/backend/lib/async-service.js
gmaclennan: Good question! My understanding of promises is that once they are "fullfilled" (either rejected or resolved) then any further `resolve()` or `reject()` is a no-op. However the ECMA spec does not explicitly say anything like this. TL;DR `reject()` should be a no-op when called more than once. I can write a test to confirm this, or we could make our own deferred promise code that keeps state, e.g.

```js
function pDefer() {
  let fulfilled = false
  deferred.promise = new Promise((resolve, reject) {
    deferred.resolve = arg => {
      if (fulfilled) return
      fulfilled = true
      resolve(arg);
    }
    deferred.reject = arg => {
      if (fulfilled) return
      fulfilled = true
      reject(arg)
    }
  });
  return deferred;
```

However I don't think this is necessary.

src/backend/lib/async-service.js
gmaclennan: We can only get here if `state.value === "stopped"` (because of switch above).
If `state.value === "stopped"` we guarantee that `#started.promise` is unfulfilled (e.g. pending) because the action `stopping` --> `stopped` (the only way to get to the `stopped` state) resets that promise.
The reset of `#started` needs to happen there, rather than here, because otherwise multiple calls to `start()` in the same tick would return different promises, and we would need to track and resolve all of them.
At least that all makes sense in my head! I tried to make sure the tests check all of this.

src/backend/lib/async-service.js
gmaclennan: Thanks for this comment. I added some tests for the error state behaviour and discovered a bug in the tests and a bug in the code. I think it is fixed now. This start() stop() seems simple but there are a lot of edge cases, so I'm happy we have this as an abstraction so we can test it in isolation.

src/backend/test/upgrade-discovery.test.js
gmaclennan: Added a test for changing discovery keys (to ensure they don't discovery each other)

src/backend/lib/async-service.js
hackergrrl: Yes, I really liked the idea to separate out the async service bits so they can be reasoned about on their own.

src/backend/lib/async-service.js
hackergrrl: I see. Yeah, it's a bit difficult to track that mentally, but it makes sense!

I don't really like the separate management of the started/stopped promises; it seems fragile. I wonder if there's a way to encapsulate them so that the logic is more coherent. This might be a v2 thing though.

hackergrrl: I haven't fully thought this through, but: what if `setState` was the only part that touched the started/stopped promises? That would centralize the logic & maybe make it easier to reason about.

src/backend/lib/upgrade-discovery.js
```diff
@@ -0,0 +1,211 @@
...
+const createDiscovery = require("dns-discovery");
+const secureJson = require("secure-json-parse");
+const { isInstallerList } = require("./schema");
+const throttle = require("lodash/throttle");
+const stream = require("stream");
+const log = require("debug")("p2p-upgrades:discovery");
+
+// TTL for discovered peers, if they are not seen for more than this period then
+// they are considered offline and forgotten
+const TTL = 4000;
```
hackergrrl: Nitpick:
```suggestion
const PEER_TTL_MS = 4000;
```

src/backend/lib/async-service.js
gmaclennan: Yes I agree, I was just thinking about this today - your comment highlighted a fragility here. As long as only this parent class manages state, everything is encapsulated and easier to reason about, but if a subclass were to change the state to "error" for example, that would break things. I will put on the todo to encapsulate everything in the parent class - I made a plan while driving to my parents'. I also agree that using promises for encapsulating this logic is harder to reason about, I think using an internal event emitter would be simpler (that is the approach I've used in the past, but this time I thought I'd try the deferred promise approach, which makes the code more concise, but at the expense of hiding state logic within promise states, so I don't like it on reflection).

src/backend/lib/upgrade-manager.js
```diff
@@ -1,192 +1,275 @@
...
+ */
 
-    this.downloader.on("state", state => {
-      if (state.search.state === "SEARCHING") {
-        this.upgradeOptions = state.search.context.upgrades;
-      } else if (state.search.state === "Idle") {
-        this.upgradeOptions = [];
-      }
+// Throttle state events (in ms)
+const EMIT_THROTTLE = 400; // milliseconds
```
hackergrrl:
```suggestion
const EMIT_THROTTLE_MS = 400; // milliseconds
```

src/backend/lib/upgrade-manager.js
```diff
@@ -1,192 +1,275 @@
...
+ *   },
+ *   // If an installer is downloaded and is more recent than the currently
+ *   // running APK, then this value will be set to the info about that installer
+ *   availableUpgrade?: {
+ *     // See schema.js
+ *   }
+ * }
+ * ```
+ *
+ * Will emit an "error" event if an upload or download fails. This is most
```
hackergrrl, line 80: Maybe a note here about how transient vs fatal errors are handled? (It's not yet clear to me!)

src/backend/lib/upgrade-manager.js
```diff
@@ -1,192 +1,275 @@
...
+class UpgradeManager extends AsyncService {
+  /** @type {UpgradeState} */
+  #state = {
+    value: "stopped",
+    downloads: [],
+    uploads: [],
+    availableUpgrade: undefined,
+  };
+  /** @type {Map<string, { progress: TransferProgress, stream: import('stream').Readable, installerInfo: InstallerExt }>} */
+  #downloads = new Map();
```
hackergrrl, line 107: What is the difference between `#name` and `_name` use for member names?

src/backend/lib/upgrade-manager.js
```diff
@@ -1,192 +1,275 @@
...
+  /**
+   * @param {object} options
+   * @param {string} options.storageDir Path to folder to store installers
+   * @param {InstallerInt} options.currentApkInfo info about the currently installed APK
+   * @param {import('./types').DeviceInfo} options.deviceInfo Arch and SDK version of current device
+   */
+  constructor({ storageDir, currentApkInfo, deviceInfo }) {
```
hackergrrl, line 114: Yay, glad `setApkInfo` was factored out.

src/backend/lib/utils.js
```diff
@@ -0,0 +1,268 @@
...
+  return {
+    versionName: result.versionName,
+    versionCode: result.versionCode,
+    applicationId: result.package,
+    arch,
+    minSdkVersion: (result.usesSdk || {}).minSdkVersion || 1,
+  };
+}
+
+// If an APK does not have native code (e.g. no `lib` folder) then we assume it runs on these ABIs
```
hackergrrl, line 287: Is this the default android bundling..?

src/backend/lib/upgrade-manager.js
```diff
@@ -1,192 +1,275 @@
...
+  /**
+   * Called whenever the list of installers in local storage changes. If there
+   * is an upgrade candidate in storage, then this will emit state with that
+   * upgrade candidate
+   *
+   * @private
+   * @param {InstallerInt[]} storedInstallers
+   */
+  _onStoredInstallers(storedInstallers) {
+    const upgradeCandidate = getUpgradeCandidate({
```
hackergrrl: I think that `getBestUpgradeCandidate` could be a more instructive name for this helper function?

src/backend/lib/upgrade-manager.js
```diff
@@ -1,192 +1,275 @@
...
-        message: this.state.downloader.check.context.message,
-      };
-    }
+  /**
+   * Called whenever the list of installers in local storage changes. If there
+   * is an upgrade candidate in storage, then this will emit state with that
+   * upgrade candidate
+   *
+   * @private
+   * @param {InstallerInt[]} storedInstallers
```
hackergrrl, line 176: I like this concept of internal and external installer types -- good idea!

src/backend/lib/upgrade-manager.js
```diff
@@ -1,192 +1,275 @@
...
+      // same hash, ignore this one
+      if (this.#downloads.has(candidateHash)) return;
+      // If we are already downloading an installer with a higher version number
+      // than this one, ignore this one
+      for (const { installerInfo } of this.#downloads.values()) {
+        if (installerCompare(upgradeCandidate, installerInfo) < 1) return;
       }
-    });
-  }
+      // If there are any in progress downoads, cancel them
```
hackergrrl:
```suggestion
      // If there are any in progress downloads, cancel them
```

src/backend/lib/upgrade-manager.js
```diff
@@ -1,192 +1,275 @@
...
+   * 2. Starts the download discovery service
+   * 3. Starts announcing on port
+   *
+   * @returns {Promise<void>} Resolves when started
+   */
+  async _start() {
+    log("got request to start services");
+    const port = await getPort();
+    await Promise.all([this._discovery.start(port), this._server.start(port)]);
+    log("Started services");
```
hackergrrl: Why does `_stop` need to call `setState` but `_start` doesn't?

src/backend/types/write-only-stream.d.ts
```diff
@@ -0,0 +1,11 @@
+// Type definitions for dns-discovery 6.2.3
+// Project: https://github.com/mafintosh/dns-discovery
+// TypeScript Version: 4.2
+
+/// <reference types="node" />
+
+declare module "write-only-stream" {
```
hackergrrl: What is this used for?

src/backend/types/gmaclennan_zip-fs.d.ts
```diff
@@ -0,0 +1,15 @@
+// Type definitions for @gmaclennan/zip-fs 1.2.0
+// Project: https://github.com/gmaclennan/zip-fs
+// TypeScript Version: 4.2
+
+/// <reference types="node" />
```
hackergrrl, line 5: What does this line do?

hackergrrl (commented):
Another batch of code reviewed! 19/46 files so far :D

src/backend/lib/utils.js
gmaclennan: Yes in my testing, if I create an empty Android app with no C++ code (eg only Java or Kotlin) then there is no "lib" folder. I assume that in that case the apk would run on any platform that Android runs on, but the main 4 (arm and x86, 32 and 64 bit) I think cover all likely platforms. In addition I don't see a future where we would ever release mapeo without native compiled code (it will either be node or rust I think)

src/backend/lib/upgrade-discovery.js
gmaclennan: Good :)

src/backend/lib/upgrade-manager.js
gmaclennan: Confession: I didn't know how to initialize [private instance fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#private_instance_fields) from the constructor, so when I needed to initialize them there I used `_name`, which is just a visual indication (as opposed to `#name` which is enforced by the language). I need to switch these all over to:

```js
class ClassWithPrivateField {
  #privateField

  constructor() {
    this.#privateField = 42
  }
}
```

src/backend/lib/upgrade-manager.js
gmaclennan: I went backwards and forwards on this, and was initially passing in just the filepath to the current Apk, and reading the InstallerInfo within `UpgradeManager`, but that made the creation of the storage, server and discovery instances to be async, which made the code much more complicated, so now I'll change `server.js` to read that currentApkInfo from the file before the `UpgradeManager` instance is created.

src/backend/lib/upgrade-manager.js
gmaclennan: Good catch! That is a bug, it should not be there in `_stop`, not sure how it sneaked in!

src/backend/types/gmaclennan_zip-fs.d.ts
gmaclennan: I don't fully understand these ["triple-slash directives"](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-path-) in Typescript, but I am copying this from other modules: it tells Typescript that the types reference types available in node.

src/backend/types/write-only-stream.d.ts
gmaclennan: That is now unused. I used it in my first attempt at `beforeAfterStream.js`. Now removed.

src/backend/lib/upgrade-manager.js
gmaclennan: Agreed! Changed.

src/backend/lib/upgrade-manager.js
gmaclennan: Added to this PR's task list

src/backend/lib/async-service.js
gmaclennan: Added to this PR's task list

src/backend/lib/upgrade-manager.js
```diff
@@ -149,34 +168,77 @@ class UpgradeManager extends AsyncService {
...
       // If we are already in the process of downloading an installer with this
       // same hash, ignore this one
-      if (this.#downloads.has(candidateHash)) return;
+      if (this.#downloads.has(candidateHash)) {
+        log(
+          `${this.#port}: Upgrade candidate ${candidateHash.slice(
+            0,
+            7
+          )} is downloading`
+        );
```
hackergrrl, line 238: Such weird & imo unreadable auto-formatting! Is the line width configurable? Most of us probably use > 80 columns -- 120 might give improved readability.

src/backend/lib/upgrade-storage.js
```diff
@@ -1,271 +1,215 @@
...
-      this.currentApk = info;
-      cb();
-    });
+  /**
+   * Get a list of available installers
+   *
+   * @returns {Promise<InstallerInt[]>}
+   */
+  async list() {
+    await this.started();
```
hackergrrl, line 116: This is a nice pattern/trick to replace the locks. Very readable.

src/backend/lib/upgrade-storage.js
```diff
@@ -1,271 +1,215 @@
...
+      .map(file => path.join(this.#storageDir, file));
+    const storedInstallerInts = await Promise.all(
+      // Catch errors trying to read the apk file and ignore it
+      apkFiles.map(file => getInstallerInfo(file).catch(() => {}))
+    );
+    /** @type {Set<InstallerInt>} */
+    const forDeletion = new Set();
+
+    for (const installer of storedInstallerInts) {
+      if (!installer) continue;
```
hackergrrl, line 77: Is this idea here to skip entries that got mapped to `{}`? If so, this won't work: `!!{} === true`.

src/backend/lib/upgrade-storage.js
```diff
@@ -1,271 +1,215 @@
...
+      // Mark any upgrade < current installed APK for deletion
+      if (installerCompare(installer, this.#currentApkInfo) === -1) {
```
hackergrrl: Sanity check: this won't delete older versions of installers for different arches/platforms, right? e.g. We don't want to delete MapeoDesktop@3.0.0 just because we're running MapeoMobile@5.0.0.

src/backend/lib/upgrade-storage.js
```diff
@@ -1,271 +1,215 @@
...
+          // Everything after here happens after createWriteStream has finished, but
+          // that's ok because this is UpgradeStorage updating its state
+          const installer = await getInstallerInfo(tmpFilepath);
+          if (!installer) {
+            // We couldn't read the info about this installer
+            throw new Error("Could not parse APK");
+          }
+          if (installer.hash !== expectedHash) {
+            // Hashes did not match, download was broken or incorrectly advertised
+            throw new Error("Invalid hash");
```
hackergrrl:
```suggestion
            throw new Error("Actual hash does not match expected hash");
```
The hash is [likely] a valid sha256 hash, but it's just not the value we expected it to be.

src/backend/lib/utils.js
```diff
@@ -0,0 +1,299 @@
...
+function installerCompare(a, b) {
+  const aIsValidSemver = semver.valid(a.versionName);
+  const bIsValidSemver = semver.valid(b.versionName);
+  if (!aIsValidSemver && !bIsValidSemver) return 0;
+  if (!aIsValidSemver) return -1;
+  if (!bIsValidSemver) return 1;
+  // Important to validate semver before here, otherwise this will throw
+  const semverCompare = semver.compare(a.versionName, b.versionName);
+  if (semverCompare !== 0) return semverCompare;
+  if (a.versionCode < b.versionCode) return -1;
```
hackergrrl, line 53: What is `versionCode`? Consider adding some comments on the `Installer{Ext,Int}` in `schema.js`.

src/backend/lib/utils.js
```diff
@@ -0,0 +1,299 @@
...
+ * @param {import('./types').DeviceInfo} options.deviceInfo
+ * @param {InstallerInt | InstallerExt} options.installer
+ * @param {InstallerInt | InstallerExt} options.currentApkInfo
+ * @returns {boolean}
+ */
+function isUpgradeCandidate({ deviceInfo, installer, currentApkInfo }) {
+  // Filter out the current APK
+  if (installer.hash === currentApkInfo.hash) return false;
+  // TODO: 64 bit devices can run 32 bit APKs
+  const isSupportedAbi = !!installer.arch.find(arch =>
```
hackergrrl: Just a little simpler.
```suggestion
  const isSupportedAbi = installer.arch.some(arch =>
```

src/backend/lib/utils.js
```diff
@@ -0,0 +1,299 @@
...
+    deviceInfo.supportedAbis.includes(arch)
+  );
+  if (!isSupportedAbi) return false;
+  if (installer.minSdkVersion > deviceInfo.sdkVersion) return false;
+  if (installer.applicationId !== currentApkInfo.applicationId) return false;
+  if (installer.platform !== "android") return false;
+  if (installer.versionCode < currentApkInfo.versionCode) return false;
+  if (!semver.valid(installer.versionName)) return false;
+  if (installerCompare(installer, currentApkInfo) < 1) return false;
```
hackergrrl: I notice we use this function a lot in the new codebase. For readability, consider some helper functions like `semver` offers, like `installerCompare.{lt,gt,eq}`. For me, I always need to pause and think for a few moments to reason about cmp function outputs.

src/backend/lib/utils.js
```diff
@@ -0,0 +1,299 @@
...
+
+/**
+ * Check if an installer is compatible with the given device. Requirements:
+ *
+ * 1. Must support arch of device
+ * 2. Must support SDK of device
+ * 3. Version code must be greater or equal to currently installed app
+ * 4. Installer must be for platform 'android'
+ * 5. applicationId must match applicationId of currently installed app
+ * 6. versionName must be valid semver
```
hackergrrl, line 127: What about `versionName` being newer than the current version?

src/backend/lib/utils.js
```diff
@@ -0,0 +1,299 @@
...
+  if (a.versionCode > b.versionCode) return 1;
+  return 0;
+}
+
+/**
+ * Check if an installer is compatible with the given device. Requirements:
+ *
+ * 1. Must support arch of device
+ * 2. Must support SDK of device
+ * 3. Version code must be greater or equal to currently installed app
```
hackergrrl, line 124: Why equal? (I'm still not clear on what versionCode is, so that may also explain my uncertainty.)

src/backend/lib/utils.js
```diff
@@ -0,0 +1,299 @@
...
+ * Useful if you want to synchronously return a stream, then asynchronously
+ * create a write stream, then do some work after the internal stream has
+ * finished writing (e.g. move a file) before the returned stream finishes.
+ *
+ * @param {object} options
+ * @param {() => Promise<stream.Writable>} options.createStream
+ * @param {(wrappedStream: stream.Writable) => Promise<void>} options.finalize
+ * @returns {stream.Writable}
+ */
+function beforeAfterStream({ createStream, finalize }) {
```
hackergrrl, line 326: This is some pretty tricky code to get 100% correct! I don't even feel totally qualified to evaluate its correctness. As far as I know, the write stream handling code I had in UpgradeStorage worked correctly. Why was this new [complex!] piece of code needed?

src/backend/types/dns-discovery.d.ts
```diff
@@ -0,0 +1,79 @@
+// Type definitions for dns-discovery 6.2.3
```
hackergrrl, line 1: Did you generate these, or write them by hand? Curious what tool, if any, you used.

hackergrrl (commented):
Implementation and types are reviewed. Just tests left. (38/54 files done.)

src/backend/test/upgrade-manager.test.js
```diff
@@ -0,0 +1,387 @@
...
+/** @typedef {import('../lib/types').InstallerInt} InstallerInt */
+/** @typedef {import("../lib/types").DevicePlan} DevicePlan */
+
+/** @type {import('../lib/types').DeviceInfo} */
+const defaultDeviceInfo = {
+  supportedAbis: ["armeabi-v7a", "arm64-v8a"],
+  sdkVersion: 21,
+};
+
+test.only("One device updates another", async t => {
```
hackergrrl: ^^^^

hackergrrl:
I notice that the sample APKs add another ~15mb to the git repo. Is there anything we can remove from these to make them smaller? I ask because any binary data we add to the repo (even if we modify or remove it in the future) is kept around forever.

hackergrrl:
I haven't had a chance to run this yet, but: how are fatal vs non-fatal errors handled on the frontend? I wasn't totally clear on the flow.

src/backend/types/dns-discovery.d.ts
gmaclennan: I write them by hand based on the docs, but often only typing the pieces that I use. Most modules already have open-source types in [Definitely Typed](http://definitelytyped.org) and can be installed by `npm i -D @types/dns-discovery`, but some less popular modules do not have community maintained types. It's confusing learning how to write these. You can just create a placeholder file with just the line: `declare module 'dns-discovery' {}` and Typescript will basically just ignore the module and not type-check your usage of it. I find writing these types helpful for sanity-checking how I am using the API of a module.

src/backend/lib/utils.js
gmaclennan: You are correct, it should not be equal. Android will allow you to install an app with an equal version code, but if we allow that it will allow an inifinite loop of updates (imagine an APK with the same version name, e.g. `1.0.0` as the current app, and the same version code, but a different hash. It would be considered an update, but then in the other direction is also considered an update). We use build number for version code, so for Mapeo `versionCode` = build number. Build number is guaranteed to increment for every build of an APK we create on our CI.

gmaclennan: Luandro is fixing this in https://github.com/digidem/mapeo-mobile/pull/596

src/backend/lib/utils.js
gmaclennan: Good point! @luandro can you add this to the comment in https://github.com/digidem/mapeo-mobile/pull/596

src/backend/lib/utils.js
gmaclennan: That's a really good idea - I have had the same issue having to pause and think what `-1` vs `1` means.

src/backend/lib/upgrade-storage.js
gmaclennan: True! Thanks.

src/backend/lib/upgrade-manager.js
gmaclennan: Agreed! I like a lot of things about Prettier (it saves so much time for me when writing code), but I don't like how it formats long template literals. There is a long discussion about it: https://github.com/prettier/prettier/issues/3368
Let's tag the discussion about format line width for a discussion outside this PR, this is not something new added here.

src/backend/lib/upgrade-storage.js
gmaclennan: Glad it makes it readable.

src/backend/lib/upgrade-storage.js
gmaclennan: `getInstallerInfo` either returns the `InstallerInt` type or throws. Line 71 catches this throw and returns undefined, so the entries of `storedInstallerInts` are either `void` or `InstallerInt` types. This skips `undefined` (`void`) entries, there are no entries mapped to `{}`.

src/backend/lib/upgrade-storage.js
gmaclennan: Good catch. It raises an important question: what should the behaviour be for other arches/platforms? Keep the most recent of each platform?

src/backend/test/upgrade-manager.test.js
gmaclennan: oops!

src/backend/lib/utils.js
gmaclennan: Yes this turned out to be more complex than I expected. I embarked upon it from a desire to create a simpler interface for UpgradeStorage, and didn't like the confusion of having to handle two callbacks - once when the write stream to tmp file finished, and once for when the file finally moves. In the end I think an interface like: `upgradeStorage.writeStream(readableStream, callback)` would have been simpler, even though it also creates a lack of clarity about where errors on `readableStream` should be handled. I lean heavily on tests here, and I copied the tests from `duplexify`. I feel confident that how we intend to use this helper, and the errors that we might get, are covered in the tests.

gmaclennan:
> I notice that the sample APKs add another ~15mb to the git repo. Is there anything we can remove from these to make them smaller? I ask because any binary data we add to the repo (even if we modify or remove it in the future) is kept around forever.

I used the default empty project template in Android Studio. Just did a check and seems like turning on `minifyEnabled true` in `build.gradle` will reduce the size of each APK by 50%, so we could reduce this to ~8mb of APKs. Overall the git repo (including these APKs) is 75Mb right now. We can also use https://git-lfs.github.com although that adds a step for contributors

I can add a todo for re-generating these sample APKs minified to save space, then we can use https://rtyley.github.io/bfg-repo-cleaner/ to cleanup the larger ones.

gmaclennan:
> I haven't had a chance to run this yet, but: how are fatal vs non-fatal errors handled on the frontend? I wasn't totally clear on the flow.

Good point, will track that here: https://github.com/digidem/mapeo-mobile/issues/604

hackergrrl:
@gmaclennan Somehow (as an example), the [androidify base APK](https://github.com/node-on-mobile/androidify/blob/master/resources/base.apk) is only 230 kilobytes. Could be worth looking at for a few mins to see if there's something obvious we can do to reduce size.

src/backend/lib/upgrade-storage.js
hackergrrl: Great question. I think this is worth filing an issue for to be thought about, but doesn't need to be answered yet / in this PR.

src/backend/lib/upgrade-storage.js
hackergrrl: I don't really understand this. I wrote this code to play with it:

```js
async function get (x) {
  if (x === -1) throw new Error('bad')
  else return true
}

const ps = [1,2,-1,3,-1].map(x => get(x).catch(() => {}))

Promise.all(ps)
  .then(res => {
    console.log('res', res)
  })
  .catch(err => {
    console.log('err', err)
  })
```
As you say, the `-1`s get mapped to `undefined`. But, if I change `{}` to `"foo"`, I get `"foo"` instead of `undefined`. Why does `{}` map to `undefined`??

src/backend/test/async-service.test.js
```diff
@@ -0,0 +1,286 @@
...
+  let startCount = 0;
+  const service = new TestService({
+    async _start() {
+      await new Promise(res => setTimeout(res, 100));
+      startCount++;
+    },
+  });
+  service.start();
+  service.start();
+  await new Promise(res => setTimeout(res, 0));
```
hackergrrl, line 60: Why is this nextTick-like logic needed?

src/backend/test/async-service.test.js
```diff
@@ -0,0 +1,286 @@
...
+});
+
+test('If `_start()` throws, then the service is left in an unrecoverable "error" state.', async t => {
+  let startCount = 0;
+  const service = new TestService({
+    async _start() {
+      await new Promise(res => setTimeout(res, 100));
+      startCount++;
+      // Throw the first time, but not on subsequent calls, but this should
+      // never be called after it has thrown once
```
hackergrrl, line 108: If it shouldn't ever be called again after throwing, why have this guard around it? Wouldn't it be better to test it /wo any protections if the state should never happen?

src/backend/test/async-service.test.js
```diff
@@ -0,0 +1,286 @@
...
+      if (startCount === 1) throw new Error("MyError");
+    },
+  });
+  try {
+    await service.start();
+  } catch (e) {
+    t.ok(e instanceof Error, "throws on first call");
+  }
+  t.equal(service.getState().value, "error", "in error state");
```
hackergrrl: It would be nice to check that the state is "started" in the earlier tests also!

src/backend/test/async-service.test.js
```diff
@@ -0,0 +1,286 @@
...
+  let started = false;
+  const service = new TestService({
+    async _start() {
+      started = true;
+    },
+    async _stop() {
+      started = false;
+    },
+  });
+  await service.start();
```
hackergrrl, line 202: TODO: check the service's `state` after each operation?

src/backend/test/async-service.test.js
```diff
@@ -0,0 +1,286 @@
...
+    async _stop() {
+      await new Promise(res => setTimeout(res, 100));
+      stopCount++;
+    },
+  });
+  await service.start();
+  service.stop();
+  service.stop();
+  await new Promise(res => setTimeout(res, 0));
+  t.equal(stopCount, 0, "Service not yet started");
```
hackergrrl:
```suggestion
  t.equal(stopCount, 0, "Service not yet stopped");
```

src/backend/test/async-service.test.js
```diff
@@ -0,0 +1,286 @@
...
+});
+
+test('If `_stop()` throws, then the service is left in an unrecoverable "error" state.', async t => {
+  let stopCount = 0;
+  const service = new TestService({
+    async _stop() {
+      await new Promise(res => setTimeout(res, 20));
+      stopCount++;
+      // Throw the first time, but not on subsequent calls, but this should
+      // never be called after it has thrown once
```
hackergrrl, line 259: Same comment here as above re: not needing an `if` guard.

src/backend/test/async-service.test.js
```diff
@@ -0,0 +1,286 @@
...
+    },
+  });
+  service.start();
+  await nextTick();
+  service.start();
+  await nextTick();
+  service.stop();
+  await nextTick();
+  await service.start();
+  t.true(started, "service is started");
```
hackergrrl, line 301: Consider adding `getState` checks across these.

src/backend/types/dns-discovery.d.ts
hackergrrl: That seems like an interesting & useful exercise for building a stronger sense of understanding for how a module works.

src/backend/test/helpers.js
```diff
@@ -0,0 +1,245 @@
...
+ * @param {InstallerInt} currentApkInfo
+ */
+async function setupTmpStorageFolder(filepaths, currentApkInfo) {
+  const tmpDir = await tmp.dir({ unsafeCleanup: true });
+  const expected = [];
+  for (const filepath of filepaths) {
+    const dstFilepath = path.join(tmpDir.path, path.basename(filepath));
+    await fsPromises.copyFile(filepath, dstFilepath);
+    expected.push({
+      // Just an empty object if readJson throws - sorry for ugly syntax
```
hackergrrl, line 85: If `readJson` fails doesn't this mean something is VERY wrong with the requested test setup? Shouldn't this function throw in that case?

src/backend/test/helpers.js
```diff
@@ -0,0 +1,245 @@
...
+/** @typedef {import('../lib/types').DeviceInfo} DeviceInfo */
+/** @typedef {import('../lib/types').DevicePlan} DevicePlan */
+
+const validApksFolder = path.join(__dirname, "./fixtures/valid-apks");
+
+module.exports = {
+  readJson,
+  hashCmp,
+  setupStorage,
+  setupTmpStorageFolder,
```
hackergrrl, line 24: This isn't used outside of this file -- remove from exports?

src/backend/test/helpers.js
```diff
@@ -0,0 +1,245 @@
...
+ *
+ * @template {object} T
+ * @param {import('tiny-typed-emitter').TypedEmitter} emitter
+ * @param {string} eventName Name of event to await
+ * @param {T} expectedValue Promise will resolve once the emitted event partially matches this expected value
+ * @param {object} [options]
+ * @param {number} [options.timeout=10000] After timeout, if `expectedValue` has not been emitted, the promise will timeout
+ * @param {string} [options.name=""]
+ * @param {import('type-fest').PartialDeep<T>} [options.always] Every event emitted whilst waiting for `expectedValue` should match this value
+ * @param {import('type-fest').PartialDeep<T>} [options.never] If any event emitted whilst waiting for `expectedValue`, promise will throw
```
hackergrrl:
```suggestion
 * @param {import('type-fest').PartialDeep<T>} [options.never] If any event emitted whilst waiting for `expectedValue` matches this value, the promise will throw
```

src/backend/test/helpers.js
```diff
@@ -0,0 +1,245 @@
...
+ * using the lodaash `isMatch` function. This will ignore any undefined
+ * properties on `expectedValue` when comparing the value returned by the event
+ * emitter.
+ *
+ * @template {object} T
+ * @param {import('tiny-typed-emitter').TypedEmitter} emitter
+ * @param {string} eventName Name of event to await
+ * @param {T} expectedValue Promise will resolve once the emitted event partially matches this expected value
+ * @param {object} [options]
+ * @param {number} [options.timeout=10000] After timeout, if `expectedValue` has not been emitted, the promise will timeout
```
hackergrrl, line 132:
```suggestion
 * @param {number} [options.timeoutMs=10000] After timeout, if `expectedValue` has not been emitted, the promise will timeout
```

hackergrrl: Suggesting using the suffix `Ms` for clarity that the units are milliseconds.

src/backend/lib/types.d.ts
```diff
@@ -0,0 +1,111 @@
...
+  /** id (hash) of the file being transferrer */
+  id: string;
+  /** bytes transferred so far */
+  sofar: number;
+  /** total number of bytes to transfer */
+  total: number;
+}
+// Info about the current device
+export interface DeviceInfo {
+  supportedAbis: InstallerExt["arch"];
```
hackergrrl, line 43: What does this syntax do?

src/backend/test/helpers.js
```diff
@@ -0,0 +1,245 @@
...
+        }
+        if (value && typeof value.availableUpgrade === "string") {
+          // @ts-ignore
+          normalizedStep[
+            key
+          ].availableUpgrade = /** @type {InstallerInt} */ (await readJson(
+            path.join(validApksFolder, value.availableUpgrade)
+          ));
+        }
+      }
```
hackergrrl: Please consider moving this into its own (pure) helper `normalizeStep()` function! It's pretty hairy and it'd be a good candidate for self-contained logic that explains what it's going and why.

hackergrrl: Also, can you just use a module like `clone` here and append `availableUpgrade` to it afterwards? I cringe at the fragility of deep copy logic like this :)

src/backend/lib/types.d.ts
```diff
@@ -0,0 +1,111 @@
...
+  | UpgradeState
+  | (AsyncServiceStateDefault & {
+      uploads: TransferProgress[];
+      downloads: TransferProgress[];
+      availableUpgrade?: string;
+    });
+
+interface ScenarioStepEvent {
+  /** Message passed to t.pass() */
+  message?: string;
```
hackergrrl, line 92: Nitpick. You use VSCode right? Does it have refactor tools for these kinds of changes?
```suggestion
  passMessage?: string;
```

hackergrrl (commented):
More comments! 47/54 files reviewed.

src/backend/lib/upgrade-storage.js
gmaclennan: This is JavaScript syntax being confusing. `() => {}` is a no-op function, the same as `function noop () {}`. The `{}` here is the empty body of a function, not an empty object. My apologies, this is a habit I've always used for writing a noop that may not be clear.

src/backend/lib/utils.js
luandro: :+1:

src/backend/lib/utils.js
luandro: :+1:

gmaclennan:
> @gmaclennan Somehow (as an example), the [androidify base APK](https://github.com/node-on-mobile/androidify/blob/master/resources/base.apk?rgh-link-date=2021-05-25T16%3A50%3A57Z) is only 230 kilobytes. Could be worth looking at for a few mins to see if there's something obvious we can do to reduce size.

I had a look around today and found [this blog post](https://github.com/fractalwrench/ApkGolf/blob/master/blog/BLOG_POST.md) which helped me make a few small changes to the default Android app template (without any of the hacks) and got the APK size down to 11kb (so fairly significant reduction!). I've pushed it to a repo: https://github.com/digidem/minimal-android. It would be nice to create a script that generates APKs with different version names, version codes, minSdks and applicationIds for our tests. I can take a look at that later in the week.

FYI this is a useful tool for quickly inspecting an APK without Android Studio: https://apkanalyzer.netlify.app
I set that up on my own netlify account because the original is no longer working: https://github.com/bundlecop/apkanalyzer.io

src/backend/test/helpers.js
gmaclennan: Good idea, will add this to the todo.

src/backend/lib/types.d.ts
gmaclennan: This selects the type of the property `arch` from the object/interface `InstallerExt`. Typescript does not support "dot" notation to access properties like you would on a Javascript object (e.g. `myObject.prop`, so you have to use `myObject["prop"]` when referring to props.

TL;DR this says that the `supportedAbis` prop has the same type as the `arch` prop of the `InstallerExt` type.

scripts/build-upgrade-test-apks.sh
```diff
@@ -0,0 +1,35 @@
...
+  mv "$apk_dir/mapeo-qa-release.apk" "$apk_dir/mapeo-v${1}.${2}.apk"
+}
+
+# Build backend
+"${dir0}/build-backend.sh"
+
+# Build a QA Release APK with version name and version code
+build_apk 5.3.0 2000
+build_apk 5.4.0 2010
+build_apk 6.0.0 2020
```
hackergrrl: Is the idea to manually edit these numbers when we do releases? This could be fragile!

src/backend/bin/upgrade-simulator.js
```diff
@@ -1,43 +1,40 @@
...
-  if (err) return onError("getport", err);
-
-  const upgradeManager = new UpgradeManager({
-    dir,
-    port,
-    version,
+const { getInstallerInfo } = require("../lib/utils");
+
+const log = debug("p2p-upgrades:simulator");
+if (!process.env.DEBUG) debug.enable("p2p*");
```
hackergrrl, line 14: Enable debug logging if NOT in debug mode?

src/backend/bin/upgrade-simulator.js
```diff
@@ -1,43 +1,40 @@
...
+
+tmp.setGracefulCleanup();
```
hackergrrl, line 16: 💪🏻 😤

gmaclennan:
@luandro, this should be ready for testing on a real device now. You will need to generate some Mapeo APKs with different versions in order to use the simulator, because the code now can't be fooled by advertising the APK as a different version, because it now checks the actual version of each APK.

If you are running the debug variant (which is what you do when you run `npm run android`), you can generate some different-versioned debug APKs to upgrade / downgrade with:

```sh
mkdir -p test-apks
cd android
ANDROID_VERSION_NAME=5.0.0 ANDROID_VERSION_CODE=1 ./gradlew assembleAppDebug
mv app/build/outputs/apk/app/debug/mapeo-app-debug.apk ../test-apks/mapeo-v5.0.0.1.apk
ANDROID_VERSION_NAME=6.0.0 ANDROID_VERSION_CODE=10 ./gradlew assembleAppDebug
mv app/build/outputs/apk/app/debug/mapeo-app-debug.apk ../test-apks/mapeo-v6.0.0.10.apk
```

That gives you two versions to test with: `v5.0.0` with a versionCode (build number) of `1` and `v6.0.0` with a versionCode `10`.

Now run Mapeo on your device specifying the exact version you want it to appear as:

```sh
ANDROID_VERSION_NAME=5.2.0 ANDROID_VERSION_CODE=5 npm run android
```

This will start Mapeo on your device as `v5.2.0` and versionCode `5`.

Now run the simulator with the older APK:

```sh
node ./src/backend/bin/upgrade-simulator.js test-apks/mapeo-v5.0.0.1.apk
```

You should see the device sharing its APK to the simulator, and in the simulator logs you should see the upgrade become available.

To test in the other direction (sharing a newer version from the simulator to the device):

```sh
node ./src/backend/bin/upgrade-simulator.js test-apks/mapeo-v6.0.0.10.apk
```

You should see the device download this version, then the UI will change to show an INSTALL button. This button should work to install the new version.

Let me know how (if!) this works for you, then we can roll out to the wider team for their testing. Because they won't be testing in the debug environment, I've created a script that will generate a few QA variant APKs with different version numbers for them to test:

```sh
./scripts/build-upgrade-test-apks.sh
```

We can run this when ready to generate a few APKs that the program team can install on different devices to test.

hackergrrl:
@gmaclennan Your notes here on how to build & set up the simulator and device together seem really useful & important! This could make great notes for a `README.md` or `HACKING.md` in the `backend/` dir.

src/backend/lib/upgrade-server.js
```diff
@@ -1,276 +1,195 @@
...
-const PROGRESS_THROTTLE_MS = 400; // milliseconds
-
-class UpgradeServer extends EventEmitter {
-  constructor(storage, port) {
+const EMIT_THROTTLE_MS = 400; // milliseconds
+
+// Only use logger in development when debugging
+// const serverLogger = process.env.DEBUG
+//   ? require("pino")({ prettyPrint: { singleLine: true } })
+//   : false;
```
hackergrrl, line 28: Dead code?

src/backend/server.js
```diff
@@ -173,30 +169,45 @@ function createServer({
...
-      rnBridge.channel.on("p2p-upgrade::stop-services", () =>
-        manager.stopServices()
-      );
-      rnBridge.channel.on("p2p-upgrade::get-state", () => manager.getState());
-      cb(null, manager);
+  function createUpgradeManager() {
+    // TODO: What is the effect of leaving these event handlers attached when
+    // the app is in the background? Ideally we would clean up event handlers
+    // each time the app goes to the background and re-attach them when back in
+    // foreground.
```
hackergrrl, line 176: Consider filing this as an issue for better visibility & to make discussion easier.

src/frontend/api.js
```diff
@@ -98,6 +97,47 @@ export type ServerPeer = {
 
 type PeerHandler = (peerList: Array<ServerPeer>) => any;
 
+// These types are manually copied from `backend/lib/types.d.ts`. This is far
+// from ideal and is a workaround until we can convert the entire codebase to
+// Typescript - right now frontend uses Flow.
```
hackergrrl, line 102: Consider filing an issue for this work :)

hackergrrl: Also consider putting a comments in `backend/lib/types.d.ts` -- maybe in all-caps -- telling future devs to make sure any changes they make there are reflected in this file!

src/frontend/api.js
```diff
@@ -217,18 +257,21 @@ export function Api({
       nodejs.start("loader.js");
       const serverStartPromise = new Promise(resolve =>
         nodejs.channel.once("status", resolve)
-      ).then(() => {
+      ).then(async () => {
```
hackergrrl, line 263: Is `async` required for this function? It doesn't `await` anywhere.

src/frontend/screens/SyncModal/UpgradeBar.js
```diff
@@ -0,0 +1,264 @@
+// @flow
```
hackergrrl, line 1: Yay! This code refactored into its own file! 🌮

src/backend/test/upgrade-discovery.test.js
```diff
@@ -0,0 +1,299 @@
...
+const hasha = require("hasha");
+const { promises: fsPromises } = require("fs");
+const { setupStorage, hashCmp } = require("./helpers.js");
+const Discovery = require("../lib/upgrade-discovery");
+const Server = require("../lib/upgrade-server");
+const fakeApkInfo = require("./fixtures/fake-apk-info");
+const _ = require("lodash");
+const getStream = require("get-stream");
+
+const validApksFolder = path.join(__dirname, "./fixtures/valid-apks");
```
hackergrrl, line 16: Doesn't `test/helpers.js` export this path already?

src/backend/test/upgrade-storage.test.js
```diff
@@ -0,0 +1,297 @@
...
+const pipeline = promisify(stream.pipeline);
+const fakeApkInfo = require("./fixtures/fake-apk-info");
+const { readJson, hashCmp, setupStorage } = require("./helpers.js");
+const fsPromises = fs.promises;
+
+/** @typedef {import('../lib/types').InstallerInt} InstallerInt */
+
+tmp.setGracefulCleanup();
+
+const validApksFolder = path.join(__dirname, "./fixtures/valid-apks");
```
hackergrrl, line 22: Doesn't `test/helpers.js` export this path already?

src/backend/test/upgrade-storage.test.js
```diff
@@ -0,0 +1,297 @@
...
+const invalidApksFolder = path.join(__dirname, "./fixtures/invalid-apks");
```
hackergrrl, line 23: Should `test/helpers.js` export this also?

src/backend/test/utils/getInstallerInfo.test.js
```diff
@@ -0,0 +1,41 @@
+const { getInstallerInfo } = require("../../lib/utils");
+const test = require("tape");
+const path = require("path");
+const { promises: fs } = require("fs");
+
+test("getInstallerInfo: read valid APKs", async t => {
+  const validApksFolder = path.join(__dirname, "../fixtures/valid-apks");
```
hackergrrl, line 8: Doesn't `test/helpers.js` export this path already?

src/backend/test/utils/getInstallerInfo.test.js
```diff
@@ -0,0 +1,41 @@
...
+    };
+    t.deepEqual(info, expected, "Info matches expected data");
+  }
+});
+
+test("getInstallerInfo: invalid APKs", async t => {
+  const invalidApksFolder = path.join(__dirname, "../fixtures/invalid-apks");
+  const apkFilepaths = (await fs.readdir(invalidApksFolder))
+    .filter(file => path.extname(file) === ".apk")
+    .map(file => path.join(invalidApksFolder, file));
```
hackergrrl, line 27: If you're using this pattern a lot, it could stand to be a parsing job that helpers.js does on init + exports.

src/backend/test/utils/getInstallerInfo.test.js
```diff
@@ -0,0 +1,41 @@
...
+  for (const filepath of apkFilepaths) {
+    await getInstallerInfo(filepath).catch(e => {
+      t.true(e instanceof Error, "Throws with error");
+    });
+  }
+});
+
+async function readJson(filepath) {
+  return JSON.parse(await fs.readFile(filepath));
+}
```
hackergrrl: `test/helpers.js` exports this!

src/backend/test/upgrade-storage.test.js
```diff
@@ -0,0 +1,297 @@
...
+  t.deepEqual(
+    installersFromList,
+    [fakeApkInfo.v0_0_1],
+    "Starts with only current apk"
+  );
+  t.deepEqual(
+    installersFromEvent,
+    [fakeApkInfo.v0_0_1],
+    "Starts with only current apk"
+  );
```
hackergrrl, line 128: I really like how easy + readable it is to express this kind of logic!

src/backend/test/upgrade-storage.test.js
```diff
@@ -0,0 +1,297 @@
...
+
+  const ws = storage.createWriteStream({ hash: "anything" });
+  try {
+    await pipeline(fs.createReadStream("non-existent-file.apk"), ws);
+  } catch (e) {
+    t.ok(e instanceof Error, "throws an error");
+  }
```
hackergrrl, line 182: Consider using e.g. `t.plan()` to set how many positive signals you expect. Otherwise, if the pipeline **doesn't** throw, this test will still pass.

src/backend/test/upgrade-storage.test.js
```diff
@@ -0,0 +1,297 @@
...
+  const testApkFilepath = path.join(
+    validApksFolder,
+    "com.example.test_SDK21_VN1.0.0_VC1.apk"
+  );
+  const ws = storage.createWriteStream({ hash: "invalid-hash" });
+  try {
+    await pipeline(fs.createReadStream(testApkFilepath), ws);
+  } catch (e) {
+    t.ok(e instanceof Error, "writeStream throws");
+  }
```
hackergrrl, line 212: Same feedback as the previous test.

src/backend/test/upgrade-storage.test.js
```diff
@@ -0,0 +1,297 @@
...
+    fakeApkInfo.v0_0_1
+  );
+
+  for (const filepath of invalidApkFilepaths) {
+    const hash = await hasha.fromFile(filepath, { algorithm: "sha256" });
+    const ws = storage.createWriteStream({ hash });
+    try {
+      await pipeline(fs.createReadStream(filepath), ws);
+    } catch (e) {
+      t.ok(e instanceof Error, "throws an error");
```
hackergrrl, line 278: Consider using e.g. t.plan() to set how many positive signals you expect. Otherwise, if the pipeline doesn't throw, this test will still pass.

src/backend/test/upgrade-server.test.js
```diff
@@ -0,0 +1,204 @@
...
+const fs = require("fs");
+const { promisify } = require("util");
+const stream = require("stream");
+const fakeApkInfo = require("./fixtures/fake-apk-info");
+const getPort = require("get-port");
+const { readJson, hashCmp } = require("./helpers.js");
+const fsPromises = fs.promises;
+const got = require("got").default;
+
+tmp.setGracefulCleanup();
```
hackergrrl, line 17: Can this be done just *once* in `helpers.js`? It should affect all test files (that include the helpers.js file, which I think is all of them).

src/backend/test/upgrade-server.test.js
```diff
@@ -0,0 +1,204 @@
...
+  const app = server.getFastify();
+
+  const inject = app.inject.bind(app);
+  const cleanup = async () => {
+    await server.stop();
+    await tmpDir.cleanup();
+  };
+  const { address, port } = /** @type {import('net').AddressInfo} */ (start
+    ? app.server.address()
+    : { address: "localhost", port: 80 });
```
hackergrrl: Is this code path used anywhere? The process can't bind to port 80 without root privileges, which is unlikely given this is a test suite.

src/backend/test/upgrade-server.test.js
```diff
@@ -0,0 +1,204 @@
...
+    fakeApkInfo.v0_0_1
+  );
+  const { filepath: _1, ...expected1 } = fakeApkInfo.v0_0_1;
+  // @ts-ignore
+  expected1.url = `${baseUrl}/installers/${expected1.hash}`;
+
+  const apkFilepath = path.join(
+    __dirname,
+    "./fixtures/valid-apks/com.example.test_SDK21_VN1.0.0_VC1.apk"
+  );
```
hackergrrl, line 70: Import `validApksFolder` ;)

src/backend/test/upgrade-server.test.js
```diff
@@ -0,0 +1,204 @@
...
+  const { cleanup, baseUrl } = await startServer(fakeApkInfo.v0_0_1, true);
+
+  // Check the server is actually running first
+  await got(`${baseUrl}/installers`);
+  t.pass("Initial request ran without error");
+
+  await cleanup();
+
+  try {
+    await got(`${baseUrl}/installers`);
```
hackergrrl, line 143: Test will still pass if this does not throw.

src/backend/test/upgrade-server.test.js
```diff
@@ -0,0 +1,204 @@
...
+
+  // Check the server is actually running first
+  await got(`${baseUrl}/installers`);
+  t.pass("Initial request ran without error");
+
+  // Don't await, so got() happens before this finishes
+  cleanup();
+
+  try {
+    await got(`${baseUrl}/installers`);
```
hackergrrl, line 162: Test will still pass if this does not throw.

src/backend/test/upgrade-server.test.js
```diff
@@ -0,0 +1,204 @@
...
+
+  // Check the server is actually running first
+  await got(`${baseUrl}/installers`);
+  t.pass("Initial request ran without error");
+
+  await server.stop();
+
+  try {
+    // Server is stopped, this should throw
+    await got(`${baseUrl}/installers`);
```
hackergrrl, line 181: Test will still pass if this does not throw.

hackergrrl (commented):
These are my final review comments. Great tests @gmaclennan! Very easy to read and follow. I like the way you've set it all up.

src/backend/test/upgrade-server.test.js
gmaclennan: I'm using fastify.inject, which mocks this to `localhost:80`. On reflection I don't like this though: it requires reading the fastify docs and it's not a test of a real-world example. I'm going to change these tests to actually start the server on a port and do a proper request.

src/backend/test/upgrade-server.test.js
gmaclennan: Not sure, risky given how different test runners might work.

src/frontend/api.js
gmaclennan: `deviceInfo` awaits on line 272

src/backend/lib/upgrade-server.js
gmaclennan: Yes, but useful to keep for uncommenting during tests.

scripts/build-upgrade-test-apks.sh
gmaclennan: This is just for testing. Version names are arbitrary, just needs to create 3 different versions for testing.

gmaclennan:
I identified 2 bugs in my own QA testing:

- A device would show "No app updates found / checked 1 device" for a second or two before an update started downloading. E.g. it was reporting a device was checked before it had actually checked it.
- The checked devices list would not be reset when navigating away from and returning to the sync screen.

These bugs are fixed now in this PR, and there are more comprehensive tests for the "checkedPeers" functionality.

PR #596, opened 2021-05-20
Determine update candidate based on new version naming scheme

luandro:
Solves #578

- [x] 1.0.0 should update to 1.0.1 and 1.2.0 but not to 1.0.0-RC
- [x] 1.1.0-RC+9a6731e should update to 1.1.0 and to 1.1.0-RC+19d986e if the build number of the latter is higher.
- [x] 1.0.0-PR571+eac5a0f should update to 1.0.0-PR571+afc1a1b if the build number of the latter is higher. It should not update to 1.0.0-PR572+4667afc or any other version that does not match /$1\.0\.0-PR571/.
- [x] Check for different applicationId
- [x] Check for incompatible minSdk
- [x] Check for invalid semver
- [x] Check for invalid sha7
- [x] Check same hash
- [x] Check for 0 size
- [x] Check for platform !== Android
- [x] Check for versionCodes


gmaclennan:
Hi Luandro, this looks like it is on the right track. For tests like this, the pattern I prefer is to maintain some kind of folder(s) of JSON files with the data to be tested and expected results, and keep the test to a loop over the data. See the test https://github.com/digidem/mapeo-mobile/blob/p2p-update-typescript/update-candidate/src/backend/test/utils/getInstallerInfo.test.js for an example.

This way it is easy to manage adding new examples to test without copying a whole bunch of test boilerplate into a new test function.

gmaclennan:
e.g. you might have a folder called `valid-upgrade-candidates` with JSON files:

```json
{
  "currentApkInfo": {...},
  "deviceInfo": {...},
  "installer": {...}
}
```

Then the test should loop over those JSON files and check that `isUpgradeCandidate()` returns true for all of them.

Similarly you could do the same with a folder `not-upgrade-candidates` with the same structure, but where each file is an example of arguments that should return false.

src/backend/test/utils/invalid-upgrade-candidates/v100PR_v100PR2.json
```diff
@@ -0,0 +1,30 @@
+{
```
gmaclennan: suggested filename: `RC-sameVN-sameVC.json` e.g. two release candidates with the same version name and same version code. Invalid candidate because RC should only update if version code is higher.

src/backend/test/utils/valid-upgrade-candidates/v100PR_v100PR2.json
```diff
@@ -0,0 +1,30 @@
...
+    "filepath": "non-existent.apk"
+  },
+  "deviceInfo": {
+    "supportedAbis": ["arm64-v8a"],
+    "sdkVersion": 1
+  },
+  "installer": {
+    "hash": "810ff2fb242a5dee4220f2cb0e6a519891fb67f2f828a6cab4ef8894633b1f50",
+    "size": 8,
+    "versionName": "1.0.0-PR572+4667afc",
```
gmaclennan: This is not a valid upgrade candidate because it is from a different PR

src/backend/test/utils/valid-upgrade-candidates/v100RC_v100RC2.json
```diff
@@ -0,0 +1,30 @@
...
+    "supportedAbis": ["arm64-v8a"],
+    "sdkVersion": 1
+  },
+  "installer": {
+    "hash": "810ff2fb242a5dee4220f2cb0e6a519891fb67f2f828a6cab4ef8894633b1f50",
+    "size": 8,
+    "versionName": " 1.1.0-RC+19d986e",
+    "minSdkVersion": 1,
+    "applicationId": "com.example",
+    "versionCode": 1,
```
gmaclennan: This is not a valid upgrade candidate, because although it comes from the same RC (`1.1.0-RC`) and it is different (because the commit hash is different) the version code is the same. We should only count as valid with a higher version code, otherwise we would get in a loop back and forth.

gmaclennan (changes requested):
Having it written this way is really helpful for reviewing the fixtures that we consider valid and invalid.

Modifying my suggestion earlier, I think this might be easier if invalid and valid are `.js` files that each export an array, that way we can:

1. Add `//@ts-check` to catch type errors
2. Add comments to explain why each is valid/invalid

luandro:
Should builds from same PR and same versionCode be considered candidates? Or only if PRs are the same and versionCode is higher?

gmaclennan:
Only if higher, otherwise we could get an infinite loop of upgrading from from one to the other.

gmaclennan:
Actually in the tests, for each input we should check that swapping the installer and cureentapkinfo returns the opposite result.

luandro:
Fixed PR check logic to only return candidate is same PR and versionCode is higher.

Swaped installer and cureentapkinfo on the valid-candidates to make sure results are opposite.

src/backend/lib/utils.js
```diff
@@ -23,6 +24,37 @@ module.exports = {
...
 };
 
+/**
+ * Check if PR is the same
+ *
+ * @param {string} a
+ * @param {string} b
+ * @returns {boolean}
+ */
+function checkSamePR(a, b) {
```
gmaclennan: Need to check a,b are valid semver first otherwise semver.parse will throw. Use semver.valid()

src/backend/lib/utils.js
```diff
@@ -64,16 +111,42 @@ function installerCompare(a, b) {
 function isUpgradeCandidate({ deviceInfo, installer, currentApkInfo }) {
   // Filter out the current APK
   if (installer.hash === currentApkInfo.hash) return false;
+  // Check for 0 or less APK size
```
gmaclennan, line 138: We pre-check all installers that they are valid APKs (at least, we check they are a zip file and have an `AndroidManifest.xml` file), so this step is unnecessary, but no harm leaving it here in just in case!

src/backend/lib/utils.js
```diff
@@ -64,16 +111,42 @@ function installerCompare(a, b) {
...
   // TODO: 64 bit devices can run 32 bit APKs
   const isSupportedAbi = !!installer.arch.find(arch =>
     deviceInfo.supportedAbis.includes(arch)
   );
   if (!isSupportedAbi) return false;
+  log(
+    "installer.minSdkVersion > deviceInfo.sdkVersion",
+    installer.minSdkVersion > deviceInfo.sdkVersion
+  );
   if (installer.minSdkVersion > deviceInfo.sdkVersion) return false;
```
gmaclennan: This is going to make the debug logs super noisy (and harder to parse in my opinion) because this function is called multiple times. I suggest we only log the reason an installer is rejected, e.g. only log just before returning `false` like:

```suggestion
  if (installer.minSdkVersion > deviceInfo.sdkVersion) {
  	log(`installer rejected: minSdkVersion ${installer.minSdkVersion} is not supported by this device`)
    return false;
  }
```

src/backend/lib/utils.js
```diff
@@ -34,12 +66,27 @@ module.exports = {
...
   if (!aIsValidSemver) return -1;
   if (!bIsValidSemver) return 1;
+  // Check if semver commit is using {sha7}
+  log(
+    "checkVersionName",
+    checkVersionName(a.versionName),
+    checkVersionName(b.versionName)
+  );
+  if (!checkVersionName(a.versionName)) return -1;
+  if (!checkVersionName(b.versionName)) return -1;
```
gmaclennan: I think the code should ignore everything after the `+`. This is just metadata that we use when debugging to identify which commit a version comes from. I don't think it should affect any upgrade logic, and we could change it in the future.

src/backend/lib/utils.js
```diff
@@ -34,12 +66,27 @@ module.exports = {
...
+  // Check if is PR and is the same
+  const prCompare = checkSamePR(a.versionName, b.versionName);
+  log("PR Compare", prCompare);
+  if (!prCompare) return 0;
```
gmaclennan: I don't think this code belongs in this function: `installerCompare` is meant to sort installers by most recent. This code should be in the `isUpgradeCandidate` function.

src/backend/lib/utils.js
```diff
@@ -64,16 +111,42 @@ function installerCompare(a, b) {
...
+  log(
+    `isUpgradeCandidate ~ installer.platform !== "android"`,
+    installer.platform !== "android"
+  );
   if (installer.platform !== "android") return false;
+  log(
+    "installer.versionCode < currentApkInfo.versionCode",
+    installer.versionCode < currentApkInfo.versionCode
+  );
   if (installer.versionCode < currentApkInfo.versionCode) return false;
```
gmaclennan:
```suggestion
  if (installer.versionCode <= currentApkInfo.versionCode) return false;
```

We need to check `<=` not `<` to avoid recursive updates or internal builds or RCs.

src/backend/test/utils/invalid-upgrade-candidates/differing_applicationId.json
```diff
@@ -0,0 +1,30 @@
...
+    "filepath": "non-existent.apk"
+  },
+  "deviceInfo": {
+    "supportedAbis": ["arm64-v8a"],
+    "sdkVersion": 1
+  },
+  "installer": {
+    "hash": "810ff2fb242a5dee4220f2cb0e6a519891fb67f2f828a6cab4ef8894633b1f50",
+    "size": 8,
+    "versionName": "1.2.0",
```
gmaclennan: This file is titled `differing_applicationId.json` but the applicationId is the same. The versionName is different, and this looks like it should be a valid upgrade candidate, not an invalid one.

src/backend/test/utils/invalid-upgrade-candidates/different_PR.json
```diff
@@ -0,0 +1,30 @@
...
+    "platform": "android",
+    "arch": ["arm64-v8a"],
+    "filepath": "non-existent.apk"
+  },
+  "deviceInfo": {
+    "supportedAbis": ["arm64-v8a"],
+    "sdkVersion": 1
+  },
+  "installer": {
+    "hash": "810ff2fb242a5dee4220f2cb0e6a519891fb67f2f828a6cab4ef8894633b1f50",
```
gmaclennan: A lot of these invalid-upgrade-candidates are being considered invalid because their hashes match (`currentApkInfo.hash === installer.hash`), so we're not checking what we mean to be checking. This one seems to differ by a single character somewhere, but that's hard to check visually. Can we ensure the hashes for `currentApkInfo` and `installer` are completely different for all the tests, apart from the one that checks hash similarity?

src/backend/test/utils/invalid-upgrade-candidates/incompatible_VC.json
```diff
@@ -0,0 +1,30 @@
...
+    "filepath": "non-existent.apk"
+  },
+  "deviceInfo": {
+    "supportedAbis": ["arm64-v8a"],
+    "sdkVersion": 1
+  },
+  "installer": {
+    "hash": "810ff2fb242a5dee4220f2cb0e6a519891fb67f2f828a6cab4ef8894633b1f50",
+    "size": 8,
+    "versionName": "1.2.0",
```
gmaclennan: Let's check version code in isolation.

```js
currentApkInfo.versionName = "1.1.0-RC+9a6731e"
currentApkInfo.versionCode = 2

installer.versionName = "1.1.0-RC+feg783a"
installer.versionCode = 1
```

This should also be invalid if `installer.versionCode === 2`

src/backend/test/utils/invalid-upgrade-candidates/incompatible_minSdk.json
```diff
@@ -0,0 +1,30 @@
...
+    "filepath": "non-existent.apk"
+  },
+  "deviceInfo": {
+    "supportedAbis": ["arm64-v8a"],
+    "sdkVersion": 1
+  },
+  "installer": {
+    "hash": "810ff2fb242a5dee4220f2cb0e6a519891fb67f2f828a6cab4ef8894633b1f50",
+    "size": 8,
+    "versionName": "1.2.0",
```
gmaclennan: Let's not check differing version names at the same time as other things.

src/backend/test/utils/invalid-upgrade-candidates/invalid_semver_sha7.json
```diff
@@ -0,0 +1,30 @@
...
+    "filepath": "non-existent.apk"
+  },
+  "deviceInfo": {
+    "supportedAbis": ["arm64-v8a"],
+    "sdkVersion": 1
+  },
+  "installer": {
+    "hash": "810ff2fb242a5dee4220f2cb0e6a519891fb67f2f828a6cab4ef8894633b1f50",
+    "size": 8,
+    "versionName": "1.1.0-RC+19d986e1",
```
gmaclennan: Can drop this test, I don't think we want to be strict about this sha7, it's just metadata for our own use in debugging.

src/backend/test/utils/invalid-upgrade-candidates/unsupported_abi.json
```diff
@@ -0,0 +1,30 @@
+{
+  "currentApkInfo": {
+    "hash": "810ff2fb242a5dee4220f2cb0e6a519891fb67f2f828a6cab4ef8894633b1f50",
```
gmaclennan: same hash

src/backend/test/utils/valid-upgrade-candidates/RC_newerPatch.json
```diff
@@ -0,0 +1,30 @@
...
+    "filepath": "non-existent.apk"
+  },
+  "deviceInfo": {
+    "supportedAbis": ["arm64-v8a"],
+    "sdkVersion": 1
+  },
+  "installer": {
+    "hash": "810ff2fb242a5dee4220f2cb0e6a519891fb67f2f828a6cab4ef8894633b1f50",
+    "size": 8,
+    "versionName": "1.1.1-RC+19d986e",
```
gmaclennan: This is not a valid upgrade candidate. See https://github.com/digidem/mapeo-mobile/issues/578

> **Release Candidates**: Should update to the next RC that shares the same `{major}.{minor}.{patch}`, and any production release that is semver => `{major}.{minor}.{patch}` of the RC.

E.g. RC version should only update to another RC version if the main semver component is equal.
This RC does not share the same `{major}.{minor}.{patch}`, so it should not update. However an RC should update to a non-RC that is semver greater.

gmaclennan (changes requested):
Looking good. Left a few comments to resolve. Big issue with hashes being the same for lots of the "invalid" fixtures, which is leading them to pass tests as invalid, but for the wrong reasons (hash being the same).

src/backend/lib/utils.js
luandro: Added semver validation before, but TS still acuses that object might still be null.

gmaclennan: `semver.valid(versionString)` returns `null` if `versionString` is not valid semver. Once you have confirmed the result of `semver.valid(versionString)` is not null or undefined, then you can safely call `semver.parse()`. TS is likely telling you the truth here :)

src/backend/test/utils/invalid-upgrade-candidates/incompatible_VC.json
luandro: Not 100% clear. Changed to simple version names (1.0.0 and 1.0.1) to make sure versionCode is failing.

luandro: Not 100% clear. Changed to simple version names (1.0.0 and 1.0.1) to make sure versionCode is failing.

src/backend/test/utils/valid-upgrade-candidates/RC_newerPatch.json
luandro: > E.g. RC version should only update to another RC version if the main semver component is equal.

So `1.0.0-RC+9a6731e` should upgrade to `1.0.0-RC+19d986e`, but how does it know it's a newer RC?

luandro: > The versionCode / build number should be used to determine updates if the two RC releases have the same version (but different +{sha7}). The higher build number "wins".

How do we determine higher build number?

gmaclennan:
versionCode = buildNumber

src/backend/lib/utils.js
luandro: Really got to see the power of TS with this!

luandro:
Thanks for the review @gmaclennan, was still a bit confused about the pre-release checking. Did a refactor and I think it's working and testing as expected.

src/backend/test/utils/invalid-upgrade-candidates/incompatible_VC.json
luandro: Came to me in a dream lol

Issue #313, opened 2020-02-17
Reduce APK size


gmaclennan:
Currently the Mapeo APK size is 89.1Mb. Reducing the size will reduce install time for users with slow internet, and reduce space taken up on the device.

The main cause of the large size is supporting different architectures in the same APK. One way to resolve this is to distribute the app through the Google Play store, and uploading an ["Android App Bundle" (aab)](https://developer.android.com/guide/app-bundle) and allowing Google Play to generate an APK depending on the architecture of the device.

However, this does not solve the problem for users downloading an APK for offline install. It will likely cause confusion for users if there are different versions, since many users do not know what architecture their phone has.

The main two architectures for Android are Arm v7, which is 32-bit, and Arm_64, which is 64-bit. The APK currently includes x86 and x86_64 builds. This is useful for running in an emulator on a x86 machine, but few (any?) phones run on x86. I _think_ it is mainly ChromeBooks (which now run Android apps) which would require x86.

Looking at the bundle size of a [current APK](https://github.com/digidem/mapeo-mobile/releases/download/v2.0.0-beta.4/Mapeo-v2.0.0-beta.4.apk) with [https://apkanalyzer.xyz](https://apkanalyzer.xyz):
 
- x86 and x86_64 libs are a total of 38.15Mb - these could be removed
- nodejs assets are about 9.97Mb. A lot of this seems to be leveldb prebuilds for different architectures - a lot of these could be removed, probably saving about 7Mb
- Most of the lib size is `libv8android.so` (7.23 MB), `libnode.so` (6.52 MB) and `libmapbox-gl.so` (1.79 MB). `libv8android.so` is large because of the bundled Intl, which is required for multi-language support. The other two are also needed.
- Java compiled code (`.dex`) is about 4Mb which could probably be reduced by using ProGuard to remove unused code, but it is not a large % of current size.
 - The javascript bundle is currently 463.96 KB for the frontend and 1.08Mb for the backend. The backend could maybe be reduced by tree-shaking, but these are not a large % of size.

Therefore, probably the best and easiest way to reduce size is to only build the APK for Arm (32-bit and 64-bit), and then strip un-used code from Nodejs assets. This would get the APK size down to about 43Mb. If we built per-architecture APKs for Google Play they would probably be about 23Mb each. Further work with ProGuard and reducing unused JS code could maybe bring this down to about 40/20Mb.

arky:
+1 

This was also requested by the users at our Thailand training programs.

[...]

PR #606, opened 2021-05-27
chore: Reduce APK size

gmaclennan:
Some changes to reduce APK size.

Current size (arm7 & arm64): 62.6Mb

- [x] Remove leveldown prebuilds (not used): 58.9Mb
- ~~Minify and shrink resources: 56.6Mb~~
- ~~Fix e2e tests~~

Overall this PR reduces the app size by 6Mb ~10%. Further steps for reducing size will be documented in #313


luandro:
Got a release apk with 53Mb. Working great, good job!

gmaclennan:
This is breaking the end-to-end tests due to proguard stripping the symbols that Detox requires to run tests. I've followed these steps but still failing: https://github.com/wix/Detox/blob/master/docs/Introduction.Android.md#7-proguard-minification-obfuscation

gmaclennan:
Removing the proguard minification from this PR because it was breaking e2e tests. Will deal with proguard in a separate PR #645. The change in this PR with just removing leveldb prebuilds is enough of a saving (4Mb) to make it worth it on its own.

PR #633, opened 2021-07-09
feature: add support for manual coordinate entry with dd and dms formats

### LINKED ISSUE #632: Format for manual coordinate
Part of the tech prio meeting : https://www.notion.so/digidem/format-for-manual-coordinate-entry-fix-7c103b7daf474e2290ce4c8e0d6ef57c

8pts

### LINKED ISSUE #641: Longitude and Latitude as default Coordinate system
The most common request we get while using Mapeo in SE. Asia is the ability to enter the observations manual co-ordinates using Latitude and Longitude (in places where GPS fix is not acquired).  Mapeo currently defaults to UTM system. 

Such Geo-coridnates could be taken from hand-held GPS systems that are more accurate and able get fixes under dense canopy and shadow of hill or mountain.

achou11:
Fixes #632 
Fixes #641 

Notes:
- adds a new dependency for a picker component: [`@react-native-picker/picker`](https://github.com/react-native-picker/picker)
  - seems to be the go-to solution for RN but it's pretty hard to work with. I created a shared component called `Select` that adds a simple and opinionated layer around the library that should be updated if we ever migrate away from this specific dep
- Refactors the screen to add the following:
  - user can select which format to use for the manual entry. the selection is initially determined by the app-wide setting they have, but otherwise becomes independent and persisted separately
  - there are three "forms" based on the selected format i.e. utm, dd, or dms
  - I didn't make many changes to the already implemented UTM form. just updates to work with the reimplemented validation/saving model
- for selecting the cardinality, I opted to use the library's `dropdown` mode instead of the `dialog` mode since having that inlined in the form helps to prevent loss of context for what you're editing. curious if there are any strong thoughts against this and whether or not we should stick to just the dialog for all use-cases
- adds unit tests for the forms in `/ManualGpsScreen` (see `ManualGpsScreen.test.tsx`)
  - pretty comprehensive for the `DdForm` and `DmsForm`, but definitely lacking tests for the following: 
    - `UtmForm`: mostly because I didn't try to change too much about the input behavior in this component, so that would need some refinement if it were subject to similar tests. I don't have the best understanding of UTM format and its conversions so wasn't as ready to completely refactor that form
    - `ManualGpsScreen`: mostly because it requires setup that's specific to React Navigation and I was running into some types issues when attempting to do that. The only meaningful tests for this I can imagine are: 
      - making sure the switch component properly changes the form displayed (this might be a better candidate for e2e/acceptance tests though)
      - ensuring that the draft is saved when we press the save button (might be better to do e2e for this instead too)
- converts some files to TS:
  - `/types.js` (just ported, not full conversion)
  - `/sharedComponents/HeaderTitle.js`
  - `/sharedComponents/IconButton.js`
  - `/sharedComponents/Text.js`

---

Preview:

- Decimal degrees:

image

- Degrees/Minutes/Seconds:

image

- Universal Transverse Mercator:

image

- Coordinate format selection example:

image

- Cardinality selection example:

image


src/frontend/screens/ManualGpsScreen/DdForm.js
```diff
@@ -0,0 +1,247 @@
...
+const DIRECTION_OPTIONS_EAST_WEST = [
+  {
+    value: "E",
+    label: m.east.defaultMessage,
+  },
+  {
+    value: "W",
+    label: m.west.defaultMessage,
+  },
+];
```
achou11: Can translations be generally used like this?

src/frontend/screens/ManualGpsScreen/DmsForm/index.js
```diff
@@ -0,0 +1,243 @@
...
+        case "seconds":
+          return {
+            ...previous,
+            seconds: value,
+          };
+        default:
+          return previous;
+      }
+    });
+  };
```
achou11: ideally this could've just been something like `update(previous => ({...previous, [unit]: value})` but Flow was complaining. I've done this in TS without much issue so a little unfortunate, but this is pretty explicit so could be worse 🤷‍♂️

ErikSin: How much time do you think it would take to refactor these flow pages in to ts pages?

achou11: Wouldn't take long but just didn't want to worry about setting up TS, which you're handling in https://github.com/digidem/mapeo-mobile/pull/634

src/frontend/screens/ManualGpsScreen/DdForm.js
gmaclennan: No, because when you change the app language at runtime then this will not change. Best to define these constants inside the component, and then use:

```js
const { formatMessage } = useIntl()
```

And then for the label use:

```js
{
  value: "E"
  label: formatMessage(m.east)
}
```

This will, at runtime, look up the language and look up the translated value, with the fallback to the `defaultMessage` if no translation is defined (or app is in English).

src/frontend/screens/ManualGpsScreen/UtmForm.js
```diff
@@ -281,17 +232,7 @@ function toLatLon({ zoneNum, zoneLetter, easting, northing }: State) {
   );
 }
 
-function parseNumber(str: string): number | void {
-  const num = Number.parseFloat(str);
-  if (Number.isNaN(num)) throw new Error("Coordenada no válido");
```
achou11, line 286: Got rid of this line in the util defined in `shared.js`. Is that gonna be okay?

src/frontend/screens/ManualGpsScreen/DdForm.tsx
```diff
@@ -0,0 +1,250 @@
...
+    }
+  };
+
+  const formatInputValue = (setState: (v: string) => void) => ({
+    nativeEvent: { text },
+  }: NativeSyntheticEvent<TextInputEndEditingEventData>) => {
+    const parsed = Number.parseFloat(text);
+    if (!isNaN(parsed)) {
+      // Do we want to enforce precision here?
+      setState(parsed.toFixed(6));
```
achou11: highlighting this. we we stop editing the field in the DD format, we format it with 6 decimal places. not sure if that's too granular or if it's okay

src/frontend/screens/ManualGpsScreen/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+  },
+  universalTransverseMercator: {
+    id: "screens.ManualGpsScreen.universalTransverseMercator",
+    defaultMessage: "Universal Transverse Mercator (UTM)",
+  },
+});
+
+const usePersistedState = createPersistedState("manualCoordinateEntryFormat");
+
+const ManualGpsScreen: StackScreenComponent = ({ navigation }) => {
```
achou11: the `StackScreenComponent` annotation here is very important for type-checking and inference. helps with props but also the `navigationOptions` configuration on the component level

src/frontend/sharedTypes.ts
```diff
@@ -0,0 +1,41 @@
+// TS port of /src/frontend/types.js
```
achou11, line 1: open to renaming this or doing more of a directory-style type of structure with this. used this name to differentiate it from `types.js`, which is still needed as long as Flow is used in the repo

src/frontend/screens/ManualGpsScreen/UtmForm.tsx
```diff
@@ -38,98 +32,86 @@ const m = defineMessages({
...
-        navigation.pop();
-      } catch (err) {
-        ToastAndroid.showWithGravity(
-          err.message,
-          ToastAndroid.LONG,
-          ToastAndroid.TOP
-        );
+        const { easting } = fromLatLon(coords.lat, coords.lon);
+        // TODO: Is it okay to round this to 3 decimal places?
+        return easting.toFixed(3);
```
achou11: highlighting this. figure that should be precise enough for UTM but don't have the most context for it

src/frontend/screens/ManualGpsScreen/UtmForm.tsx
```diff
@@ -38,98 +32,86 @@ const m = defineMessages({
...
+    } else {
+      return "";
     }
+  });
+  const [northing, setNorthing] = React.useState(() => {
+    if (typeof coords?.lat === "number" && typeof coords?.lon === "number") {
+      try {
+        const { northing } = fromLatLon(coords.lat, coords.lon);
+        // TODO: Is it okay to round this to 3 decimal places?
+        return northing.toFixed(3);
```
achou11: same question about precision here as [this comment](https://github.com/digidem/mapeo-mobile/pull/633/files#r671519224)

src/frontend/sharedTypes.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export type NavigationProp = NavigationStackProp<
+  NavigationRoute,
+  NavigationParams
+>;
+
+export type StackScreenComponent = NavigationStackScreenComponent<
+  NavigationParams
+>;
```
ErikSin, line 41: It does seem like all the shared info is shared through our context providers, so I would say for simplicity we can just name the navigation params directly in the component. But that is a personal preference, and I think we can do it both ways unless you have any strong opposition to that!

achou11: yeah it's probably sufficient to define this per component. this just based on whatever was defined in `types.js`

src/frontend/lib/utils.js
```diff
@@ -122,22 +122,43 @@ export function getLastPhotoAttachment(
...
   const degrees = Math.floor(absolute);
   const minutesNotTruncated = (absolute - degrees) * 60;
   const minutes = Math.floor(minutesNotTruncated);
   const seconds = (minutesNotTruncated - minutes) * 60;
-  return `${degrees}° ${minutes}' ${seconds.toFixed(3)}"`;
+  return {
+    raw: {
+      degrees,
+      minutes,
+      seconds: Number.parseFloat(seconds.toFixed(3)),
```
gmaclennan: I think maybe don't round seconds here, only when formatting to a string.

src/frontend/lib/utils.js
```diff
@@ -122,22 +122,43 @@ export function getLastPhotoAttachment(
...
+  degrees,
+  minutes,
+  seconds,
+}: {|
+  degrees: number,
+  minutes: number,
+  seconds: number,
+|}) {
+  return Number.parseFloat(
+    (degrees + minutes / 60 + seconds / 3600).toFixed(6)
```
gmaclennan: as above, let's only `toFixed()` when (a) formatting to a string for user display or (b) parsing user input, just to avoid gradual loss of precision through conversions

src/frontend/lib/utils.js
```diff
@@ -122,22 +122,43 @@ export function getLastPhotoAttachment(
...
   const minutes = Math.floor(minutesNotTruncated);
   const seconds = (minutesNotTruncated - minutes) * 60;
-  return `${degrees}° ${minutes}' ${seconds.toFixed(3)}"`;
+  return {
+    raw: {
+      degrees,
+      minutes,
+      seconds: Number.parseFloat(seconds.toFixed(3)),
+    },
+    formatted: `${degrees}° ${minutes}' ${seconds.toFixed(3)}"`,
```
gmaclennan: nitpick: separate function `formatDMS()` e.g. `formatDMS(toDegreesMinutesSeconds(coord))`. But it's just a personal preference.

src/frontend/screens/ManualGpsScreen/DdForm.tsx
gmaclennan: 6 decimal places is good accuracy I think. However not sure about the behaviour of formatting to fewer decimal places when changing field, this might be confusing for the user e.g. they move to the next field and digits disappear from the number they just entered. I think maybe skip this `onEndEditing`? Although you might be thinking of a scenario that I'm not considering here.

src/frontend/screens/ManualGpsScreen/UtmForm.tsx
gmaclennan: For UTM ok to round to 1 decimal place (0.1m), but I think we should only round when we save, not change the users input.

src/frontend/sharedTypes.ts
gmaclennan: I _think_ these navigation params hold data that is not in context providers - this is the part of the state that is held in the navigation. However it's been a while since I looked at this. It is in prep for deep-linking support, e.g. a link that opens a screen in the app directly for a particular observation or photo, should we need that in the future.

gmaclennan (commented):
Looking good @achou11. These components are a great candidate for unit testing. Now that we have Jest set up, do you think you could add some unit tests for these? It would be good to simulate different user inputs and check that we get out what we expect - that is easier than reading the code and doing mental calculations to check the logic I think.

src/frontend/screens/ManualGpsScreen/DdForm.tsx
achou11: > behaviour of formatting to fewer decimal places when changing field

I actually enforce a maximum length for text inputs here (11 characters, which includes a potential decimal), so the behavior you'd more likely see is that we'd _add_ decimal places when changing fields e.g. inputting `1.50` gets formatted to `1.500000`, which is still a little weird so i may just omit the `toFixed` anyways

tangentially, is it better to remove that character length limit as a whole? maybe it conflicts with real-world use cases

src/frontend/sharedTypes.ts
achou11: i won't make big adjustments to this in that case. can always follow up if needed

src/frontend/screens/ManualGpsScreen/shared.ts
```diff
@@ -0,0 +1,36 @@
...
+type Coordinates = { lat?: number | null; lon?: number | null };
+
+export type ConvertedCoordinateData = {
+  coords?: Coordinates;
+  error?: Error;
+};
+
+export type FormProps = {
+  coords?: Coordinates;
+  onValueUpdate: (convertedCoordinates: ConvertedCoordinateData) => void;
```
ErikSin, line 11: @achou11 Im unsure as to what you are doing here with `onValueUpdate` Prop

achou11: this is used as a format-agnostic state updater for the `ManualGpsScreen` that's used by any of the forms within it (`utm`, `dd`, `dms`). The idea is that we keep a top-level state for the end coordinates that we want to eventually save (lat + lon) but we need to be able to update that state in any of the forms. The forms themselves handle their own coordinate state based on the specifics of their format (including string formatting and whatnot), and then whenever valid input changes occur, we call this callback so that the global draft state is updated with the converted coordinates (e.g. UTM form will have UTM-specific input changes occur, then if the values are valid, we convert them to lat+lon and provide that in this callback)

gmaclennan (changes requested):
This looks really good @achou11. I've been testing it out on a real device, and there is one change of behaviour from the previous UTM-only version of this screen which I think might be a bit of a usability issue: Previously the behaviour was to use the last-known-location to set the zone number and letter, but the coordinates were left blank. The new behaviour is to use the last-known-coordinates to pre-fill the form fields. I think this introduces more possibility of error because:

1. Deleting data in form fields is prone to error, e.g. the user might click the cursor to the middle of the existing value and not be able to delete all of it, or they might not delete everything by mistake
2. They might forget to change / input one of the fields.
3. For the UTM screen, at least, the initial value shows with 9 decimal places of precision, which means that the actual value is hidden (it overflows the form field) and the same might happen on smaller-screened phones with the other entry methods.

I think it makes sense to use the last-known-location to set the cardinality (or the Zone in UTM), because (A) it avoids needing to add an "unselected" state to those pickers, and (B) in most cases (apart from crossing the equator since last known location or moving between UTM zones) it will be correct, so if the user forgets to change it it will be ok in most circumstances.

I know this perhaps requires some changes to the architecture (passing down additional props) but I think it's best if cardinality & UTM zone is pre-filled based on last-known-location and that the coordinate fields start off empty. Open to thoughts / push back though!

achou11:
@gmaclennan that all makes sense! 

> I know this perhaps requires some changes to the architecture (passing down additional props) but I think it's best if cardinality & UTM zone is pre-filled based on last-known-location and that the coordinate fields start off empty

i don't think it'll be too much revisioning but I agree that if we are going to pre-fill fields, it should just meta fields and not numeric ones, as you've suggested 👍 One clarification: do we want to use the last known location or the last saved draft value for determining those pre-filled meta fields? Assuming it's the former based on your wording

gmaclennan (changes requested):
The labels for the form fields disappear when I focus them and the keyboard appears. This appears to be caused by a white area above the keyboard which is covering them, see the video below:


https://user-images.githubusercontent.com/290457/127694926-6c3dd307-3245-4d89-bec2-37c4819216ff.mp4

achou11:
> The labels for the form fields disappear when I focus them and the keyboard appears. This appears to be caused by a white area above the keyboard which is covering them, see the video below:

@gmaclennan is that testing the app with [this commit](https://github.com/digidem/mapeo-mobile/pull/633/commits/f4c8e6e3a33d7a8db7f0532666401c082e153f0e)? tried to address that there so just curious if it's still an issue (or as much of an issue, don't think the commit fully remedies that)

gmaclennan:
> One clarification: do we want to use the last known location or the last saved draft value for determining those pre-filled meta fields? Assuming it's the former based on your wording

I _think_ the last-known-location (e.g. `location.savedPosition.coords`). This updates every time the location updates. The draft value will be the actual location, which will be the same as `location.savedPosition` _except_ when the GPS on the phone is turned off, when the draft.value.coords will be undefined, but `location.savedPosition` will have the last location the app knew about (including across app restarts).

gmaclennan:
> @gmaclennan is that testing the app with [this commit](https://github.com/digidem/mapeo-mobile/pull/633/commits/f4c8e6e3a33d7a8db7f0532666401c082e153f0e)? tried to address that there so just curious if it's still an issue (or as much of an issue, don't think the commit fully remedies that)

Yes, testing with that commit (but before your rebase)

gmaclennan:
I don't feel I've got my head into this code enough to properly review the unit tests. Happy to go with your judgement on this, or if @ErikSin wants to take a look?

For this type of thing I've often follow a pattern of "given this user input, this is the expected result", and I often try to write these as an array / JSON and then write a test that iterates over that and tries all the inputs and checks the outputs. Not sure if it is the right approach here. It can be helpful when a user comes back with a bug like "I tried to enter this value and it did not work", and we can replicate this in the test. I think the main thing these unit tests want to check is that the parsing and formatting works correctly, and that the form fields correctly handle the user entering invalid values, such as text.

ErikSin:
@gmaclennan I can take a look at this!

achou11:
> I don't feel I've got my head into this code enough to properly review the unit tests. Happy to go with your judgement on this, or if @ErikSin wants to take a look?
> 
> For this type of thing I've often follow a pattern of "given this user input, this is the expected result", and I often try to write these as an array / JSON and then write a test that iterates over that and tries all the inputs and checks the outputs. Not sure if it is the right approach here. It can be helpful when a user comes back with a bug like "I tried to enter this value and it did not work", and we can replicate this in the test. I think the main thing these unit tests want to check is that the parsing and formatting works correctly, and that the form fields correctly handle the user entering invalid values, such as text.

Yeah taking a step back I think the tests I have are pretty implementation-specific, which is probably not ideal for this kind of testing. Think they still serve some value because they are somewhat addressing and emulating potential user actions, but certainly can be more expressive and comprehensive (e.g. testing that only numeric values can be written for certain inputs).

I think the parsing/formatting/conversion stuff is a little harder to test but that may be because I'm a little unfamiliar about the best way to test that or write the tests for them. Definitely open to adding / revising the tests to better fit the goals of having them in the first place 😄

achou11:
@gmaclennan I _think_ i've fixed the issue related to the keyboard via https://github.com/digidem/mapeo-mobile/pull/633/commits/29a9a649f1f52cf267475c433d439f6b87b15aa1 but would be helpful to get some external confirmation 😄

ErikSin:
@achou11 I will look at it today!

PR #634, opened 2021-07-10
Create tsconfig.json

ErikSin:
Added a TSCONFIG file to the frontend

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,7 @@
 {
-  presets: ["module:metro-react-native-babel-preset"]
+  "presets": [["module:metro-react-native-babel-preset"],
+    ["@babel/preset-typescript",{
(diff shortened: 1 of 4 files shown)


achou11 (approved):

gmaclennan:
React Native does not require the babel transform, and it breaks things. Also I think we should use the `tsconfig.json` from the [react-native template](https://github.com/react-native-community/react-native-template-typescript/blob/master/template/tsconfig.json), the key differences being:

- `noEmit: true`: there is no emit step for a React Native project since Metro does the bundling
- `jsx: react-native`: Not sure what this does, but it's in the template.
- `lib: ["es2017"]`: this is the version of JS supported (and used by react-native modules)

Obviously keeping the `include` option from this PR.

PR #635, opened 2021-07-14
add tsconfig

ErikSin:
Added updated tsconfig, took out babel configuration for tyspecript (babel-react-native presets automatically do it)

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "@storybook/react-native": "^5.3.19",
     "@storybook/react-native-server": "^5.3.19",
     "@turf/random": "^6.0.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -121,6 +121,7 @@
     "@storybook/react-native": "^5.3.19",
     "@storybook/react-native-server": "^5.3.19",
     "@turf/random": "^6.0.2",
+    "@types/react-native": "^0.64.12",
```
achou11: wondering if we should keep this at `0.63.X` just to align with our current version of react native, in case there are any significant differences between the type defs

ErikSin: Sure! Ill change it now!

Issue #623, opened 2021-07-08
(Front) Project Joiner - Join/Create project screen


ximenabb:
**Description :**
Project Joiner must Join or Create a project before landing in the app

**Screen**
[Here](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=45808%3A3&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1)

*Points*
3pts

PR #637, opened 2021-07-14
feature: join project screen with qr code

### LINKED ISSUE #624: (Front) Project Joiner - QR Code screen
**Description :**
Project Joiner clicked on Join Project and the QR Code Screen appears

**Screen is:**
-  'Join a project' buttonto go Back to the main screen
-  'Send join request instead' button
- A QR code display

**Screen**
[Here](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=45808%3A52&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1)

**Points**
13pts

achou11:
Closes #624 

Notes:
- adds new deps [react-native-qrcode-svg](https://github.com/awesomejerry/react-native-qrcode-svg) and [react-native-svg](https://github.com/react-native-svg/react-native-svg)
- upgrades `@react-native-community/netinfo` to [`6.0.0`](https://github.com/react-native-netinfo/react-native-netinfo/releases/tag/v6.0.0), which only introduces a breaking change to a hook that we don't use
- creates a shared `WifiBar` component that was extracted from the `SyncModal`
- creates a `useWifiStatus` hook that extracts the wifi network detection logic from `SyncModal`. At the moment, only returns the ssid but can see it providing other network info in the future
- this screen is currently not used anywhere in the app (depends on some https://github.com/digidem/mapeo-mobile/issues/623 being addressed)
- **some open questions that need to be addressed before this can be merged:**
  - what details are we encoding in the QR code?

---

Preview:

image


src/frontend/screens/JoinProject.tsx
```diff
@@ -0,0 +1,125 @@
...
+const JoinProject = () => {
+  const { ssid } = useWifiStatus();
+
+  const deviceName: string =
+    "Android " + getUniqueId().slice(0, 4).toUpperCase();
+
+  return (
+    <View style={styles.pageContainer}>
+      {/* TODO: Probably want to do something smarter here if no ssid is detected */}
+      {ssid && <WifiBar deviceName={deviceName} ssid={ssid} />}
```
achou11: couple of questions here:

- do we want the wifi bar to be pressable here? e.g. to take them to system settings or something similar
- what should we show if no ssid is detected?

src/frontend/screens/JoinProject.tsx
```diff
@@ -0,0 +1,125 @@
...
+      <FormattedMessage {...m.title} />
+    </HeaderTitle>
+  ),
+  headerLeft: ({ onPress }: { onPress: () => void }) => (
+    <IconButton onPress={onPress}>
+      <BackIcon color={WHITE} />
+    </IconButton>
+  ),
+  headerStyle: {
+    backgroundColor: MEDIUM_BLUE,
```
achou11: this blue doesn't quite match the mockup but don't think we have a defined color for what was used there

src/frontend/screens/JoinProject.tsx
```diff
@@ -0,0 +1,125 @@
...
+    alignItems: "center",
+  },
+  sendJoinRequest: {
+    fontSize: 16,
+    fontWeight: "bold",
+    textDecorationLine: "underline",
+  },
+});
+
+export default JoinProject;
```
achou11: very open to renaming this component. just needed something that was clear enough to work with as i was building

src/frontend/sharedComponents/Button.tsx
```diff
@@ -54,6 +45,8 @@ const Button = ({
     >
       <TouchableNativeFeedback
         disabled={disabled}
+        // TODO: There's some typing issue with react-native-gesture-handler here. Using the definition from react native works fine
+        // @ts-expect-error
         background={TouchableNativeFeedback.Ripple(VERY_LIGHT_BLUE, false)}
```
achou11: think there's something off about `react-native-gesture-handler`'s typedef for the `TouchableNativeFeedback`. Linter gives me an error that the `Ripple` static method doesn't exist on the class but it's supposed to be there to mirror the React Native API

src/frontend/screens/JoinProject.tsx
```diff
@@ -0,0 +1,125 @@
...
+            </Text>
+            <Text style={styles.instructionsDescription}>
+              <FormattedMessage {...m.instructionsDescription} />
+            </Text>
+          </View>
+        </View>
+        <Button
+          variant="text"
+          onPress={() =>
+            Alert.alert("Work in progress", "Feature not implemented yet")
```
achou11: just a placeholder for now, since a different ticket will focus on implementing this

src/frontend/screens/JoinProject.tsx
```diff
@@ -0,0 +1,125 @@
...
+    "Android " + getUniqueId().slice(0, 4).toUpperCase();
+
+  return (
+    <View style={styles.pageContainer}>
+      {/* TODO: Probably want to do something smarter here if no ssid is detected */}
+      {ssid && <WifiBar deviceName={deviceName} ssid={ssid} />}
+      <View style={styles.container}>
+        <View>
+          <View style={styles.qrCodeContainer}>
+            <QRCode value="https://digital-democracy.org" size={250} />
```
achou11: Just a placeholder. Need to figure out what to encode here

src/frontend/screens/JoinProject.tsx
achou11: ah i think [this mockup](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=46062%3A501&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1) probably answers the second question.

src/frontend/screens/JoinProject.tsx
achou11: based on discussion, seems like we want to encode public key, IP address, and port. @gmaclennan any thoughts on what to use for encoding these (not familiar with best practices if any) and also, may need to figure out how to get these details in the first place

gmaclennan: > based on discussion, seems like we want to encode public key, IP address, and port. 

Yes, now I'm also thinking maybe device name and/or device ID so that we can show those on the admin side when they scan the join request?

> @gmaclennan any thoughts on what to use for encoding these (not familiar with best practices if any)

The easiest option is to encode this as a JSON e.g.

```json
{
   "publicKey": "dadceeb7c5d377dfdcc0b814acb336c0cf484a02ac5a87d8408e05eddd56d7dc",
   "host": "10.532.432.1:5998",
}
```

_however_ it depends on how much "space" we have in the QR code. The above JSON is 107 characters (it's a 256-bit key encoded in hex). If space is of essence we could encode this to binary: 107 characters is 107 bytes _if_ we use only UTF-8 characters. In binary we have a 256-bit key = 32 bytes, plus 4 bytes for the IP, and 2 bytes for the port, so if we encode it as a very simple concatenated buffer of publicKey | IP | port then that's only 38 bytes. But there is a lot less flexibility there, and more work if we want to add custom length fields. We could also use protobuf to encode it. Again, it's more overhead. If the JSON encoding doesn't result in that large a QR code then we could just use that. Simple and easy to understand.

> and also, may need to figure out how to get these details in the first place

This will rely on backend Mapeo Core methods that have not yet been written.

achou11: > Yes, now I'm also thinking maybe device name and/or device ID so that we can show those on the admin side when they scan the join request?

ah yes definitely need that too

thanks for explaining the options! my initial thought is to resort to the JSON format in that case

> This will rely on backend Mapeo Core methods that have not yet been written.

right ok then i won't consider that as blocking for this PR for the time being

src/frontend/screens/JoinProject.tsx
```diff
@@ -0,0 +1,125 @@
...
+          <Text style={styles.sendJoinRequest}>
+            <FormattedMessage {...m.sendJoinRequest} />
+          </Text>
+        </Button>
+      </View>
+    </View>
+  );
+};
+
+JoinProject.navigationOptions = () => ({
```
ErikSin: You can type your React component as a `NavigationStackScreenComponent` => `import {NavigationStackScreenComponent,} from "react-navigation-stack" ` if you want typesafety here.

achou11: yeah i can that. i do have a more robust typing that i've created for these stack screen components in [this PR](https://github.com/digidem/mapeo-mobile/pull/633/files#diff-ca3239535e66a9a29ff9649d1ba99896a78b92acfeda726078c45b03acdcc9baR39-R41) so could add that annotation as a follow-up

src/frontend/screens/JoinProject.tsx
gmaclennan: > do we want the wifi bar to be pressable here? e.g. to take them to system settings or something similar

On the sync screen, pressing the wifi bar opens the wifi settings. I think the same behaviour here makes sense.

A slight improvement on this behaviour that I think would be clearer for the user (to avoid an unexpected jump to the phone settings app, e.g. when the user tries to press the back button in the top navigation bar and accidentally presses the wifi bar by mistake): Open a dialog/alert that explains why you would open the settings.

(A) No Wifi

Alert title: "No Wifi"
Alert message: "Open the settings app to turn on Wifi"
Alert OK: "Open Settings"
Alert cancel: "Cancel"

(B) Wifi network

Alert title: "Wifi: %network_name%"
Alert message: "You need to be on the same Wifi network to connect to other phones. Open the settings app to change your Wifi network"
Alert OK: "Open Settings"
Alert cancel: "Cancel"

Maybe Sabella has an idea here? I don't think this (additional alert dialog) is a blocker for merging this though.

src/frontend/hooks/useWifiStatus.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    // Subscribe to NetInfo to know when the user connects/disconnects to wifi
+    subscriptions.push({
+      remove: NetInfo.addEventListener(handleConnectionChange),
+    });
+
+    if (keepAwake) {
+      // Keep the screen awake whilst on this screen
+      KeepAwake.activate();
+    }
```
gmaclennan: This makes sense to put in a separate useEffect, since this useEffect will re-run for reasons other than `keepAwake` changing value.

src/frontend/hooks/useWifiStatus.ts
```diff
@@ -0,0 +1,83 @@
...
+        }
+      }
+    };
+
+    if (setListening) {
+      setListening(true);
+    }
+
+    // Subscribe to NetInfo to know when the user connects/disconnects to wifi
+    subscriptions.push({
```
gmaclennan: Really not sure why I had support for multiple subscriptions here. There is only one in this hook (unless I'm missing something), so having this as an array vs. just `subscription = addEventListener()` is unnecessary, but not harmful.

src/frontend/hooks/useWifiStatus.ts
```diff
@@ -0,0 +1,83 @@
...
+          onConnectionChange();
+        }
+
+        if (setListening) {
+          setListening(true);
+        }
+      }
+    };
+
+    if (setListening) {
```
gmaclennan: This does not belong here, it can sit it its own useEffect in the sync screen.

src/frontend/screens/SyncModal/index.js
```diff
@@ -50,53 +48,14 @@ const SyncModal = ({ navigation }: Props) => {
...
-      // Unsubscribe all listeners
-      subscriptions.forEach(s => s.remove());
-      // Turn off keep screen awake
-      KeepAwake.deactivate();
-    };
-  }, [syncGetPeers]);
+  const { ssid } = useWifiStatus({
+    keepAwake: true,
+    // On connection change, ensure we have the latest peer info from backend.
+    onConnectionChange: syncGetPeers,
```
gmaclennan: This can be a useEffect in this component that re-runs when `ssid` changes, it simplifies the implementation of the useWifiStatus hook.

src/frontend/hooks/useWifiStatus.ts
```diff
@@ -0,0 +1,83 @@
...
+      try {
+        ssid = await NetworkInfo.getSSID();
+      } catch (e) {
+        bugsnag.notify(e);
+      } finally {
+        // Even if we don't get the SSID, we still want to show that a wifi
+        // network is connected.
+        setSsid(ssid);
+
+        if (onConnectionChange) {
```
gmaclennan: See comment below in the sync component. I think we can remove this from here.

src/frontend/screens/JoinProject.tsx
gmaclennan: > what should we show if no ssid is detected?

Same as the sync modal I think, there is a "NoWifiBar" component.

src/frontend/sharedComponents/WifiBar.tsx
```diff
@@ -0,0 +1,61 @@
+import * as React from "react";
+import { StyleSheet, TouchableNativeFeedback, View } from "react-native";
```
gmaclennan: Use export from `./Touchables` - it will make adding iOS support easier in the future (since it exports different things based on platform), and it uses the react-native-gesture-handler touchable, that has better performance, although I think the new "Pressable" in react-native might have solved this now.

gmaclennan (commented):
Ok looks good @achou11. A couple of small comments about the useWifi hook. Before this merges I'd like to get a sense of what size the QR code will be and how much data we might be able to put in that.

src/frontend/screens/JoinProject.tsx
achou11: > Same as the sync modal I think, there is a "NoWifiBar" component.

Yeah I added support for that in [this PR](https://github.com/digidem/mapeo-mobile/blob/17bd81be6dd934914848188d6b15b80f93327e3f/src/frontend/sharedComponents/WifiBar.tsx) but will port to here

src/frontend/hooks/useWifiStatus.ts
achou11: should it still be part of this hook or is it more specific to the Sync screen in general?

src/frontend/hooks/useWifiStatus.ts
achou11: will pose this as a comment on the commit that updates this hook, but is the listening state directly related to the presence of `ssid` or should they be treated as different things? We use `listen` as the first argument of `usePeers` but based on your comment and just from a quick look at the code, it seems like `usePeers` could use the presence of `ssid` instead of a separate `listen` variable. Possible im missing something apparent

in the `SyncModal/index.js` component, instead of:

```js
  const [listen, setListening] = React.useState(false);
  const { ssid } = useWifiStatus();
  const [peers, syncPeer, syncGetPeers] = usePeers(listen, deviceName);
  
  // defining useEffects that update `listen` based on changes in `ssid`
  ...
```

can we replace it with this?

```js
  // const [listen, setListening] = React.useState(false);
  const { ssid } = useWifiStatus();
  const [peers, syncPeer, syncGetPeers] = usePeers(!!ssid, deviceName);
```

src/frontend/screens/SyncModal/index.js
```diff
@@ -50,53 +49,20 @@ const SyncModal = ({ navigation }: Props) => {
     },
   ] = React.useContext(ConfigContext);
 
-  const [listen, setListening] = React.useState<boolean>(false);
-  const [peers, syncPeer, syncGetPeers] = usePeers(listen, deviceName);
-  const [ssid, setSsid] = React.useState<null | string>(null);
+  const { ssid } = useWifiStatus();
+  const [peers, syncPeer, syncGetPeers] = usePeers(!!ssid, deviceName);
```
achou11, line 54: See https://github.com/digidem/mapeo-mobile/pull/637#discussion_r675168229 for context but wondering if using the `ssid` here makes sense, as opposed to having a separate `listen` variable that gets updated based on changes to the `ssid` via an effect

src/frontend/screens/SyncModal/index.js
```diff
@@ -50,53 +49,20 @@ const SyncModal = ({ navigation }: Props) => {
...
-    };
-  }, [syncGetPeers]);
+    return KeepAwake.deactivate;
+  }, []);
+
+  React.useEffect(() => {
+    if (ssid) {
+      syncGetPeers();
+    }
+  }, [ssid, syncGetPeers]);
```
achou11, line 66: I'm assuming that we only want to call `syncGetPeers` when an `ssid` is present

src/frontend/hooks/useWifiStatus.ts
achou11: i decided to move this to the sync screen specifically

src/frontend/hooks/useWifiStatus.ts
gmaclennan: Yes, I can't see any reason why this would not work. Looks good to me. This sync code needs a cleanup at some point! It's too complicated.

src/frontend/screens/SyncModal/index.js
gmaclennan: I _think_ so. Is it possible that a device can be connected to wifi but `ssid` be `null` or `undefined`?

src/frontend/hooks/useWifiStatus.ts
gmaclennan: Yeah, this is there for the sync screen because sync can take a while and we want users to be able to leave it open and watch sync progress without the device going to sleep.

src/frontend/screens/SyncModal/index.js
achou11: just from a quick search i'd think it's highly unlikely that one can connect to a network that doesn't have an ssid, so feeling okay about this i guess 🤷

src/frontend/screens/SyncModal/index.js
achou11: addressed in https://github.com/digidem/mapeo-mobile/pull/637#discussion_r675639800

src/frontend/screens/SyncModal/index.js
```diff
@@ -50,53 +49,20 @@ const SyncModal = ({ navigation }: Props) => {
...
-    return () => {
-      // When the modal closes, stop announcing for sync
-      setListening(false);
-      // Unsubscribe all listeners
-      subscriptions.forEach(s => s.remove());
-      // Turn off keep screen awake
-      KeepAwake.deactivate();
-    };
-  }, [syncGetPeers]);
+    return KeepAwake.deactivate;
```
gmaclennan: Small nitpick, for the sake of a few extra characters it's always best to do `return () => KeepAwake.deactivate()` rather than this, for the same reason it's best not to pass a function directly to `myArray.map(library.method)`, because a non-breaking change could lead to an unexpected argument.

achou11: ah yes good point! will fix

src/frontend/hooks/useWifiStatus.ts
```diff
@@ -0,0 +1,41 @@
...
+        // Even if we don't get the SSID, we still want to show that a wifi
+        // network is connected.
+        setSsid(ssid);
+      }
+    };
+
+    // Subscribe to NetInfo to know when the user connects/disconnects to wifi
+    const unsubscribe = NetInfo.addEventListener(handleConnectionChange);
+
+    return unsubscribe;
```
ErikSin: Why are you calling this `unsubscribe`?

src/frontend/screens/JoinProject.tsx
```diff
@@ -0,0 +1,129 @@
...
+    id: "screens.JoinProject.sendJoinRequest",
+    defaultMessage: "Send Join Request instead",
+  },
+});
+
+const JoinProject = () => {
+  const { ssid } = useWifiStatus();
+
+  const deviceName: string =
+    "Android " + getUniqueId().slice(0, 4).toUpperCase();
```
ErikSin: What is the reasoning for only using 4 characters here?

src/frontend/screens/JoinProject.tsx
ErikSin: Lets rename it JoinProjectQR Screen? Just to differentiate it from the other join project splash screen

src/frontend/screens/JoinProject.tsx
```diff
@@ -0,0 +1,129 @@
...
+    id: "screens.JoinProject.instructionsDescription",
+    defaultMessage: "Show this QR code to your Project Admin",
+  },
+  sendJoinRequest: {
+    id: "screens.JoinProject.sendJoinRequest",
+    defaultMessage: "Send Join Request instead",
+  },
+});
+
+const JoinProject = () => {
```
ErikSin: You can type this as NavigationStackScreenComponent

src/frontend/screens/JoinProject.tsx
```diff
@@ -0,0 +1,129 @@
...
+  headerTitle: () => (
+    <HeaderTitle style={{ color: WHITE }}>
+      <FormattedMessage {...m.title} />
+    </HeaderTitle>
+  ),
+  headerLeft: ({ onPress }: { onPress: () => void }) => (
+    <IconButton onPress={onPress}>
+      <BackIcon color={WHITE} />
+    </IconButton>
+  ),
```
ErikSin: Once we add this to a nav stack, it should automatically do this.

src/frontend/sharedComponents/Button.tsx
achou11: upgrading to latest version of `react-native-gesture-handler` fixes this. upgrade is handled is #649 so will mark this as resolved

ErikSin (approved):
Great Job

src/frontend/hooks/useWifiStatus.ts
achou11: `NetInfo.addEventListener` returns an unsubscribe function ([docs](https://github.com/react-native-netinfo/react-native-netinfo#usage)). I'll be a little more explicit about the return here, which is just the `useEffect` cleanup function) i.e. `return () => unsubscribe()`

src/frontend/screens/JoinProject.tsx
achou11: not super sure since this is copied from elsewhere in the app 😅 probably length considerations if i had to guess

src/frontend/hooks/useWifiStatus.ts
ErikSin: Oh yeah, it was def the syntax that was confusing me. Now that you've pointed out that it is a cleanup functions it makes alot more sense!

PR #642, opened 2021-07-18
User Prompt to leave project before joining a new one

ErikSin:
### Figma Screen:
[Screen](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=45997%3A0&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1)

# Description
This screen indicates to user that they must leave their current project before they can join a new project.

## To Do
Buttons need to be hooked up to navigation when the leave project flow is merged into main


src/frontend/context/ConfigContext.tsx
```diff
@@ -13,67 +12,67 @@ import type { Status } from "../types";
...
+  fields: FieldsMap;
+  metadata: Metadata;
+  messages: Messages;
+  status: Status;
 };
 
 export type ConfigContextType = [
   State,
-  { reload: () => any, replace: (fileUri: string) => any }
+  { reload: () => any; replace: (fileUri: string) => any }
```
achou11, line 75: i'm assuming these don't actually return `any` and would be a good opportunity to change it to `void` (assuming it doesn't break a whole bunch of other stuff)

src/frontend/screens/AlreadyOnProject.tsx
```diff
@@ -0,0 +1,120 @@
...
+};
+
+export const AlreadyOnProj: NavigationStackScreenComponent = () => {
+  const [config] = useContext(ConfigContext);
+  const { formatMessage: t } = useIntl();
+  const nav = useNavigation();
+
+  const name = useMemo(() => {
+    return config.metadata.name;
+  }, [config]);
```
achou11, line 55: curious if `useMemo` is useful/necessary here?

src/frontend/AppStack.js
```diff
@@ -84,9 +85,10 @@ const AppStack = createStackNavigator(
     ObservationEdit: ObservationEdit,
     ManualGpsScreen: ManualGpsScreen,
     ObservationDetails: ObservationDetails,
+    AlreadyOnProj: AlreadyOnProj,
   },
   {
-    initialRouteName: "Home",
+    initialRouteName: "AlreadyOnProj",
```
achou11, line 91: guessing this was changed for debugging purposes?

achou11 (approved):
Might re-review after the rebase but looks good otherwise. There are a couple of small non-blocking comments too

achou11:
Closing since these changes were already merged via #715

PR #643, opened 2021-07-19
feature: Add Send Join Request screen

### LINKED ISSUE #627: (Front) Project Joiner - Verification Code screen
**Description** :
Project Joiner clicks on 'Join and Sync', has no wifi so chooses to send a Verification Code instead

**Screen is:**
- Display of the verification code
- CTA redirecting to the Signal app
- Share Sheet + Deep linking

**Screen**
[Here
](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=46062%3A501&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1)

**Technical Documentation**

**Points**
13pts

achou11:
Closes #627 

Notes:
- Opening this PR as an extension of #637 to benefit from some changes in that branch.
- Remaining questions:
  - how should we generate the verification code? Is that something that would be handled on the backend?
  
---

Preview:

- Initial:

image

- Share sheet opened:

image

- Flow:

send-join-request-flow

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,6 @@
     "@types/react-native": "^0.63.2",
     "@types/react-native-vector-icons": "^6.4.8",
     "@types/react-native-keep-awake": "^2.0.3",
-    "@types/react-navigation": "^3.4.0",
     "@types/react-test-renderer": "^17.0.1",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^25.3.0",
--- a/src/frontend/screens/SendJoinRequestScreen.tsx
+++ b/src/frontend/screens/SendJoinRequestScreen.tsx
@@ -1,0 +1,124 @@
+import * as React from "react";
+import { Share, StyleSheet, View } from "react-native";
+import { FormattedMessage, defineMessages } from "react-intl";
+import { getUniqueId } from "react-native-device-info";
+import {
+  NavigationStackScreenComponent,
+  TransitionPresets,
+} from "react-navigation-stack";
+
+import useWifiStatus from "../hooks/useWifiStatus";
+import { MEDIUM_BLUE, WHITE } from "../lib/styles";
+import HeaderTitle from "../sharedComponents/HeaderTitle";
+import { BackIcon } from "../sharedComponents/icons";
+import Button from "../sharedComponents/Button";
+import IconButton from "../sharedComponents/IconButton";
+import WifiBar from "../sharedComponents/WifiBar";
+import Text from "../sharedComponents/Text";
+
+const m = defineMessages({
+  title: {
+    id: "screens.SendJoinRequest.title",
+    defaultMessage: "Send Join Request",
+  },
+  verificationCode: {
+    id: "screens.SendJoinRequest.verificationCode",
+    defaultMessage: "Verification Code",
+  },
+  instructions: {
+    id: "screens.SendJoinRequest.instructions",
+    defaultMessage: "Share this code with the Project Coordinator",
+  },
+  share: {
+    id: "screens.SendJoinRequest.share",
+    defaultMessage: "Share",
+  },
+});
+
+const SendJoinRequestScreen: NavigationStackScreenComponent = () => {
+  const { ssid } = useWifiStatus();
+
+  const deviceName = "Android " + getUniqueId().slice(0, 4).toUpperCase();
+
+  // TOOD: Need to properly generate
+  const generateVerificationCode = () => "12345";
+
+  // TODO: Need to properly generate a deeplink
+  const shareLink = "https://digital-democracy.org";
+
+  return (
+    <View style={styles.pageContainer}>
+      <WifiBar deviceName={deviceName} ssid={ssid} />
+      <View style={styles.container}>
+        <View>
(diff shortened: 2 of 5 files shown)


src/frontend/screens/JoinProject.tsx
```diff
@@ -62,9 +63,7 @@ const JoinProject = () => {
         </View>
         <Button
           variant="text"
-          onPress={() =>
-            Alert.alert("Work in progress", "Feature not implemented yet")
-          }
+          onPress={() => navigation.navigate("SendJoinRequest")}
```
achou11: TODO: register this screen

src/frontend/screens/SendJoinRequest.tsx
```diff
@@ -0,0 +1,120 @@
...
+  },
+});
+
+const SendJoinRequest = () => {
+  const { ssid } = useWifiStatus();
+
+  const deviceName = "Android " + getUniqueId().slice(0, 4).toUpperCase();
+
+  // TOOD: Need to properly generate
+  const generateVerificationCode = () => "12345";
```
ErikSin: Given that none of this flow relies on the backend, I would just generate the code here, hash it, and then send it with the link. Then it will need to be unhashed on the other side

ErikSin: So i was looking into this, we could use [Bcrypt](https://github.com/dcodeIO/bcrypt.js#readme).

We can salt and hash the verfication code and then just add the hashed code to the end of the deep link. Then we can use BCrypt to compare the hashed password with the code that the admin types in. BCrypt only does comparisons. AKA it will only return true or false if the hashed password === the typed in password, but there is no function to actually unhash the password

achou11: @gmaclennan any thoughts here? Two things we need to think about:

1. how to generate the verification code. is this something that will be handled client-side or on the backend? if client-side, was thinking about doing something super naïve for now like using `Math.random` and trimming the length 😅 

2. how to encrypt the verification code for sharing (as Erik's comment addresses)

achou11: renamed this file so reference is here: https://github.com/digidem/mapeo-mobile/pull/643/files#diff-d7f3d746b36db064fae37a46b05b4372f403633c29e705c7faa8e026ea68c6ebR43-R44

package.json
```diff
@@ -134,7 +134,6 @@
     "@types/react-native": "^0.63.2",
     "@types/react-native-vector-icons": "^6.4.8",
     "@types/react-native-keep-awake": "^2.0.3",
-    "@types/react-navigation": "^3.4.0",
```
achou11, line 137: `react-navigation` comes with its own types

src/frontend/screens/SendJoinRequestScreen.tsx
```diff
@@ -0,0 +1,124 @@
...
+            <FormattedMessage {...m.share} />
+          </Text>
+        </Button>
+      </View>
+    </View>
+  );
+};
+
+SendJoinRequestScreen.navigationOptions = () => ({
+  ...TransitionPresets.SlideFromRightIOS,
```
achou11, line 74: Felt more natural to have the screen come in from the side and this is how you configure that apparently (without fiddling around with lower-level animation configs)

src/frontend/screens/SendJoinRequest.tsx
gmaclennan: This action (sending a join request) will be to send the device public key (a 256-bit random number). The purpose of the verification step is to ensure that the public key is from who we think it is, to avoid an attack of the kind "hey, I'm Andrew, please let me join your project", when I'm not Andrew :)

The validation I was thinking was a side-channel communication to check with the project joiner that the invite has indeed come from them. The direct way to do this is to compare the public key:

Project Admin: "Hey, I've just received a join request from you, with the public key [reads 256-bit number], is that you?"
Project Joiner: "Oh hey, [checked 256-bit number against public key on their screen], oh yeah, that matches, that is me."

Now the Project Admin knows the invite is definitely from the person they think it is.

However, reading out a 256-bit number is not a great experience! The idea with the verification code is to hash the public key to get a human readable code that could be compared (a [fingerprint](https://en.wikipedia.org/wiki/Public_key_fingerprint)). Signal has something similar called ["Safety numbers"](https://signal.org/blog/safety-number-updates/) which are a combination of two 30-digit fingerprints of each users public key.

The goal is to have a shorter number derived from the public key that the user can read out. I'm not sure what method Signal uses to derive that 30-digit number, or how they decided what was enough entropy, but it's still a bit of a clunky step. It's not necessary in Mapeo when we are scanning a QR code in-person, because the person is there present with us. It would only be necessary for join requests that are sent via a message.

I think the attack area here is quite small, and I'm not sure adding a 30-digit verification code is ever going to be actually used by any user. Most likely a check of "did you just send me a join request?" "yes/no" would be enough - an attacked would have to be very clever to intercept that and fake it with their own, and also intercept the reply.

In summary: I say remove this verification code step from the flow.

PR #649, opened 2021-07-23
feature: project invite modal

### LINKED ISSUE #625: (Front) Project Joiner - Join Project Invite
**Description :**
Project Joiner is anywhere in the app and receives the Join Project invite

**Screen is:**
- Bottom nav slider with the option to Decline or Join project
- if appears during an observation creation of needs to be queued 
- Display: 
   - project name
   - project ID
   - Role

**Screen**
[Here
](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=45820%3A59&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1)

**Points**
34pts

achou11:
Closes #625 

Notes:
- Adds new dependency: [react-native-bottom-sheet](https://github.com/gorhom/react-native-bottom-sheet/) and upgrades `react-native-gesture-handler` to `^1.10.3`
- The touchables used in the modal are slightly different than what's used as the default in our button component due to compat issues with RNGH's touchables
- General architecture is set up as such:
  - `ProjectInviteContext`: holds state + handlers for dealing with invites. this sits **outside** the app container so that it's globally available to all screens
  - `useProjectInviteListener`: basic hook to react to changes in the invites state that comes from the `ProjectInviteContext`. Its purpose is to be used in specific screens with access to the `navigation` prop so that navigation events can be triggered e.g. a new invite has been received while in a screen within the map view, let's navigate to the `ProjectInviteModal` screen. In this PR, we use this in the following places:
    - bottom tab navigator for the main app, since that's persisted throughout the whole app, which is a good candidate for listening to this kind of change
    - join project via qr code and send join request screens 
  - `ProjectInviteModal`: standalone modal component that sits at the top stack navigator alongside the "app" stack, displaying relevant invite information and actions
- Couple of open implementation questions that should be considered before merging:
  - Still not sure about how to best incorporate disabling the modal based on the current screen. For the `AppStack`, the listener is embedded in the bottom tab navigator which I don't think has access to the current route outside of it?
  - do we want to support the idea of having a queue of invites? I've modeled it as such but seems more likely that we'll only ever handle invites one at a time
  
---

Preview:

- Static:

image

- Interactions: 

project-invite-modal

### DIFF
--- a/src/frontend/context/ProjectInviteContext.tsx
+++ b/src/frontend/context/ProjectInviteContext.tsx
@@ -1,0 +1,42 @@
+import * as React from "react";
+
+export interface ProjectInvite {
+  project: {
+    id: string;
+    name: string;
+  };
+  role: string;
+}
+
+const ProjectInviteContext = React.createContext<{
+  invite?: ProjectInvite;
+  removeInvite: (inviteId: string) => void;
+}>({ removeInvite: () => null });
+
+export const ProjectInviteProvider = ({
+  children,
+}: React.PropsWithChildren<{}>) => {
+  const [invites, setInvites] = React.useState<ProjectInvite[]>([]);
+
+  // TODO: Need to integrate backend/interactions that add invite
+
+  // TODO: Need to incorporate disabling based on current route (needs coordination with useProjectInviteListener)
+
+  const removeInvite = (projectId: string) =>
+    setInvites(previous =>
+      previous.filter(invite => invite.project.id !== projectId)
+    );
+
+  return (
+    <ProjectInviteContext.Provider
+      value={{
+        invite: invites[0],
+        removeInvite,
+      }}
+    >
+      {children}
+    </ProjectInviteContext.Provider>
+  );
+};
+
+export default ProjectInviteContext;
--- a/src/frontend/hooks/useProjectInviteListener.ts
+++ b/src/frontend/hooks/useProjectInviteListener.ts
@@ -1,0 +1,19 @@
+import * as React from "react";
+import {
+  NavigationParams,
+  NavigationScreenProp,
+  NavigationState,
+} from "react-navigation";
+import ProjectInviteContext from "../context/ProjectInviteContext";
+
+export default function useProjectInviteListener<
+  N extends NavigationScreenProp<NavigationState, NavigationParams>
+>(navigation: N) {
+  const { invite } = React.useContext(ProjectInviteContext);
+
+  React.useEffect(() => {
+    if (invite) {
+      navigation.navigate("ProjectInviteModal");
+    }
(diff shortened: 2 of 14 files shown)


src/frontend/context/ProjectInviteContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+    id: string;
+    name: string;
+  };
+  role: string;
+}
+
+const ProjectInviteContext = React.createContext<{
+  invite?: ProjectInvite;
+  removeInvite: (inviteId: string) => void;
+}>({ removeInvite: () => null });
```
ErikSin, line 14: What is this for?

src/frontend/hooks/useProjectInviteListener.ts
```diff
@@ -0,0 +1,19 @@
...
+  NavigationParams,
+  NavigationScreenProp,
+  NavigationState,
+} from "react-navigation";
+import ProjectInviteContext from "../context/ProjectInviteContext";
+
+export default function useProjectInviteListener<
+  N extends NavigationScreenProp<NavigationState, NavigationParams>
+>(navigation: N) {
+  const { invite } = React.useContext(ProjectInviteContext);
```
ErikSin: Is there a reason you dont just use the useNavigation Hook in this hook?

src/frontend/context/ProjectInviteContext.tsx
achou11: are you talking about the initialization or the actual `removeInvite` function? 

if the former, you need to initialize the context with a valid value. decided that a no-op instead of it just not existing would be a better way to go.

if the latter, it's so that we can process the invite after some action occurs (such as a declining the invite)

src/frontend/hooks/useProjectInviteListener.ts
achou11: nah could probably just use the hook. thought this [note](https://reactnavigation.org/docs/typescript/#annotating-usenavigation) about typing the hook was interesting though (so could be bc of that but not enough of a reason really)

src/frontend/context/ProjectInviteContext.tsx
ErikSin: I honestly don't remember this comment, but it looks fine now that I'm looking at it again.

src/frontend/hooks/useProjectInviteListener.ts
ErikSin: Cool. That was just out of curiosity!

ErikSin (approved):

ErikSin (approved):

PR #651, opened 2021-07-24
Create or Join project screen

ErikSin:
# Initial Splash Screen when user is asked to Join or Create Project

### Figma Mock Up
[Screen](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=45808%3A3&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1)

### To Keep note of
Switch navigator was added, with a new navigation stack. Read more about switch navigator [here](https://reactnavigation.org/docs/4.x/switch-navigator/)

### To do:
* We need to check whether user is already part of a project and navigate accordingly.
* Merged with other screens and flow


src/frontend/NavigationStacks/AppStack.tsx
```diff
@@ -72,8 +71,6 @@ const AppStack = createStackNavigator(
     // $FlowFixMe
     CoordinateFormat,
     // $FlowFixMe
-    Experiments,
-    // $FlowFixMe
     PhotosModal: PhotosModal,
     // $FlowFixMe
```
achou11: think you can get rid of the `$FlowFixMe` for this since it's a TS file now

src/frontend/context/ConfigContext.tsx
```diff
@@ -13,67 +12,64 @@ import type { Status } from "../types";
...
+  fields: FieldsMap;
+  metadata: Metadata;
+  messages: Messages;
+  status: Status;
 };
 
 export type ConfigContextType = [
   State,
-  { reload: () => any, replace: (fileUri: string) => any }
+  { reload: () => any; replace: (fileUri: string) => any }
```
achou11: I'd probably replace the `any` returns with `void`

src/frontend/sharedComponents/Button.tsx
```diff
@@ -1,51 +1,39 @@
...
+  color?: ColorScheme;
+  disabled?: boolean;
+  fullWidth?: boolean;
+  onPress: (event: GestureResponderEvent) => void;
+  style?: ViewStyleProp;
+  testID?: string;
+  variant?: Variant;
+}
 
 const Button = ({
```
achou11, line 38: Some changes to this component were merged into `develop` recently so make sure to update this to reflect that. Think you could also copy one of the implementations from my PRs ([this](https://github.com/digidem/mapeo-mobile/blob/b9bfe59b8b5ff74270e440ae3ea8e04d4183e054/src/frontend/sharedComponents/Button.tsx) or [this](https://github.com/digidem/mapeo-mobile/blob/7c64d6105f86c00c499e9708394c9b866997b637/src/frontend/sharedComponents/Button.tsx))

achou11 (approved):
looks good! left a couple of comments that i think you've already addressed

PR #652, opened 2021-07-24
Leave Project Flow

ErikSin:
# UI for Leaving a Project
The majority of this PR is just the UI as the backend functionality has not yet been created.

### Figma Mock Up
[Screens](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=45951%3A20&scaling=min-zoom)

### To Do
*Removing a Project from phone backend functionality


android/settings.gradle
```diff
@@ -1,4 +1,6 @@
 rootProject.name = 'mapeo'
+include ':@react-native-community_checkbox'
+project(':@react-native-community_checkbox').projectDir = new File(rootProject.projectDir, '../node_modules/@react-native-community/checkbox/android')
```
achou11, line 3: is this necessary? i think react native autolinking should handle this automatically ([link](https://github.com/react-native-checkbox/react-native-checkbox#mostly-automatic-installation))

ios/Podfile
```diff
@@ -8,6 +8,8 @@ target 'mapeo' do
 
   use_react_native!(:path => config["reactNativePath"])
 
+  pod 'RNCCheckbox', :path => '../node_modules/@react-native-community/checkbox'
```
achou11, line 11: same comment about autolinking taking care of this automatically. don't think this is needed

src/frontend/AppStack.tsx
```diff
@@ -24,6 +23,9 @@ import LanguageSettings from "./screens/Settings/LanguageSettings";
 import CoordinateFormat from "./screens/Settings/CoordinateFormat";
 import Experiments from "./screens/Settings/Experiments";
 import HomeHeader from "./sharedComponents/HomeHeader";
+import { LeaveProjectScreen } from "./screens/LeaveProject/LeaveProject";
+import { LeaveProjectProgress } from "./screens/LeaveProject/LeaveProjectProgess";
+import { LeaveProjectCompleted } from "./screens/LeaveProject/LeaveProjectCompleted";
```
achou11, line 28: more of an aside but i'm all for named exports, though we should stick to a convention (e.g. named exports or default exports for React components)

src/frontend/AppStack.tsx
```diff
@@ -87,9 +89,12 @@ const AppStack = createStackNavigator(
     ObservationEdit: ObservationEdit,
     ManualGpsScreen: ManualGpsScreen,
     ObservationDetails: ObservationDetails,
+    LeaveProjectScreen: LeaveProjectScreen,
+    LeaveProjectProgress: LeaveProjectProgress,
+    LeaveProjectCompleted: LeaveProjectCompleted,
   },
   {
-    initialRouteName: "Home",
+    initialRouteName: "LeaveProjectScreen",
```
achou11, line 97: remove this change since it was for debugging purposes?

src/frontend/screens/LeaveProject/LeaveProject.tsx
```diff
@@ -0,0 +1,200 @@
...
+  headerTitle: () => (
+    <HeaderTitle>
+      <FormattedMessage {...m.headerTitle} />
+    </HeaderTitle>
+  ),
+};
+
+export const LeaveProjectScreen: NavigationStackScreenComponent = () => {
+  const [confirmDelete, setConfirmDelete] = React.useState(false);
+  const [pristine, setPristine] = useState(true);
```
achou11, line 65: this feels like an odd variable name. not really sure what it means in this context

src/frontend/screens/LeaveProject/LeaveProject.tsx
```diff
@@ -0,0 +1,200 @@
...
+        <Text style={{ padding: 5, flex: 1 }}>
+          <FormattedMessage {...m.syncWarning} />
+        </Text>
+      </View>
+
+      <View style={styles.confirmContainer}>
+        <View style={styles.checkContainer}>
+          <CheckBox
+            value={confirmDelete}
+            onValueChange={() => setConfirmDelete(!confirmDelete)}
```
achou11, line 111: nitpick but i generally lean towards the [functional update](https://reactjs.org/docs/hooks-reference.html#functional-updates) syntax when it comes to toggling the value i.e.

```js
setConfirmDelete(previous => !previous)
```

android/settings.gradle
ErikSin: So I didnt add it, it was added though the autolinking command

achou11: ah gotcha nvm then

src/frontend/screens/LeaveProject/LeaveProject.tsx
ErikSin: Pristine vs dirty. So a pristine checkbox means it hasnt been touched, therefore it shouldnt show an error. But once it has been touched, it has been made dirty, so therefore show an error. It avoids having an error right away when the user hasnt done anything yet

achou11: ah ok. is there a different name we could use then? maybe `checkboxTouched` or `touched`? `pristine` is not super intuitive for me personally

ErikSin: It would change the logic if I used touched. I can use "untouched"?

achou11: sure 👍 generally weary of variable names with negatives (i'd rather just update the logic) but think that's fine

src/frontend/AppStack.tsx
ErikSin: Im going to resolve this for now, but yes this is something we should talk about!

achou11:
Closing this since #660 was merged already

PR #653, opened 2021-07-25
Added Deeplinking

ErikSin:
# DeepLinking

Deeplinking can now be done through mapeo://projInvite

this link takes the user to a placeholder page (Category chooser)

### To Do
* Have Deeplink navigate to bottom sheet that @achou11 is working on.
* Deeplink event listeners will need to also be added to the onboarding navigation stack


src/frontend/DeepLink.ts
```diff
@@ -0,0 +1,52 @@
...
+  params?: NavigationParams,
+  action?: NavigationNavigateAction
+) => boolean;
+
+export function deepLinkSetUp(navigate: nav) {
+  console.log("This is being activated");
+  if (Platform.OS === "android") {
+    Linking.addEventListener("url", ({ url }) => {
+      handleUrl(navigate, url);
+    });
```
achou11: Curious if the callback for add+remove event listener has to point to the same function object? I believe in the context of web, these two are not equivalent implementations:

```js
const handleUrl = () => {}

...
Foo.addEventListener('event', handleUrl)
...
Foo.removeEventLIstener('event', handleUrl)
```

vs.

```js
// This one is flawed because we're not technically removing the same handler in this case

Foo.addEventListener('event', () => handleUrl())
...
Foo.removeEventLIstener('event', () => handleUrl())
```

src/frontend/screens/MapScreen.js
```diff
@@ -42,6 +43,14 @@ const MapScreen = ({ navigation }: Props) => {
     [navigation, newDraft]
   );
 
+  React.useEffect(() => {
+    deepLinkSetUp(navigation.navigate);
+
+    return () => {
+      deepLinkTakeDown(navigation.navigate);
+    };
+  }, []);
```
achou11: I'd probably create a hook for this just to isolate the code a little more

ErikSin: Good idea!

achou11 (approved):
Couple of comments but lgtm

PR #656, opened 2021-07-26
OTF Sprint 5 feature branch

achou11:
Notes:
- less obvious things to test:
  - sync modal. there was some refactoring work for this component as part of the wifi bar refactor so it'd be good to make sure that this still works properly


src/frontend/screens/SendJoinRequestScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+  },
+});
+
+const SendJoinRequestScreen: NavigationStackScreenComponent = () => {
+  const { ssid } = useWifiStatus();
+
+  const deviceName = "Android " + getUniqueId().slice(0, 4).toUpperCase();
+
+  // TOOD: Need to properly generate
+  const verificationCode = Math.random().toString().slice(-5);
```
achou11: @gmaclennan just wanted to highlight this given [this comment](https://github.com/digidem/mapeo-mobile/pull/643#discussion_r676217582) on the original PR. Would love your thoughts/input!

src/frontend/hooks/useDeepLink.tsx
```diff
@@ -0,0 +1,46 @@
...
+      Linking.getInitialURL().then(url => {
+        handleUrl({ url });
+      });
+    }
+  }
+
+  function deepLinkTakeDown() {
+    Linking.removeEventListener("url", handleUrl);
+  }
+}
```
gmaclennan: @ErikSin wondering why not use the [React Navigation](https://reactnavigation.org/docs/4.x/deep-linking) built-in support for deep linking? It avoids us having to do our own routing (e.g. matching URLs and directing to the right screen) since react-navigation does that for us, and it also passes through the router params, which we will also need for passing the encoded project key. It will be easy then to add a path for each screen so that we can open screens that are not accessible in the app (e.g. the new screens you are creating).

FYI the adb command to test deep linking is:

```
adb shell am start -W -a android.intent.action.VIEW -d "mapeo://xxxxxxxx" com.mapeo.debug
```

To avoid using ADB also useful is https://play.google.com/store/apps/details?id=com.manoj.dlt

achou11: Yeah I'm thinking that using React Navigation's built-in support might cause fewer headaches down the line

ErikSin: Right! So the reasoning why I did this was if we move to react nav v5, its slightly different. So I wanted it to be cross compatible first, and then I think switching will be easy enough

achou11: just created a PR to use React Navigation's built-in support for it: #658

src/frontend/context/ProjectInviteContext.tsx
```diff
@@ -0,0 +1,42 @@
...
+}>({ removeInvite: () => null });
+
+export const ProjectInviteProvider = ({
+  children,
+}: React.PropsWithChildren<{}>) => {
+  const [invites, setInvites] = React.useState<ProjectInvite[]>([]);
+
+  // TODO: Need to integrate backend/interactions that add invite
+
+  // TODO: Need to incorporate disabling based on current route (needs coordination with useProjectInviteListener)
```
achou11: highlighting this todo since it's something I figured we can work on as a follow-up

src/frontend/screens/CreateOrJoinScreen.tsx
```diff
@@ -0,0 +1,88 @@
...
+      >
+        <Text style={styles.buttonText}>
+          <FormattedMessage {...m.createProject} />
+        </Text>
+      </Button>
+    </View>
+  );
+};
+
+//TODO research typesafe option
```
achou11: is this todo still relevant? pretty sure the `NavigationStackScreenComponent` annotation handles this too

src/frontend/screens/SendJoinRequestScreen.tsx
```diff
@@ -0,0 +1,126 @@
...
+            <FormattedMessage {...m.share} />
+          </Text>
+        </Button>
+      </View>
+    </View>
+  );
+};
+
+SendJoinRequestScreen.navigationOptions = () => ({
+  ...TransitionPresets.SlideFromRightIOS,
```
achou11: although it's more of an iOS pattern, we have this screen sliding in from the right side - felt more natural and intuitive than from the bottom

src/frontend/screens/CreateOrJoinScreen.tsx
ErikSin: Correct, this was an old comment that I just didn't erase

gmaclennan (commented):
Great work on this, there is a lot here! Some thoughts / feedback:

**Feature Flag**
How can we merge this without affecting existing functionality until it is ready? I suggest we create one app container from `AppStack` and one from `WithModalsStack` and conditionally add them based on a config variable. When we are ready for user-testing we can change this to be based on app settings (an experimental feature). E.g.

```js
const OriginalAppContainer = createAppContainer(AppStack)
const OnboardingAppContainer = createAppContainer(WithModalsStack)

const AppContainer = process.env['FEATURE_ONBOARDING'] ? OnboardingAppContainer : OriginalAppContainer
```

We can either do this with [`react-native-config`](https://github.com/luggit/react-native-config) or [`babel-plugin-transform-inline-environment-variables`](https://stackoverflow.com/a/37823398/3071863). I think with inline environment variables is more suited for our use, then we can test it with:

```sh
FEATURE_ONBOARDING=true npm run android
```

**Switch Navigator**
How do we choose which screen to show, particularly if we choose to default to the "test" project? I think we need to have a default loading screen, then based on the current project choose which screen to show in the switch navigator. In the loading screen 

```js
const LoadingScreen = ({ navigation }) => {
  // This needs to be added to ConfigContext - Gregor can do that
  const { projectId } = React.useContext(ConfigContext)
  
  React.useEffect(() => {
    // Might need to check whether configContext.state === 'loading' here?
    if (!projectId) navigation.navigate("Onboarding")
    else navigation.navigate("App")
  }, [projectId])
  
  return <LoadingAnimation />
}
```

**Project Invites**
Is it possible to add this without adding another context? I'm wondering if we need the context, or we can just wrap the whole AppContainer e.g.

```js
const AppContainerWrapper = () => {
  const nav = React.useRef()
  
  React.useEffect(() => {
    // TODO: Add this method to the API
    const unsubscribe = api.addInviteListener((invite) => {
      // we don't want to show this if the user is in the process of creating an observation
      // so can we navigate dependent on the current screen, or set a listener for when
      // the user saves the observation and then navigates to the invite modal?
      nav.current && nav.current.navigate("ProjectInviteModal", { invite: invite })
    })
    return () => unsubscribe()
  })
  
  return <AppContainer ref={nav} />
}
```

I think we can pass the invite around as a navigation param so that we use the same code whether the invite comes from the api (via the code above) or via a deep link (which also passes params) e.g.

```js
const ProjectInviteModal = ({ navigation }) => {
  const invite = navigation.getParam("invite")
  const [error, setError] = React.useState(invite ? null : new Error("Missing invite"))
  const [inviteDetails, setInviteDetails] = React.useState()
  
  React.useEffect(() => {
    // Gregor TODO: Add this method to the api
    api.decryptInvite(invite).then(data => setInviteDetails(data)).catch(e => setError(e))
  }, [invite])
  
  return error ? <ErrorReadingInvite /> : inviteDetails ? <ModalContent /> : <Loading />
}
```

**Leaving Project Screens**
These are implemented as 3 separate screens right now. I think that might be difficult for hooking up the API and capturing the state of whether leaving is complete. I suggest we implement this as a single screen with 3 states, something like:

```js
React.useEffect(() => {
  setStatus("leaving")
  api.leaveProject().then(() => {
    setStatus("done")
  })
}, [])
```

**Settings**
Are we going to add buttons to the ProjectConfig screen in the settings? E.g. leave project? We can conditionally show this based on the feature flag env variable / react-native-config as above.

gmaclennan:
One further thought: In terms of progressing this from:

turn on screens with env config -> users can turn on in settings

Perhaps we could use the projectSettings.experiments context from the start, initially setting it based on the environment variable but without any UI to change it, then adding the UI later means we don't have to change any of the conditions throughout the app to show the screens.

achou11:
Think using inline env variables for feature flagging will be a good way to go for our use case(s), so can work on integrating that.

> **Project Invites**
> Is it possible to add this without adding another context? I'm wondering if we need the context, or we can just wrap the whole AppContainer e.g.

I'm so annoyed that I didn't know that you could access the nav via the `ref`... yeah let's definitely go with the changes that you suggested for the invite stuff – it'll be much easier to implement and maintain

ErikSin:
# To Do
- [X] Feature Flag
- [ ] Loading Screen/Switch Navigator
- [X] Take out Project Invite Context
- [X] Pass invite as Nav Param (Instead of Context)
- [ ] Leave Project Screen Consolidation
~Add Button for leaving project~ (will be handled in https://github.com/digidem/mapeo-mobile/issues/663)

@achou11 @ximenabb

ErikSin (approved):

src/frontend/hooks/useProjectInviteListener.ts
```diff
@@ -0,0 +1,24 @@
...
+    callbackRef.current = onInviteReceived;
+  }, [onInviteReceived]);
+
+  useEffect(() => {
+    const handleInvite = (key: string) => callbackRef.current(key);
+    const unsubscribe = subscribeToProjectInvites(handleInvite);
+
+    return () => unsubscribe();
+  }, []);
+}
```
gmaclennan, line 24: Does this pattern have any unintentional side effects? The way I would write this would be:

```js
export default function useProjectInviteListener(onInviteReceived) {
  useEffect(() => {
    const unsubscribe = subscribeToProjectInvites(onInviteReceived)
    return () => unsubscribe()
  },
  [onInviteReceived]
  )
}
```

However subscriptions are a bit of a pain with React I think, and I'm unsure about what is "safe", e.g. see https://gist.github.com/bvaughn/e25397f70e8c65b0ae0d7c90b731b189
I wonder if there is an existing module we can use for this to make sure we're not going to hit some weird edge-case

achou11: To my knowledge it doesn't unless there's something super low-level. This subscription pattern is inspired by hooks similar to `useEventListener` [here](https://usehooks.com/useEventListener/) where using a ref for the callback prevents potentially over (un)subscribing listeners. not adamant keeping it this way and fine with going with the more traditional solution 

Other relevant references:
https://overreacted.io/making-setinterval-declarative-with-react-hooks/
https://github.com/donavon/use-event-listener

gmaclennan:
Merged via #715

Issue #631, opened 2021-07-08
(Front) Project Joiner - Deeplink Flow


ximenabb:
**Description :**
Project Joiner receives a deeplink via Signal to join project

**Screens**
- Deeplink flow
- Invite Screen

**Screen**
[Here
](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=45887%3A101&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1)
**Points**
13pts

PR #658, opened 2021-07-27
feature: use deeplinking via React Navigation

achou11:
Related to #631 

Notes:

- alternative implementation to #653 that uses React Navigation's built-in support for deeplinking.
- not an immediate need but probably will be good to have stricter or more usable path names for certain screens in our navigator (React Navigation defaults to the key name). For example, we could update the `MainStack` path to be `""` like so:

```js
  {
    Main: { screen: MainStack, path: "" },
    ProjectInviteModal,
  },
```
- tested using the following command:

```sh
# example path: mapeo://main/onboarding
adb shell am start -W -a android.intent.action.VIEW -d "mapeo://WHATEVER_PATH_YOU_WANT" com.mapeo.debug
```


src/frontend/App.js
```diff
@@ -26,6 +26,8 @@ const log = debug("mapeo:App");
 const NAV_STORE_KEY = "@MapeoNavigation@9";
 const ERROR_STORE_KEY = "@MapeoError";
 
+const DEEPLINK_URI_PREFIX = "mapeo://";
```
achou11: Is there a good place to keep constants like this? Thinking it'd be useful for anything that may generate a deeplink to have access to this (such as the Send Join Request screen)

ErikSin: Yes, great idea!

achou11: created `/src/frontend/constants.ts` for this

PR #659, opened 2021-07-27
Update WifiBar.tsx

ErikSin:
Added a default function to onPress of Wifi bar


achou11 (approved):
i think all consumers of this component currently provide an `onPress` that does exactly this, so up to you if you want to update them to remove the prop since it's the default now

PR #662, opened 2021-07-28
Add temporary qa-specific changes for OTF work

achou11:
Not planning to keep in the OTF branch. Just needed to make some changes so we build an APK that allows us to test the OTF work


achou11:
Was able to build locally so just resorting to that instead

PR #674, opened 2021-08-02
Refactor project invite implementation

achou11:
Based on [feedback](https://github.com/digidem/mapeo-mobile/pull/656#pullrequestreview-720151063) in #656 

Notes:
- Removes the context-based approach of connecting and coordinating project invites throughout the app. We instead use a straightforward (un)subscribe pattern that's initiated via a hook on the app container level.
- Adds loading and error states for the project invite modal
  - the error state is minimal and has two button options: close the modal or try to fetch the project invite details again

---

Preview:

refactored-invite-modal-flow


src/frontend/AppContainer.tsx
```diff
@@ -1,5 +1,144 @@
...
+const ERROR_STORE_KEY = "@MapeoError";
+
+const EDITING_SCREEN_NAMES = [
+  "AddPhoto",
+  "Camera",
+  "CategoryChooser",
+  "ManualGpsScreen",
+  "ObservationDetails",
+  "ObservationEdit",
+];
```
achou11, line 31: This list will inevitably grow but these screens stood out as the ones where we'd want the invite modal to not show up

src/frontend/AppContainer.tsx
```diff
@@ -1,5 +1,146 @@
...
+    }
+  }, []);
+
+  //   useProjectInviteListener(invite => {
+  //     if (inviteModalEnabled) {
+  //       openInviteModal(invite);
+  //     } else {
+  //       setQueuedInvite(invite);
+  //     }
+  //   });
```
achou11, line 122: When the backend supports subscribing to invite requests, we can uncomment this (and update `useProjectInviteListener`)

src/frontend/hooks/useProjectInviteListener.ts
```diff
@@ -1,14 +1,25 @@
...
-      navigation.navigate("ProjectInviteModal");
-    }
-  }, [invite, navigation]);
+import { useEffect, useRef } from "react";
+
+// Gregor TODO: Add method to API for (un)subscribing to project invites
+const subscribeToProjectInvites = (callback: (inviteKey: string) => void) => {
+  const intervalId = setInterval(() => callback("abc123"), 3000);
+  return () => clearInterval(intervalId);
+};
```
achou11, line 7: Placeholder for when the API supports the subscribe to project invites functionality

src/frontend/screens/ProjectInviteModal/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+                name: `Project ${Math.floor(Math.random() * 1000)}`,
+                id: "123",
+              },
+              role: "Project Participant",
+            })
+          : rej(new Error("Could not get invite details")),
+      1000
+    );
+  });
+};
```
achou11, line 89: placeholder for when we have an API endpoint that provides the invite details based on the invite request key/info that's provided

src/frontend/App.js
```diff
@@ -19,47 +18,7 @@ import useUpdateNotifierEffect from "./hooks/useUpdateNotifierEffect";
...
-          log("Crashed on last open, skipping load of navigation state");
-          return null;
-        } else {
-          return navState;
-        }
-      } catch (error) {
-        bugsnag.leaveBreadcrumb("Error loading nav state", { error });
-        log("Error reading navigation and error state", error);
-      }
-    };
```
achou11, line 62: moved these to `AppContainer.tsx`

src/frontend/AppContainer.tsx
```diff
@@ -1,5 +1,146 @@
...
+
+  React.useEffect(() => {
+    if (inviteModalEnabled && queuedInvite) {
+      setQueuedInvite(null);
+      openInviteModal(queuedInvite);
+    }
+  }, [inviteModalEnabled, queuedInvite, openInviteModal]);
```
achou11, line 129: Handles showing the invite modal after the user navigates to a screen that permits it, assuming there's an invite that's been queued. May be worth adding a small delay for the `openInviteModal` call since it can be a little visually jarring for some people

src/frontend/AppContainer.tsx
```diff
@@ -1,5 +1,146 @@
...
+  return (
+    <AppContainer
+      loadNavigationState={loadNavigationState}
+      onNavigationStateChange={onNavStateChange}
+      persistNavigationState={persistNavigationState}
+      ref={nav => {
+        if (nav) {
+          navRef.current = nav;
+        }
+      }}
```
ErikSin, line 140: Huh....cool. A little confused as to what is going on under the hood, but Im glad to know I can just do this

src/frontend/screens/ProjectInviteModal/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+  const sheetRef = React.useRef<BottomSheetModal>(null);
+
+  const inviteKey: string | null = navigation.getParam("invite");
+
+  const [status, setStatus] = React.useState<
+    LoadingStatus | ErrorStatus | SuccessStatus
+  >(
+    inviteKey
+      ? { type: "loading" }
+      : { type: "error", info: { error: new Error(t(m.noInviteFound)) } }
```
ErikSin, line 103: Can you just wrap this in anonymous function for optimization please!

src/frontend/screens/ProjectInviteModal/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+  const fetchInviteDetails = React.useCallback(async (inviteKey: string) => {
+    try {
+      setStatus({ type: "loading" });
+
+      const inviteDetails = await getProjectInviteDetails(inviteKey);
+
+      if (mountedRef.current) {
+        setStatus({ type: "success", info: { inviteDetails } });
+      }
```
ErikSin, line 114: Can you use `!!` when evaluating for truthy values please!

src/frontend/screens/ProjectInviteModal/index.tsx
```diff
@@ -0,0 +1,273 @@
...
+  type: "error";
+  info: { error: Error };
+};
+
+type SuccessStatus = {
+  type: "success";
+  info: {
+    inviteDetails: ProjectInviteDetails;
+  };
+};
```
ErikSin, line 37: I really like this pattern of using UseState() instead of the typical useReducer(). Much easier to read and understand. I will probably also start doing this.

src/frontend/hooks/useProjectInviteListener.ts
```diff
@@ -1,14 +1,25 @@
...
+
+export default function useProjectInviteListener(
+  onInviteReceived: (inviteKey: string) => void
+) {
+  const callbackRef = useRef(onInviteReceived);
+
+  useEffect(() => {
+    callbackRef.current = onInviteReceived;
+  }, [onInviteReceived]);
```
ErikSin, line 16: Im a little confused, where are we calling the function, onInviteReceived(), here?

ErikSin (commented):
Good Job, 1 really small change (just an optimization in the Project Invite Modal.

I also do not really understand the hook, so I am going to ask you about it tomorrow.

src/frontend/screens/ProjectInviteModal/index.tsx
achou11: yeah disjoint unions are super powerful when your state updates don't involve a lot of logic (in that case useReducer is usually the better solution)

src/frontend/screens/ProjectInviteModal/index.tsx
achou11: i don't think that's necessary in this case because this will only be run on initialization (i.e. even if `inviteKey` changes on rerenders). Providing the function option is more important if the value you're deriving from is 1) computationally expensive and 2) something that is only needed in the scope of the initialization (as opposed to the rest of the component)

in this case, getting `inviteKey` is probably not computationally expensive and it's also used elsewhere in the component, so it has to be computed at the component level and not just at the `useState` level

src/frontend/screens/ProjectInviteModal/index.tsx
achou11: in this case `mountedRef.currrent` is a `boolean` because we initialized the `useRef` with a boolean value (`true` in this case). When you provide an initial value, the type for `current` is inferred so `mountedRef.current` is of type `boolean`, which is why there's no coercion

src/frontend/hooks/useProjectInviteListener.ts
achou11: we're not calling it here, we're just updating the ref value whenever `onInviteReceived` does change. This is so that we don't potentially cause the effect that handles the subscription logic to run multiple times, but when the subscription is triggered, it uses an updated value of the callback that we provided.

you'll notice that in the useEffect with the subscription, we call `callbackRef.current`, which refers to the `onInviteRecieved` callback that's provided

achou11: the `useEventListener` hook [here](https://usehooks.com/useEventListener/) has a similar implementation that does a good job explaining this pattern and why it's useful:

```js

  // Update ref.current value if handler changes.
  // This allows our effect below to always get latest handler ...
  // ... without us needing to pass it in effect deps array ...
  // ... and potentially cause effect to re-run every render.
  useEffect(() => {
    savedHandler.current = handler;
  }, [handler]);
```

src/frontend/AppContainer.tsx
achou11: yeah this is more of an old-school way to assign component refs in React. Usually could just pass the `navRef` as the prop value but since react navigation is providing a class component instance (i think), you have to be more explicit about how you assign it (especially since TS is in play here)

PR #675, opened 2021-08-02
Add inline enivronment variable support for feature flagging

achou11:
Based on [feedback](https://github.com/digidem/mapeo-mobile/pull/656#pullrequestreview-720151063) from #656 

Notes:
- Integrates [babel-plugin-transform-inline-environment-variables](https://www.npmjs.com/package/babel-plugin-transform-inline-environment-variables) so that we can pass environment variables when starting the app that can be used within the app and when it's bundled e.g. `FEATURE_X=true npm start -- --reset-cache`
  - unfortunately, if you update/remove an env variable from your command, it looks like you need to pass the `--reset-cache` flag to `npm start`, or else it'll use an outdated definition


src/frontend/App.js
```diff
@@ -25,7 +25,9 @@ YellowBox.ignoreWarnings(["Require cycle:"]);
 if (__DEV__) debug.enable("*");
 const log = debug("mapeo:App");
 // WARNING: This needs to change if we change the navigation structure
-const NAV_STORE_KEY = "@MapeoNavigation@9";
+const NAV_STORE_KEY = `@MapeoNavigation@${
+  process.env.FEATURE_ONBOARDING ? 9 : 8
```
achou11: Slightly odd-looking but seems sensible to me. Open to other thoughts though

src/frontend/AppContainer.tsx
```diff
@@ -1,5 +1,8 @@
 import { createAppContainer } from "react-navigation";
 
+import { AppStack } from "./NavigationStacks/AppStack";
 import { WithModalsStack } from "./NavigationStacks/WithModalsStack";
 
-export default createAppContainer(WithModalsStack);
+export default createAppContainer(
+  process.env.FEATURE_ONBOARDING ? WithModalsStack : AppStack
```
achou11: Generally, are we okay with just checking the existence of the flag as opposed to a specific value? i.e. not checking if `process.env.FEATURE_ONBOARDING === "true"`

ErikSin (commented):
This is incomplete, correct? I may be missing a key part to this, but we havent included the commands or the variable name for the process.env.{variable}?

achou11:
> This is incomplete, correct? I may be missing a key part to this, but we havent included the commands or the variable name for the process.env.{variable}?

Depends. We can add commands if we want but this is more just so we can support providing the variables when working locally, so I think it makes sense that when working on the new stuff locally, you'd have to pass the env variable when starting your process e.g. `FEATURE_ONBOARDING=true npm start --reset-cache`. Not sure if we're ready to start doing this for any build commands since we want the default app to be the one without the OTF work

src/frontend/AppContainer.tsx
achou11: think i'm gonna opt for making it more explicit so that we check for the actual value too

ErikSin:
@achou11 Ahh I see. I think we should add it as a package.json command. I know the feature is technically not ready, but the people who are running the commands are either devs, or people on the program team, so having the command there and explicit would just make it overall easier

PR #677, opened 2021-08-04
Hotfix v5.2.1

gmaclennan:
Includes updated default config with Portuguese translations. A bug in the translations workflow caused them to be excluded from v5.2.0.


gmaclennan:
Going to merge this because I believe the failing test is due to a bug in the test suite that is now fixed in the develop branch. I have done a brief test on a device and it works for me. The only change should be an updated translations for the default config.

PR #683, opened 2021-08-10
Project config screen

ErikSin:
# Completed Project config screen

### Hidden under the OTF feature flag
Run script `start-otf` 

Screen is nested in the setting, under "Project Config"

## To Do:
Determine roles and who can "Add a Person"

###  Figma Screen:
[Screen](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=45997%3A72&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1)

### DIFF
--- a/src/frontend/screens/LeaveProject/LeaveProjectCompleted.tsx
+++ b/src/frontend/screens/LeaveProject/LeaveProjectCompleted.tsx
@@ -2,12 +2,7 @@ import React from "react";
 import { defineMessages, FormattedMessage } from "react-intl";
 import { View, Text, StyleSheet, Image } from "react-native";
 import Button from "../../sharedComponents/Button";
-import {
-  NavigationStackOptions,
-  NavigationStackScreenComponent,
-} from "react-navigation-stack";
 import { useNavigation } from "react-navigation-hooks";
-import { ILeaveSharedProp } from ".";
 
 const m = defineMessages({
   projectDeleted: {
@@ -24,15 +19,7 @@ const m = defineMessages({
   },
 });
 
-const navOption: NavigationStackOptions = {
-  headerShown: false,
-};
-
-export const LeaveProjectCompleted = ({
-  screenStateHook,
-}: ILeaveSharedProp) => {
-  const [screen, setScreen] = screenStateHook;
-
+export const LeaveProjectCompleted = () => {
   const nav = useNavigation();
 
   return (
--- a/src/frontend/screens/LeaveProject/LeaveProjectCompleted.tsx
+++ b/src/frontend/screens/LeaveProject/LeaveProjectCompleted.tsx
@@ -7,6 +7,7 @@ import {
   NavigationStackScreenComponent,
 } from "react-navigation-stack";
 import { useNavigation } from "react-navigation-hooks";
+import { ILeaveSharedProp } from ".";
 
 const m = defineMessages({
   projectDeleted: {
@@ -27,7 +28,11 @@ const navOption: NavigationStackOptions = {
   headerShown: false,
 };
 
-export const LeaveProjectCompleted: NavigationStackScreenComponent = () => {
+export const LeaveProjectCompleted = ({
+  screenStateHook,
+}: ILeaveSharedProp) => {
+  const [screen, setScreen] = screenStateHook;
+
   const nav = useNavigation();
 
   return (
@@ -55,8 +60,6 @@ export const LeaveProjectCompleted: NavigationStackScreenComponent = () => {
     </View>
   );
 };
-
-LeaveProjectCompleted.navigationOptions = navOption;
 
 const styles = StyleSheet.create({
   screenContainer: {
--- a/src/frontend/screens/LeaveProject/index.tsx
+++ b/src/frontend/screens/LeaveProject/index.tsx
@@ -1,0 +1,36 @@
+import React from "react";
+import { useState } from "react";
+import { NavigationStackScreenComponent } from "react-navigation-stack";
+import { LeaveProjectCompleted } from "./LeaveProjectCompleted";
+import { LeaveProjectInitial } from "./LeaveProjectInitial";
+import { LeaveProjectProgress } from "./LeaveProjectProgess";
+
+enum ScreenStates {
+  initial,
+  progress,
+  completed,
+}
+
+export interface ILeaveSharedProp {
+  screenStateHook: [
+    ScreenStates,
+    React.Dispatch<React.SetStateAction<ScreenStates>>
+  ];
+}
+
+const LeaveProjectScreen = () => {
+  const { initial, progress, completed } = ScreenStates;
+  const [screen, setScreen] = useState(initial);
+
+  if (screen === initial) {
+    return <LeaveProjectInitial screenStateHook={[screen, setScreen]} />;
+  }
+
+  if (screen === progress) {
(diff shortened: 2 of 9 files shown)


src/frontend/screens/LeaveProject/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  if (screen === initial) {
+    return <LeaveProjectInitial screenStateHook={[screen, setScreen]} />;
+  }
+
+  if (screen === progress) {
+    return <LeaveProjectProgress screenStateHook={[screen, setScreen]} />;
+  }
+
+  return <LeaveProjectCompleted screenStateHook={[screen, setScreen]} />;
```
achou11: this works but seems a little odd to me. i would just provide a callback as the prop like so i.e.

 ```tsx
const goToNextScreen = () => setScreen(previousScreen => previousScreen + 1)
...
<LeaveProjectInitial next={goToNextScreen} />
...
<LeaveProjectProgress next={goToNextScreen} />
...
```

File: src/frontend/screens/LeaveProject/LeaveProjectCompleted.tsx [hunk 2]
```diff
@@ -27,7 +28,11 @@ const navOption: NavigationStackOptions = {
   headerShown: false,
 };
 
-export const LeaveProjectCompleted: NavigationStackScreenComponent = () => {
+export const LeaveProjectCompleted = ({
+  screenStateHook,
+}: ILeaveSharedProp) => {
+  const [screen, setScreen] = screenStateHook;
```
achou11: this isn't being used here so i'd omit the prop entirely

src/frontend/screens/Settings/ProjectConfig.tsx
```diff
@@ -114,26 +117,85 @@ const ProjectConfig = () => {
...
           {projectKeySlice ? projectKeySlice + "**********" : "MAPEO"}
         </Text>
+
+        {!!configName && (
+          <React.Fragment>
+            <Text style={[styles.centerText, { marginTop: 10 }]}>
+              <FormattedMessage {...m.name} />
+            </Text>
+            <Text style={[styles.centerText, styles.projectKey]}>
+              {configName || ""}
```
achou11: the fallback shouldn't be necessary here because of the `!!configName` check

src/frontend/screens/Settings/ProjectConfig.tsx
```diff
@@ -114,26 +117,85 @@ const ProjectConfig = () => {
...
+        </React.Fragment>
+      )}
     </View>
   );
 };
 
 ProjectConfig.navigationOptions = {
   headerTitle: () => (
-    <HeaderTitle>
+    <HeaderTitle style={{}}>
```
achou11, line 198: is this necessary?

src/frontend/screens/Settings/ProjectConfig.tsx
```diff
@@ -192,4 +240,12 @@ const styles = StyleSheet.create({
     justifyContent: "center",
     alignItems: "center",
   },
+  LeaveBttn: {
```
achou11: nit, but curious why the Pascal-case here?

src/frontend/sharedComponents/Button.tsx
```diff
@@ -59,9 +59,7 @@ const Button = ({
     <View style={touchableStyle as ViewStyle}>
       {
         typeof children === "string" ? (
-          <Text style={[styles.textBase, textStyle]}>
-            {children.toUpperCase()}
-          </Text>
+          <Text style={[styles.textBase, textStyle]}>{children}</Text>
```
achou11: curious if this would "break" other parts of the app where we expect the copy to be all uppercase (although i agree with the change in general)

src/frontend/screens/LeaveProject/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  },
+});
+
+enum ScreenStates {
+  initial,
+  progress,
+  completed,
+}
+
+export interface ILeaveSharedProp {
```
achou11: Not a big deal but i'd lean towards making this more descriptive with the name because the `I` doesn't really give me much context as a newcomer

src/frontend/screens/LeaveProject/index.tsx
ErikSin: Yeah, I am unecessarily passing props here, I will change it!

src/frontend/screens/Settings/ProjectConfig.tsx
ErikSin: Right, I originally did this, and then decided to use !!configName afterwards and forgot to take it out. good catch

src/frontend/screens/Settings/ProjectConfig.tsx
ErikSin: Its throwing an error if I dont. And I didnt want to spend time converting the component to typescript (its not a simple conversion) I could also ts-ignore it, but I thought this was nicer

src/frontend/screens/Settings/ProjectConfig.tsx
ErikSin: just a typo!

src/frontend/sharedComponents/Button.tsx
ErikSin: Good Question. There might be some wierd casing with some button down the line. But I couldn't think of any thing that would actually break the app

achou11: yeah i just mean "break" as in will change the expected UI somewhere unintentionally. probably not a big deal but would be mindful of these kinds of changes

src/frontend/screens/LeaveProject/index.tsx
ErikSin: I think `I` is a pretty standard way for defining interfaces in the ts world! But we can also develop a best practice for us if you feel like that would be better!

achou11: ah got it. yeah i thought it was an abbreviation for `Initial` 😂 i've seen that pattern but usually for things that are closer to generics or for library authors, so i think i'd personally lean away from that for app code

src/frontend/sharedComponents/Button.tsx
ErikSin: OK, I will add this to the QA Script (specifically to take notice of any buttons that may be distorted or odd). We do have to change this eventually (for accessibility reasons). My mentality is to do it sooner rather then later so that we can catch bugs sooner.

src/frontend/screens/LeaveProject/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+export interface LeaveProjSharedProp {
+  next: () => void;
+}
+
+export const LeaveProjectScreen: NavigationStackScreenComponent = () => {
+  const { initial, progress } = ScreenStates;
+  const [screen, setScreen] = useState(initial);
+
+  function nextScreenState() {
+    setScreen(screen => screen + 1);
```
achou11: nit because JS scopes are a thing, but i'd rename the `screen` arg to something different than the state variable at the component level, just to completely avoid misreading

achou11 (approved):
left a non-blocking comment but lgtm otherwise

PR #684, opened 2021-08-10
Unable to link

ErikSin:
# Unable to link (using QR code) screen

This flow lets user know there was a problem adding a user using a QR code.
 
User will be given the option to see show thier own QR code or send a link

### To access screen:
Run the following command
`adb shell am start -W -a android.intent.action.VIEW -d "mapeo://Main/App/UnableToLeave" com.mapeo.debug`

### To do
QR Code, link, and device name are all placeholders while we wait for backend functionality

### Figma Screen:
[Screen](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=46006%3A0&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1)

### DIFF
--- a/src/frontend/NavigationStacks/AppStack.tsx
+++ b/src/frontend/NavigationStacks/AppStack.tsx
@@ -27,7 +27,7 @@ import { LeaveProjectProgress } from "../screens/LeaveProject/LeaveProjectProges
 import { LeaveProjectCompleted } from "../screens/LeaveProject/LeaveProjectCompleted";
 import { AlreadyOnProj } from "../screens/AlreadyOnProject";
 import { AddToProjectScreen } from "../screens/AddToProjectScreen";
-import UnableToLeaveScreen from "../screens/UnableToLink";
+import UnableToLinkScreen from "../screens/UnableToLink";
 import { JoinProjectQrScreen } from "../screens/Onboarding";
 
 const HomeTabs = createBottomTabNavigator(
@@ -81,7 +81,7 @@ export const AppStack = createStackNavigator(
     LeaveProjectCompleted: LeaveProjectCompleted,
     AlreadyOnProj: AlreadyOnProj,
     AddToProject: AddToProjectScreen,
-    UnableToLeave: UnableToLeaveScreen,
+    UnableToLink: UnableToLinkScreen,
     JoinProjectQr: JoinProjectQrScreen,
   },
   {
--- a/src/frontend/NavigationStacks/AppStack.tsx
+++ b/src/frontend/NavigationStacks/AppStack.tsx
@@ -85,7 +85,7 @@ export const AppStack = createStackNavigator(
     JoinProjectQr: JoinProjectQrScreen,
   },
   {
-    initialRouteName: "UnableToLeave",
+    initialRouteName: "Home",
     // TODO iOS: Dynamically set transition mode to modal for modals
     mode: "card",
     headerMode: "screen",
--- a/src/frontend/NavigationStacks/AppStack.tsx
+++ b/src/frontend/NavigationStacks/AppStack.tsx
@@ -27,6 +27,8 @@ import { LeaveProjectProgress } from "../screens/LeaveProject/LeaveProjectProges
 import { LeaveProjectCompleted } from "../screens/LeaveProject/LeaveProjectCompleted";
 import { AlreadyOnProj } from "../screens/AlreadyOnProject";
 import { AddToProjectScreen } from "../screens/AddToProjectScreen";
+import UnableToLeaveScreen from "../screens/UnableToLink";
+import { JoinProjectQrScreen } from "../screens/Onboarding";
 
 const HomeTabs = createBottomTabNavigator(
   {
@@ -79,9 +81,11 @@ export const AppStack = createStackNavigator(
     LeaveProjectCompleted: LeaveProjectCompleted,
     AlreadyOnProj: AlreadyOnProj,
     AddToProject: AddToProjectScreen,
+    UnableToLeave: UnableToLeaveScreen,
+    JoinProjectQr: JoinProjectQrScreen,
   },
   {
-    initialRouteName: "Home",
+    initialRouteName: "UnableToLeave",
     // TODO iOS: Dynamically set transition mode to modal for modals
     mode: "card",
     headerMode: "screen",
--- a/src/frontend/screens/Onboarding/JoinProjectQrScreen.tsx
+++ b/src/frontend/screens/Onboarding/JoinProjectQrScreen.tsx
@@ -1,8 +1,9 @@
 import * as React from "react";
 import { StyleSheet, View } from "react-native";
 import { useNavigation } from "react-navigation-hooks";
+import { StackActions, NavigationActions } from "react-navigation";
 import { NavigationStackScreenComponent } from "react-navigation-stack";
-import { FormattedMessage, defineMessages } from "react-intl";
+import { FormattedMessage, defineMessages, useIntl } from "react-intl";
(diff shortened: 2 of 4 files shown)


src/frontend/screens/Onboarding/JoinProjectQrScreen.tsx
```diff
@@ -26,14 +27,41 @@ const m = defineMessages({
...
   },
+  cancel: {
+    id: "screens.Onboarding.JoinProjectQrScreen.cancel",
+    defaultMessage: "Cancel",
+  },
+});
+
+//Cancels the QR code screen and resets the navigation stack so the
+//user CANNOT hit the back button to go back to this screen
+const resetAction = StackActions.reset({
```
achou11, line 46: curious but does `StackActions.pop` not do this? just from the name i'd imagine it would remove it from the stack and therefore be unable to navigate back to it via the back button, but not super familiar with the details of the API

src/frontend/screens/Onboarding/JoinProjectQrScreen.tsx
```diff
@@ -26,14 +27,41 @@ const m = defineMessages({
...
+
+  IsAdmin:boolean Flag is passed (through navigation) if the 
+  user is in an admin
+*/
 export const JoinProjectQrScreen: NavigationStackScreenComponent = () => {
   const navigation = useNavigation();
+  const { formatMessage: t } = useIntl();
+
+  const isAdmin: boolean =
+    JSON.stringify(navigation.getParam("isAdmin", false)) === "true";
```
achou11, line 64: do you need to stringify here if `isAdmin` is supposed to be a boolean? or will it sometimes be a string too and therefore safer to convert anyways?

src/frontend/screens/Onboarding/JoinProjectQrScreen.tsx
ErikSin: `pop` and `back` do the same thing. So `pop` will just take the current screen of the stack, bringing you back to the previous screen. But in this case, I want the user to go back home AND for the stack to completly reset

src/frontend/screens/Onboarding/JoinProjectQrScreen.tsx
ErikSin: yeah.....all the documentation suggest that I stringify it....I really dont understand why, but I figure its only one line of code so why not!

src/frontend/screens/UnableToLink.tsx
```diff
@@ -0,0 +1,132 @@
...
+  // TODO: Decide on a URL structure for this deep link
+  const shareLink = `${URI_PREFIX}main/onboarding?code=${verificationCode}`;
+
+  async function handleShareInvite() {
+    await Share.share({ message: shareLink, url: shareLink });
+  }
+
+  return (
+    <View>
+      <WifiBar deviceName="TO DO-GET NAME" ssid={ssid} />
```
achou11, line 63: Would require moving some files around, but you could potentially use the `WithWifiBar` component to avoid some boilerplate related to including that wifi status. Not adamant about introducing it but if it works well for this use case, it's an option

src/frontend/screens/Onboarding/JoinProjectQrScreen.tsx
achou11: yeah makes sense. guess it's possibly to safeguard against navigation initiated by deeplinks that contain params in the link? (in which case it'd always be a string unless you parsed it)

src/frontend/screens/Onboarding/JoinProjectQrScreen.tsx
```diff
@@ -48,18 +76,34 @@ export const JoinProjectQrScreen: NavigationStackScreenComponent = () => {
...
-          </Text>
-        </Button>
+
+        {isAdmin ? (
+          //Admin cancel Button
+          <Button
+            variant="outlined"
+            onPress={() => navigation.dispatch(resetAction)}
+          >
+            {t(m.cancel)}
```
achou11, line 94: is there a reason we can't use the `<Text ...><FormattedMessage ... /></Text>` style here?

src/frontend/screens/UnableToLink.tsx
```diff
@@ -0,0 +1,132 @@
...
+        >
+          {t(m.showCode)}
+        </Button>
+
+        <Button
+          style={[styles.buttons]}
+          onPress={handleShareInvite}
+          variant="outlined"
+        >
+          {t(m.shareInvite)}
```
achou11, line 95: similar question as [this](https://github.com/digidem/mapeo-mobile/pull/684/files#r686313615) about using the `Text` and `FormattedMessage` components

src/frontend/NavigationStacks/AppStack.tsx
```diff
@@ -26,6 +26,8 @@ import { LeaveProjectScreen } from "../screens/LeaveProject/LeaveProject";
 import { LeaveProjectProgress } from "../screens/LeaveProject/LeaveProjectProgess";
 import { LeaveProjectCompleted } from "../screens/LeaveProject/LeaveProjectCompleted";
 import { AlreadyOnProj } from "../screens/AlreadyOnProject";
+import UnableToLeaveScreen from "../screens/UnableToLink";
```
achou11: Should the component name be `UnableToLinkScreen`? (also might as well convert to a named export too)

src/frontend/NavigationStacks/AppStack.tsx
```diff
@@ -77,6 +79,8 @@ export const AppStack = createStackNavigator(
     LeaveProjectProgress: LeaveProjectProgress,
     LeaveProjectCompleted: LeaveProjectCompleted,
     AlreadyOnProj: AlreadyOnProj,
+    UnableToLeave: UnableToLeaveScreen,
```
achou11: Guessing the route name should be `UnableToLink`?

src/frontend/screens/Onboarding/JoinProjectQrScreen.tsx
ErikSin: This jsut avoids having to style the text. By using a string directly, the button styling automatically applies to the text

achou11: oh interesting, good to know!

src/frontend/NavigationStacks/AppStack.tsx
ErikSin: haha im actually going to change those other ones to default exports (ill follow the architecture of making screen default exports and screen states name exports)

src/frontend/NavigationStacks/AppStack.tsx
ErikSin: Changed! Good catch

src/frontend/NavigationStacks/AppStack.tsx
```diff
@@ -79,6 +81,8 @@ export const AppStack = createStackNavigator(
     LeaveProjectCompleted: LeaveProjectCompleted,
     AlreadyOnProj: AlreadyOnProj,
     AddToProject: AddToProjectScreen,
+    UnableToLink: UnableToLinkScreen,
+    JoinProjectQr: JoinProjectQrScreen,
```
achou11, line 85: Does it make sense to name this slightly different than what's defined in the onboarding stack? Not totally sure myself but just noticed that there are now two routes pointing to the same component (although for different use cases)

achou11 (approved):
Left a question that might be worth thinking about but lgtm otherwise

PR #682, opened 2021-08-09
Implement Add to Project flow

achou11:
Closes #664
Closes #665
Closes #666

Notes:
- ran into issues with using `expo-camera`'s QR scanning feature and pinpointed it to an issue with our older version of the dependency. had to upgrade it and `react-native-unimodules`, which also means I had to upgrade many other Expo modules that we use. I combed through changelogs for each one and especially tested the camera screen to ensure that no breaking changes would be introduced with these upgrades - so far, so good
  - `react-native-unimodules` [changelog](https://github.com/expo/expo/blob/master/packages/react-native-unimodules/CHANGELOG.md#0101--2020-05-29)
  - `expo-camera` [changelog](https://github.com/expo/expo/blob/master/packages/expo-camera/CHANGELOG.md#830--2020-05-27)
  - `expo-linear-gradient` [changelog](https://github.com/expo/expo/blob/master/packages/expo-linear-gradient/CHANGELOG.md#821--2020-05-29)
  - `expo-localization` [changelog](https://github.com/expo/expo/blob/master/packages/expo-localization/CHANGELOG.md#821--2020-05-29)
  - `expo-location` [changelog](https://github.com/expo/expo/blob/master/packages/expo-location/CHANGELOG.md#830--2020-07-16)
  - `expo-sensors` [changelog](https://github.com/expo/expo/blob/master/packages/expo-sensors/CHANGELOG.md#901--2020-05-29)
- implements this flow as a single screen with various states: scanning, when the device is found, and when the device is successfully added. i'm assuming we'll need another intermediate loading screen that is displayed after inviting a device and waiting for it to respond, but could do that as a follow-up

---

Preview:

add-to-project-flow

Issue #666, opened 2021-07-29
(Front) Project Coordinator - Add device success


ximenabb:
Description :
Project Coordinator is on the Project Config screen and clicks on "Add Person". Scans QR and successfully adds device

Screen is:
Connecting Device
Success!
XXXX (Device ID) has been added to XXXX (Project name)
(GO TO SYNC)

**Screen**
[Here
](https://www.figma.com/proto/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=45997%3A744&scaling=scale-down-width&page-id=45808%3A2&hide-ui=1)

**Points**
5pts

PR #688, opened 2021-08-16
Add connecting step in Add to Project flow

achou11:
Part of #666 that I forgot to implement in #682 

Notes:
- updates the message ids for the add to project steps
- adjusted action in success step to navigate to home screen (map), mostly just for QA purposes so you're not stuck in the add to project flow loop

---

Preview:

add-to-project-flow

### DIFF
--- a/src/frontend/screens/AddToProjectScreen/index.tsx
+++ b/src/frontend/screens/AddToProjectScreen/index.tsx
@@ -1,6 +1,7 @@
 import * as React from "react";
 import { View, StyleSheet } from "react-native";
 import { NavigationEvents } from "react-navigation";
+import { useNavigation } from "react-navigation-hooks";
 import { NavigationStackScreenComponent } from "react-navigation-stack";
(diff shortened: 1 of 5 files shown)


src/frontend/screens/AddToProjectScreen/index.tsx
```diff
@@ -9,27 +10,49 @@ import { MEDIUM_BLUE, WHITE } from "../../lib/styles";
...
 
+  const sendInvite = async (deviceId: string) => {
+    try {
+      setStep("connect");
+      await sendInviteToDevice(deviceId);
+      setStep("success");
+    } catch (err) {
+      // TODO: provide some messaging about failure state
+      setStep("found");
+    }
```
achou11, line 53: We need some designs for this. I'm assuming at the very least we can bring the user back to the found step, but maybe we'd want a separate screen/step if the invited device declined the invite. Will ping Sabella

PR #695, opened 2021-08-20
chore: enable TestButler for e2e tests

gmaclennan:
See https://github.com/wix/Detox/blob/master/docs/Introduction.Android.md#8-test-butler-support-optional

I have not noticed a difference, but might help in CI. Spotted this when debugging another Detox testing issue.


ErikSin:
I'm not really sure how to check this? So we can probably just approve it, @achou11 what do you think?

achou11 (approved):
Changes align with the documentation so if you haven't noticed any differences, lgtm

PR #696, opened 2021-08-20
chore: lock and cache gradle dependencies

gmaclennan:
Currently builds are non-deterministic because in Gradle our project uses some dynamic dependencies (similar to how version ranges work in npm). This means that our build could break if a dependency has a breaking change that we were not expecting.

We've also seen a lot of issues with failures in dependency downloads. This PR also caches dependencies so they should load from the cache rather than the source repository.


gmaclennan:
Had second thoughts about the locking here. Because of the way react native works we could end up with lots of lock file conflicts that we are forever resolving (already seeing this in the bitrise build). Going to close this and open a new PR that just does dep caching.

PR #697, opened 2021-08-20
chore: cache nodejs-native-assets

gmaclennan:
Every CI run we rebuild leveldb for Android, which takes 15-20 minutes. We are locking the version of leveldb in `src/backend/package-lock.json` so there is no need to rebuild it every time. This is an attempt to cache the built leveldb to avoid having to rebuild it on each CI run. Hopefully this will speed up CI runs on Github significantly.

The better solution to this is probably to publish our own leveldown package on npm with prebuilds for Android, but I'm not so sure how to do that right now.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -93,9 +93,18 @@ jobs:
       - name: Build translations
         run: |
           npm run build:translations
+      - name: nodejs-mobile build cache
+        id: cache
+        uses: actions/cache@v2
+        with:
+          path: android/build/nodejs-native-assets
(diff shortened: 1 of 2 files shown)


achou11 (approved):
🤞

gmaclennan:
This did not lead to as large a speed up as I had hoped. It significantly reduces the time spend building leveldb, and the resulting build still works, but it seems like Gradle build takes time on other tasks now. I wonder whether there are some issues with the Github MacOS image, so I'm testing Ubuntu in #700

gmaclennan:
Moved to #708 and merged with other improvements

PR #699, opened 2021-08-21
chore: cache npm dependencies in Github CI

gmaclennan:
Will hopefully speed up builds - sometimes npm install takes 5 minutes!


gmaclennan:
See #708

PR #700, opened 2021-08-21
chore: Run Github CI on Ubuntu

gmaclennan:
**Experimental**

Github CI has been running very slowly with MacOS for simple things like
npm ci and the Android emulator also runs very slowly. We originally used the MacOS image because Detox tests were not running reliably on Ubuntu, but this might have changed now, so this is an experiment to see if switching to Ubuntu works and if it is any quicker to run the tests.


gmaclennan:
Moved to #708

PR #708, opened 2021-08-23
chore: speedup CI e2e tests

gmaclennan:
This PR attempts to increase the speed of end-to-end tests that run on Github actions. These tests have been taking 35-40 minutes to complete. This PR includes several changes, each of which should make incremental gains to the test speed:

- [x] **Build on Ubuntu**. Currently we build and run tests on MacOS. The tests need to run on MacOS because only the MacOS images on Github actions have hardware acceleration for the Android emulator. The Android emulator runs incredibly slowly on the Github Actions Ubuntu image. However, the build step is much faster on Ubuntu. This change builds the test APKs first on Ubuntu, then uses those APKs to run the e2e tests on MacOS.
- [x] **npm & Gradle dependency cache**. We cache the dependencies used by npm and Gradle. This results in a small speed up (the cost of loading the cache is close to the speed gain) but importantly it also avoids problems we have seen with errors from Gradle repositories causing the dependency download to fail, resulting in a failing test.
- [x] **nodejs-mobile native module cache**. Building native modules for nodejs-mobile is one of the most costly steps in the build. Mapeo Mobile needs leveldb to be built for Android. However, this rarely changes between builds, so we can cache the built leveldb between test runs to avoid building it every time.
- [x] **Gradle parallel builds**. Pass the `--parallel` option to Gradle. Results in a small speed increase, and does not (yet) seem to have adverse side-effects.
- [x] **Start emulator in the background**. Currently we start the emulator, then wait for it to boot before continuing. We can save some time by starting the emulator in the background and executing a different task whilst it boots.
- [x] **Switch to 64-bit emulator image**. We were using a 32-bit Android image. The 64-bit image seems to run a bit faster.
- [x] **Latest emulator version**. Not sure about the speedup, but hopefully installing the latest (30.8.4) version of the emulator vs the default (30.7) will improve the emulator
- [x] **Turn off verbose Detox logging**. This was on for debugging, but seems to slow down the test run a bit.
- [x] **Turn off postinstall scripts**. For the e2e test job we do not need any of the postinstall scripts to run (the scripts are used for Detox iOS builds and for React Native build steps). Skipping these scripts improves the `npm ci` time.

Together these changes seem to improve the build+test time from 35-40 minutes to about 25 minutes. The Android build step (`npm run build-detox-android`) in particular has been reduced from 20-25 minutes to 11-12 minutes. There is still a cost of splitting the task across two jobs - build on Ubuntu then test on MacOS - but the speed gains are worth it. It's hard to benchmark because it seems like the speed of the Github Actions MacOS instances varies a lot, but seems like this should shave some time off our builds overall.

**Future steps**
We might be able to speed up builds by using a specialized build service like [EAS Build](https://docs.expo.dev/build/introduction/). However about half the time is running the e2e tests in the emulator, which is still quite slow even with just a handful of tests. Genymotion integrates with Detox for running e2e tests on an emulator on their infrastructure, but it is expensive. We might have to stick with this free option (Github actions) for now.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -41,6 +41,12 @@ jobs:
           key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-node-
+      - name: Cache Gradle files
+        uses: actions/cache@v1
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}-${{ hashFiles('**/gradle/wrapper/gradle-wrapper.properties') }}
+          restore-keys: ${{ runner.os }}-gradle-
       - name: Node
         uses: actions/setup-node@v1
       - name: Use specific Java version for sdkmanager to work
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -125,17 +125,19 @@ jobs:
           kextstat | grep intel
           echo "\nEmulator config"
           cat $HOME/.android/avd/emu.avd/config.ini
-      - name: Install node_modules
-        run: |
-          mkdir -p nodejs-assets
-          npm ci
-      - name: Android Emulator
-        timeout-minutes: 10
-        continue-on-error: true
+      - name: Start Android Emulator
         run: |
           echo "Starting emulator"
           mkdir -p artifacts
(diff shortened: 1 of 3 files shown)


gmaclennan:
This seems to be stabilising at about 25-27 minutes for running the build and the tests. This is a saving of about 10-15 minutes on our previous build & test time. Since this is not touching any application code, only the CI build config, and as tests still seem to be passing, I'm going to go ahead and merge this so future PRs can benefit from the faster run time.

PR #709, opened 2021-08-23
chore(CI): setup Cirrus CI for running build & e2e tests

gmaclennan:
This PR adds [Cirrus CI](https://cirrus-ci.org) for running Android end-to-end tests. This is an attempt to solve the slow running of tests on Github Actions. #708 reduced the time taken by build & tests from 35-50 minutes down to 25 minutes, but it is still slow. One of our limitations with testing is that we are running the e2e tests in the Android Emulator (by necessity) and this requires VM acceleration to run at a decent speed, which needs KVM enabled on Linux, or HAXM enabled on MacOS. Only the MacOS machines on Github Actions have VM acceleration, and most CI services do not have KVM enabled for CI instances.

Cirrus CI allows you to explicitly request an instance with KVM enabled, which runs the emulator a lot faster. Their free community cluster is also very generous, with up to 8 CPUs and 32GB available.

The build step on Cirrus CI runs in about 6 minutes, and the e2e tests run in about 5 minutes. Overall the build and test is running in 11-12 minutes, which is a significant improvement on what we can get on Github Actions.

I'm going to merge this since it does not change any existing code, it just adds another check to each PR. We can see how it continues to perform, and if it continues to work well then we can switch off the Github Actions and set Cirrus CI as the required check for PR merges.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -8,7 +8,7 @@ connected_check_task:
     JAVA_TOOL_OPTIONS: -Xmx6g
     GRADLE_OPTS: -Dorg.gradle.daemon=true -Dorg.gradle.parallel=true -Dorg.gradle.jvmargs=-Dfile.encoding=UTF-8 -Dkotlin.incremental=false -Dkotlin.compiler.execution.strategy=in-process
   container:
-    image: cirrusci/android-sdk:29
+    image: digidem/docker-android
     kvm: true
     cpu: 8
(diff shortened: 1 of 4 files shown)


gmaclennan:
Squashed and merged without Bitrise tasks completing, because they are backed up right now, and because this PR only touches the `.cirrus.yml` file, which is ignored by Bitrise.

PR #711, opened 2021-08-23
fix: fix merging of settings state in settings context

achou11:
Updates are applied left to right according to the [docs](https://docs-lodash.com/v4/merge/) so I think we'd want the existing values in the default settings to be overridden by values defined in the new state. Otherwise some user-defined changes can potentially be overridden by a default value (e.g. coordinate system format)


gmaclennan:
Yes, you are correct. I was not thinking clearly when writing this! This is why I need to start writing unit tests more!

gmaclennan (approved):

PR #723, opened 2021-08-27
Api to Typescript Conversion

### LINKED ISSUE #706: Converting Api file to Typescript
Currently the Api.js (src/api.js) file is written with flow annotations.

This files needs to be converted in typescript from flow

**Points**
13pts

ErikSin:
Converted Api.js to Api.ts

Converted the REST api calls using type assertions.

Fixes #706


src/frontend/api.ts
```diff
@@ -301,80 +367,67 @@ export function Api({
...
-        data => {
-          const messages = data && data[locale];
-          if (!messages) return {};
-          return flatten(messages);
-        }
-      );
+    getConfigMessages: async (locale = "en") => {
+      const data = await get(`presets/default/translations.json?${Date.now()}`);
+      // @ts-ignore
+      const messages = data && data[locale];
```
ErikSin, line 337: @gmaclennan I am a litte confused by this, but I will ask at the group hug on monday to get it sorted out.

src/frontend/context/ObservationsContext.js
```diff
@@ -31,8 +31,19 @@ export type Observation = {
   userId?: string,
   type: "observation",
   links?: string[],
-  schemaVersion: 4,
+  schemaVersion: number,
```
gmaclennan: This was actually intended to be `4`, rather than `number`, since when this changes the type is different.

src/frontend/context/ConfigContext.tsx
```diff
@@ -102,7 +102,7 @@ type Props = {
 export const ConfigProvider = ({ children }: Props) => {
   const [config, setConfig] = React.useState(defaultConfig);
   const [status, setStatus] = React.useState<Status>("idle");
-  const [reloadToken, setReloadToken] = React.useState();
+  const [reloadToken, setReloadToken] = React.useState<{}>();
```
gmaclennan: This is not a great pattern. I am to blame.

src/frontend/lib/mapeo-schema.d.ts
```diff
@@ -0,0 +1,189 @@
+// Refer to https://github.com/digidem/mapeo-schema/blob/master/docs/README.md
+declare module "mapeo-schema" {
+  // TODO: do we need this?
```
gmaclennan: The idea is that all database types are an extension of this, e.g.

```ts
type Observation = Common & {...}
```

However presets and fields are not yet stored in the database, and we do not use `Node` and `Way` types here, so it's not that useful when the only type based on `Common` is `Observation`.

src/frontend/lib/mapeo-schema.d.ts
```diff
@@ -0,0 +1,189 @@
...
+    tags: {
+      [key: string]: any;
+    };
+    terms?: string[];
+  }
+
+  // TODO
+  type MbFilter = (string | number | boolean | MbFilter)[];
+
+  export interface Filter {
```
gmaclennan: Not yet implemented.

src/frontend/api.ts
```diff
@@ -154,15 +167,55 @@ const SERVER_START_TIMEOUT = 30000;
...
 const pixelRatio = PixelRatio.get();
 
-export function Api({
-  baseUrl,
-  timeout = DEFAULT_TIMEOUT,
-}: {
-  baseUrl: string,
-  timeout?: number,
-}) {
+interface PublicApi {
```
gmaclennan: Nit-pick: I feel like this is easier to maintain when we define these types inline above the functions that they refer to. Typescript will imply the same of the exported api.

src/frontend/api.ts
gmaclennan: We support the category/preset names being translated into different languages, as well as translations for the questions/fields. We do this for the default config: https://github.com/digidem/mapeo-default-config/tree/master/messages. See here for the structure of the built `translations.json` file: https://unpkg.com/mapeo-default-settings@3.5.0/dist/translations.json

gmaclennan (commented):
Really great, glad you and @achou11 are ploughing through this conversion, this will really help going forwards.

src/frontend/api.ts
achou11: Yeah i implemented this interface initially to get a better idea of the API type, but looking back it's definitely harder to work with than just inlining the annotations and taking advantage of TS inference. Just updated to fix via https://github.com/digidem/mapeo-mobile/pull/723/commits/1f03d570fb9f6392a749726794287b1613f9d76b 👍

src/frontend/context/ConfigContext.tsx
achou11: I've seen this pattern (or similar using a "nonce" 😆 ) in other codebases - we've all been there 😅

src/frontend/context/ObservationsContext.js
achou11: ah makes sense. will have to make some adjustments elsewhere if we restore this.

if we restore this to `4`, there'll be a TS error here because we're trying to match a `number` type (defined in [`mapeo-schema.d.ts`](https://github.com/digidem/mapeo-mobile/blob/1f03d570fb9f6392a749726794287b1613f9d76b/src/frontend/lib/mapeo-schema.d.ts#L52)) to `4`:

https://github.com/digidem/mapeo-mobile/blob/1f03d570fb9f6392a749726794287b1613f9d76b/src/frontend/api.ts#L642

src/frontend/lib/mapeo-schema.d.ts
achou11: Makes sense. can remove if it just adds noise for now (same for node and way types)

src/frontend/lib/mapeo-schema.d.ts
achou11: can remove to reduce noise if desired 👍

src/frontend/lib/mapeo-schema.d.ts
```diff
@@ -0,0 +1,189 @@
...
+        gpsAvailable?: boolean;
+        locationServicesEnabled?: boolean;
+        networkAvailable?: boolean;
+        passiveAvailable?: boolean;
+      };
+    };
+    refs?: {
+      id: string;
+    }[];
+    schemaVersion: number;
```
achou11: @gmaclennan based on https://github.com/digidem/mapeo-mobile/pull/723/commits/3f2c612b2696b953e1913cdabc9e3131d361d8da#r697847676, should this be `4` then? Seems like it should based on https://github.com/digidem/mapeo-schema/blob/master/docs/observation.md#schemaversion

src/frontend/context/ObservationsContext.js
achou11: just realized the `convertFromServer` function I highlighted in above comment is meant for converting observations from `3` to `4`, so can disregard

achou11: actually scratch that – i'm a little bit confused about what's going on. I'm assuming that the type from `mapeo-schema` should reflect what's being returned by the server right? however, looks like we're inserting/updating observation values we persist as version 3:

```ts
    createObservation: async (
      value: ObservationValue
    ): Promise<Observation> => {
      const valueForServer = {
        ...value,
        type: "observation",
        schemaVersion: 3,
      };
      const serverObservation = (await post(
        "observations",
        valueForServer
      )) as ServerObservation;
      return convertFromServer(serverObservation);
    },
```

Based on https://github.com/digidem/mapeo-schema/blob/master/docs/observation.md#schemaversion, the server version _should_ be 4 but that doesn't align with what's currently implemented. Is our implementation just outdated or am I missing something?

src/frontend/lib/mapeo-schema.d.ts
achou11: Decided to keep it for now but open to removing if you have strong feelings on it

src/frontend/lib/mapeo-schema.d.ts
gmaclennan: Yes. The `mapeo-schema` repo exports Flow types (which are generated from the JSON Schema). The goal is to have a single source of truth for types/schema. The `mapeo-schema` repo should probably be updated to export TypeScript rather than Flow. One option that works well is https://github.com/sinclairzx81/typebox. Another option is to use https://jsontypedef.com instead of Json Schema, which maps more easily to Typescript, but does not support everything that Json Schema does.

src/frontend/lib/mapeo-schema.d.ts
gmaclennan: Yes let's remove for now. Maybe these types can be published with the `mapeo-schema` repo in the future.

achou11: Cool. yeah ideally we'd just add this type def file to `mapeo-schema` instead of creating it here

src/frontend/context/ObservationsContext.js
gmaclennan: _ermmm..._ I am also confused, this does not seem correct since I thought we were using schema version `4`. I might have got in a muddle about this, or forgotten to update this piece.

src/frontend/context/ConfigContext.tsx
achou11: Addressed this in #730

gmaclennan: 👍

src/backend/constants.js
```diff
@@ -1,5 +1,8 @@
 // @flow
 
+//In the api.tsx file, I copy and pasted this typing (directly under the imports) in typescript.
+//I didn't want to break the flow typing, but when this gets converted to ts, we also need to edit
+//the api.tsx file to just import this instead of hardcoding it in the file.
```
gmaclennan: Is this used anywhere else in the frontend other than `api.ts`? The backend is not using flow, so the flow typing is perhaps not needed. However, the backend currently does not use a buildstep to compile typescript, it does typing in JSDoc, so we should type this in JSDoc, or use a `constants.d.ts` file.

src/frontend/api.ts
```diff
@@ -195,13 +210,13 @@ export function Api({
     status = value;
 
     if (status === STATUS.LISTENING) {
-      while (pending.length) pending.shift().resolve();
+      while (pending.length) pending.shift()?.resolve();
```
gmaclennan, line 189: I'm guessing the `?` is here because Typescript isn't smart enough to know that if `pending.length` then `pending.shift()` will never be undefined?

src/frontend/api.ts
```diff
@@ -384,26 +402,25 @@ export function Api({
...
       id: string,
       value: ObservationValue,
-      options: {|
-        links: Array<string>,
-        userId?: $ElementType<ServerObservation, "userId">,
-      |}
-    ): Promise<Observation> {
+      options: {
+        links: Array<string>;
+        userId?: string;
```
gmaclennan:
```suggestion
        userId?: ServerObservation["userId"];
```

src/frontend/context/ObservationsContext.tsx
```diff
@@ -1,60 +1,74 @@
...
+    manualLocation?: boolean;
+    position?: Position;
+    positionProvider?: {
+      gpsAvailable?: boolean;
+      locationServicesEnabled?: boolean;
+      networkAvailable?: boolean;
+      passiveAvailable?: boolean;
+    };
+  };
+}
```
gmaclennan: This seems incorrect, the metadata should be on the `ObservationValue`. Is there a reason for this change?

gmaclennan: I think there are errors in the existing implementation with how this is handled and converted. `ObservationValue` was implemented at the app level at some point to make it easy to separate the properties that are user-writeable from the read-only properties (that are managed by the server). It looks like mapeo-schema and the implementation in mobile have drifted out of sync. We need to look at how collected data will have been saved and modify the schema/type to reflect that.

src/frontend/lib/mapeo-schema.d.ts
```diff
@@ -0,0 +1,127 @@
...
+      lastSavedPosition?: Position;
+      manualLocation?: boolean;
+      position?: Position;
+      positionProvider?: {
+        gpsAvailable?: boolean;
+        locationServicesEnabled?: boolean;
+        networkAvailable?: boolean;
+        passiveAvailable?: boolean;
+      };
+    };
```
gmaclennan: I think that this _may_ have been implemented differently in mobile than what the schema says. Since mobile is the only app that is creating data, we should probably modify the schema to match the data that has been collected.

gmaclennan (changes requested):
Thanks for all the work on this, the conversion to Typescript is a big help to everything. I think the existing implementation of the Observation and Observation Value, and the schema version that was being used by the app, were incorrect, since the schema and the app implementation seems to have drifted apart. Before we merge this I would like to be sure how data has been written by mobile in the past, and modify the types to reflect this.

src/frontend/api.ts
achou11: it's there because the returned result of `shift` could be undefined based on `pending`'s type def + initialization

```ts
const pending: Array<{
  resolve: () => any;
  reject: (err: Error) => void;
}> = [];
```

src/frontend/context/ObservationsContext.tsx
achou11: > This seems incorrect, the metadata should be on the ObservationValue. Is there a reason for this change?

hm not sure. looks like it was added via https://github.com/digidem/mapeo-mobile/pull/723/commits/9b52009da9ebeb6325812f41b45a953f6c6eed91 (probably because TS was complaining about a missing `metadata` field or something)

> I think there are errors in the existing implementation with how this is handled and converted. ObservationValue was implemented at the app level at some point to make it easy to separate the properties that are user-writeable from the read-only properties (that are managed by the server). It looks like mapeo-schema and the implementation in mobile have drifted out of sync. We need to look at how collected data will have been saved and modify the schema/type to reflect that.

Yeah this refactor has surfaced some bigger issues so definitely need to think + work through this 😄

src/frontend/lib/mapeo-schema.d.ts
achou11: yeah there's a misalignment with the definition from `mapeo-schema` and what mobile has. your suggestion makes sense – just have to fully figure out the shape that mobile is using 😅

src/frontend/api.ts
achou11: wait my explanation makes less sense lol. what you described is exactly the reasoning actually 😆

src/backend/constants.js
achou11: I updated this file to use JSDoc via https://github.com/digidem/mapeo-mobile/pull/723/commits/d7757da315fc26fadeef1d114cd34385b8416e2d

Not sure if I've done it idiomatically so feel free to suggest adjustments

src/frontend/context/ObservationsContext.tsx
ErikSin: So apparantly I did this....but I really dont remember doing this, but Im looking more into it now

ErikSin: Oh yeah! I see what is happening. The API call is attempting to map `ServerObservation` onto on object of type `Observation`:
image

ErikSin: So should `convertFromServer` actually be returning 2 objects, `Observation` and `Observation Value`?

src/frontend/@types/mapeo-schema.d.ts
```diff
@@ -0,0 +1,147 @@
+// Type definitions for mapeo-schema 3.0.0
+// Project: https://github.com/digidem/mapeo-schema/
+// Definitions by: Andrew Chou <https://github.com/achou11>
+
+declare module "mapeo-schema" {
```
achou11, line 13: This file should be removed when updating to the next major version of `mapeo-schema` (see https://github.com/digidem/mapeo-schema/pull/13)

achou11:
Creating separate PR to make it a little easier to parse this

PR #724, opened 2021-08-27
Practice Mode

ErikSin:
# Practice Mode

When user is using Mapeo Default Config, an orange practice mode banner will surround all the screens.

The text "Practice Mode" will not appear when the user is in the Project Configuration Screen.

### To do
* Determine if user is in practice mode based on project key NOT project name (Ask @gmaclennan how to do this)

image

image


src/frontend/screens/JoinRequestModal.tsx
```diff
@@ -107,7 +107,7 @@ export const JoinRequestModal: NavigationStackScreenComponent<{
           ]}
           icon={
             <Circle radius={40}>
-              <MaterialIcon color="#000033" name="person-add" size={30} />
+              <MaterialIcon color="#000033" name="highlight" size={30} />
```
achou11: is this change supposed to be here? this component refers to [this](https://www.figma.com/file/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=46972%3A741)

src/frontend/context/ConfigContext.tsx
```diff
@@ -67,19 +67,23 @@ export type State = {
   fields: FieldsMap;
   metadata: Metadata;
   messages: Messages;
-  status: Status;
+  status?: Status;
```
achou11: nit but probably shouldn't be optional?

src/frontend/context/ConfigContext.tsx
```diff
@@ -67,19 +67,23 @@ export type State = {
...
+  practiceModeBar: boolean;
+  practiceModeBarToggle: React.Dispatch<React.SetStateAction<boolean>> | null;
```
achou11: curious if the practice mode should be tied to settings instead of the config. guess it's derived from the config but not sure what that means in terms of where it should live

src/frontend/screens/Settings/ProjectConfig.tsx
```diff
@@ -86,6 +86,12 @@ const ProjectConfig = () => {
     ]);
   }, [didError, t]);
 
+  React.useEffect(() => {
+    config.practiceModeBarToggle!(false);
```
achou11: is the `!` supposed to be there?

src/frontend/sharedComponents/PracticeMode.tsx
```diff
@@ -0,0 +1,68 @@
...
+  title: {
+    id: "sharedComponents.PracticeMode.title",
+    defaultMessage: "Practice Mode",
+    description: "Indicated to user that they are on practice mode",
+  },
+});
+
+interface PracticeModeProps {
+  children: React.ReactNode;
+}
```
achou11: nit but can reduce this to use/extend `React.PropsWithChildren`

src/frontend/screens/Settings/ProjectConfig.tsx
```diff
@@ -86,6 +86,12 @@ const ProjectConfig = () => {
...
+
+    return () => config.practiceModeBarToggle!(true);
+  }, []);
```
achou11: think it would make more sense for this toggle to be handled at the app level, instead of having individual screens handle it. ~will comment in that component to explain~ see [comment](https://github.com/digidem/mapeo-mobile/pull/724/files#r698557849)

src/frontend/AppContainerWrapper.tsx
```diff
@@ -154,17 +155,19 @@ const AppContainerWrapper = () => {
...
-      onNavigationStateChange={onNavStateChange}
-      persistNavigationState={persistNavigationState}
-      ref={nav => {
-        if (nav) {
-          navRef.current = nav;
-        }
-      }}
-      uriPrefix={URI_PREFIX}
-    />
+    <PracticeMode>
```
achou11: I would update this component to take a prop e.g. `hideBar` that's determined based on the nav state. then we can set some app-level state variable that calculates the passed value whenever the navigation state changes i.e.

```tsx
const [hidePracticeModeBar, setHidePracticeModeBar] = React.useState(() => {
  // do an initial calculation if necessary
  ...
}

const onNavigationStateChange = (previousState, currentState) => {
  ...
  // Figure out if the screen is a screen where we want to hide practice mode UI
  const hideBar = someHelperFunction(previousState, currentState)
  setHidePracticeModeBar(hideBar)
  ... 
}

return (
  {/* Use the prop in this component to conditionally display the bar */}
  <PracticeMode hideBar={hidePracticeModeBar}>
    <AppContainer onNavigationStateChange={onNavigationStateChange}  />
  </PracticeMode>
)
```

With this, you can remove the added `practiceModeBar` and `practiceModeBarToggle` fields in the config context as you currently have it and not worry about having to update individual components/screens to call something when they mount to handle this.

src/frontend/context/ConfigContext.tsx
achou11: realized this is specific to the bar at the bottom and not practice mode as a whole. see [this comment](https://github.com/digidem/mapeo-mobile/pull/724/files#r698557849) for suggestion

src/frontend/screens/Settings/ProjectConfig.tsx
achou11: ah nvm i see it's using [this operator](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-)

src/frontend/sharedComponents/PracticeMode.tsx
```diff
@@ -0,0 +1,68 @@
...
+  //TODO change how we determine whether we are in practice mode or not
+  if (config.metadata.name === "mapeo-default-settings") {
+    return (
+      <View style={styles.container}>
+        {children}
+
+        {config.practiceModeBar && (
+          <View style={styles.bottomBar}>
+            {/* Placeholder icon */}
+            <MaterialIcons name="highlight" color={WHITE} size={30} />
```
achou11: I've been using this icon since it's somewhat close (you'll have to apply a transform to rotate it)

image

src/frontend/context/ConfigContext.tsx
achou11: ah nvm i see you added the annotation to the default state, which doesn't have the field defined

src/frontend/screens/JoinRequestModal.tsx
ErikSin: Oh, no I was just messing around with something and I forgot to change it back.

src/frontend/sharedComponents/PracticeMode.tsx
ErikSin: Hmm, Im not sure I know the difference

achou11: there isn't much but `React.PropsWithChildren` is basically the definition you have there but built into React's type defs

src/frontend/AppContainerWrapper.tsx
ErikSin: yeah, this architecture is probably more maintainable than my current architecture

src/frontend/sharedComponents/PracticeMode.tsx
ErikSin: Right, so this needs a typing, should it be `any`?
Like the child here can be anything as its simply a ui wrapper, no logic involved

achou11: yeah in that case you can do something like:

```tsx
const SomeComponent = ({ children }: React.PropsWithChildren<{}>) => { ... } 
```

Again, this is a nit so not really something you have to change

ErikSin: haha well if anything has come from this, I kinda understand PropsWithChildren now (React specfic typing is brutal) vs ReactNode. Arguably I think ReactNode is better for this use case due to how general the child can be!

achou11: Decent reference for understanding the various approaches to the same goal: https://www.newline.co/@bespoyasov/how-to-define-props-with-children-in-react-typescript-app--56bd18be

src/frontend/sharedComponents/PracticeMode.tsx
ErikSin: did you know this existed: https://oblador.github.io/react-native-vector-icons/

Sabella just introduced it to me!

achou11: haha yeah that's what i've been referencing

achou11: ah wait nvm i've been using: https://react-icons.github.io/react-icons/icons?name=md

ErikSin: Wow! Ive literally been going into each library and looking at icons that way.....

src/frontend/AppContainerWrapper.tsx
```diff
@@ -22,15 +30,6 @@ const log = debug("mapeo:App");
...
 
-const EDITING_SCREEN_NAMES = [
-  "AddPhoto",
-  "Camera",
-  "CategoryChooser",
-  "ManualGpsScreen",
-  "ObservationDetails",
-  "ObservationEdit",
-];
-
```
ErikSin, line 33: Moved this to the constants file!

src/frontend/AppContainerWrapper.tsx
```diff
@@ -100,6 +102,11 @@ const AppContainerWrapper = () => {
     const previousRouteName = getRouteName(previousState);
     const currentRouteName = getRouteName(currentState);
 
+    //sets practice bar on of off depedning if route name is included in NO_PRACTICE_BAR array
```
achou11: Small typo:

```suggestion
    // sets practice bar on of off depending if route name is included in NO_PRACTICE_BAR array
```

src/frontend/AppContainerWrapper.tsx
```diff
@@ -100,6 +102,11 @@ const AppContainerWrapper = () => {
...
+    setHidePracticeBar(
+      currentRouteName !== null && NO_PRACTICE_BAR.includes(currentRouteName)
```
achou11: optional but could extract this into a helper function in this file since you do an identical thing in the initializer for `hidePracticeBar`. Something like:

```ts
const shouldHidePracticeBar = (route: string | null) => route !== null && NO_PRACTICE_BAR.includes(route)
```

achou11 (approved):
Couple small additional comments but otherwise think good to go 👍

achou11:
hm maybe a flaky test? https://github.com/digidem/mapeo-mobile/pull/724/checks?check_run_id=3508694906

might be worth re-running the tests and seeing if that's the case

PR #730, opened 2021-09-01
chore: Clean up config context

achou11:
Notes:
- More of a take-it-or-leave-it PR, but spurred from https://github.com/digidem/mapeo-mobile/pull/723#discussion_r697847731, with the intention of eliminating the less-than-ideal `reloadToken` implementation
- Converts implementation to use async/await syntax


src/frontend/context/ConfigContext.tsx
```diff
@@ -95,81 +95,87 @@ const defaultContext: ConfigContextType = [
...
 
-type Props = {
-  children: React.ReactNode;
-};
-
-export const ConfigProvider = ({ children }: Props) => {
+export const ConfigProvider = ({ children }: React.PropsWithChildren<{}>) => {
+  const mountedRef = React.useRef(false);
   const [config, setConfig] = React.useState(defaultConfig);
   const [status, setStatus] = React.useState<Status>("idle");
```
gmaclennan: This would be a change from current behaviour, but I think it makes more sense for this initial state to be `loading`, since on first mount it is always going to go into `loading` state after the first render. This would avoid a "flash" of `idle` state for components consuming this component.

src/frontend/context/ConfigContext.tsx
```diff
@@ -95,81 +95,87 @@ const defaultContext: ConfigContextType = [
 
 const ConfigContext = React.createContext<ConfigContextType>(defaultContext);
 
-type Props = {
-  children: React.ReactNode;
-};
-
-export const ConfigProvider = ({ children }: Props) => {
+export const ConfigProvider = ({ children }: React.PropsWithChildren<{}>) => {
+  const mountedRef = React.useRef(false);
```
gmaclennan: We use this is a few places, perhaps handy to extract to a custom hook e.g. https://www.npmjs.com/package/ismounted or our own - that we avoid the possibility of forgetting to set this in a `useEffect`. With putting it into a `useEffect` that does other things (e.g. the initial `loadConfig` effect below) there is always the risk that we refactor that resulting in the useEffect running after first mount, which would break this `isMounted`

src/frontend/context/ConfigContext.tsx
```diff
@@ -95,81 +95,87 @@ const defaultContext: ConfigContextType = [
...
-      })
-      .catch(err => {
-        bugsnag.notify(err, report => {
-          report.severity = "error";
-        });
-        log("Error loading presets and fields", err);
-        if (didCancel) return; // if component was unmounted, don't set state
-        setStatus("error");
-      });
+    mountedRef.current = true;
```
gmaclennan: I think it's safest to set the mountedRef in a useEffect on its own, rather than with other functions and other dependencies. That way we can be sure that it _only_ runs on first load / unmount. Otherwise there is a risk that during a refactor we forget this restriction and add something to this useEffect that makes it run multiple times.

src/frontend/context/ConfigContext.tsx
```diff
@@ -95,81 +95,87 @@ const defaultContext: ConfigContextType = [
...
+        report.severity = "error";
+      });
+
+      log("Error loading presets and fields", err);
+
+      if (!mountedRef.current) return; // if component was unmounted, don't set state
+
+      setStatus("error");
+    }
+  }, [intl.locale]);
```
gmaclennan: There is a behaviour change and potential race condition introduced by this refactor. If the user changes the locale during the `await Promise.all()` call, then they could end up with the config showing in the previous locale. In the previous implementation, if `intl.locale` changed, then the config would be reloaded. It is highly unlikely this would ever happen, but I think best to avoid that bug.

I think this can be solved by changing how `api.getConfigMessages()` works, to return messages for all locales, so you can remove the `intl.locale` dependency for this useCallback. Then, in the `mergedMessages` memo below, get `config.messages[intl.locale]` and add `intl.local` to that useMemo() so that we get the right messages when the locale changes.

gmaclennan (changes requested):
Thanks for this cleanup, I think it makes this easier to understand. There is a small potential race condition though (with virtually zero chance of happening!) that would be good to clean up before merging.

src/frontend/context/ConfigContext.tsx
achou11: Could we solve this by making the locale an argument for `loadConfig` and then calling it when the locale changes via a use effect?

```ts
// This now takes the locale as an argument
const loadConfig = React.useCallback(async (locale: string) => {
  // ...
}

// Only *possible* issue is that it introduces a similar problem for the `replaceConfig` function
// I think this would be less problematic since it's called as a direct result of a user-initiated action
const replace = React.useCallback(
  async fileUri => {
    try {
      setStatus("loading");
      await api.replaceConfig(fileUri);
      await loadConfig(intl.locale);
    } catch (err) {
      // ...
    }
  },
  [intl.locale, loadConfig]
);

// Now we reload the config each time the locale changes too
React.useEffect(() => {
  loadConfig(intl.locale)
}, [loadConfig, intl.locale])

// need to add intl.locale as a dep here too
const contextValue: ConfigContextType = React.useMemo(
  () => [
    { ...config, status },
    { reload: () => loadConfig(intl.locale), replace },
  ],
  [config, status, loadConfig, replace, intl.locale]
);
```

src/frontend/context/ConfigContext.tsx
achou11: yeah turns out we had a custom hook already defined for this so updated it and used here (and in other areas that did a similar thing)

src/frontend/context/ConfigContext.tsx
achou11: yeah turns out we had a custom hook already defined for this so updated it and used here (and in other areas that did a similar thing)

src/frontend/context/ConfigContext.tsx
achou11: Implemented above change via https://github.com/digidem/mapeo-mobile/pull/730/commits/e360ef190730a9db3f6d4187406a5d9af44e963d

ErikSin (approved):
I just reviewed the code, don't really know how to test this in the ui....

PR #500, opened 2020-12-01
[DO NOT MERGE] chore: Update to latest nodejs-mobile

gmaclennan:

PR #739, opened 2021-09-15
chore: Configure to build with EAS [WIP]

gmaclennan:
Add configuration for building with [EAS Build](https://docs.expo.dev/build/introduction/) and for building a [custom dev client](https://docs.expo.dev/clients/introduction/) that allows mobile development without requiring Android Studio or XCode to be installed.

I was able to get EAS build working correctly. The infrastructure is slower than Cirrus CI, but faster than Github. The nice thing about it is it can also run locally `eas build --platform android --local`, so we could run it on Cirrus CI if we wanted to. I like the idea of configuring the build process once and being able to run anywhere.

I had limited success with the custom development client. I was able to build it locally, but it did not work when building with EAS build. I think this is because it required an upgrade to Android Gradle Plugin, and that seems to have stopped nodejs-mobile from building native modules. It was only working for me because the native modules were already built and cached in my build folder. It was nice to work with: After installing the dev client on my phone I was able to run `expo start --dev-client` and then work on Mapeo as if I was using expo. We just need to update the dev client whenever we change a native dep or change the backend.

The latest React Native 0.65 release includes updates to Gradle and Android Gradle Plugin. It might be worth updating that first, ensuring everything works, and then rebasing this PR on that. We _may_ need to update nodejs-mobile to 0.6.2 (we are running 0.6.1) but I want to avoid that because our tests show that sync has errors when running on Node v12.19.0 for some reason (see https://github.com/digidem/mapeo-mobile/pull/500).

Next steps:
- Update Mapeo to React Native v0.65 in `develop`
- Update this PR to only add configuration for EAS Build
- Rebase this PR
- Work on Dev client in separate PR


gmaclennan:
@achou11 I think we can re-visit this now, see my next-steps in the comment above.

arky:
+1 
I would really love if this could be merged into develop branch. Alternatively this could go into separate 'experimental builds' branch so we can smooth out the rough edges.

PR #746, opened 2021-09-21
Observation view and Observation edit view autofocus

### LINKED ISSUE #734: Autofocus text fields within Observation Details
**Raised by** : Emilie Reiser

**Description :**

New users struggle to figure out how to enter information in text fields within Observation Details. 

Unlike the description/notes field in ObservationEdit/ObservationView screens, the "placeholder" content for ObservationDetails fields is displayed as a hint rather than a true placeholder.

127199817-1f6b3487-15f3-47dd-b733-d0805daecc1d

Current view when arriving at a Details text field

**Goal :** To avoid confusion, let's autofocus all text inputs within ObservationDetails so that cursor is visible and keyboard is opened by default.
https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/screens/ObservationDetails/TextArea.js

3 points

ErikSin:
# Details:
Previously, the observation view screen and observation edit view screen had placeholder text in the text input forms. This placeholder text was confusing users, as it was ambiguous as to whether the text input was an input or just text.

This PR, auto focuses the text input so that the cursor and keyboard appear when they navigate to these pages, in the hopes of making it obvious that the user is interacting with a text input.

The question.js and textarea.js were converted to tsx files in the process

# Initial issue:
Closes #734

### DIFF
--- a/src/frontend/screens/ObservationDetails/Question.tsx
+++ b/src/frontend/screens/ObservationDetails/Question.tsx
@@ -10,15 +10,14 @@ export type QuestionProps = {
   field: Field;
   value: any;
   onChange: (value: any) => any;
-  autoFocus?: boolean;
 };
 
-const Question = ({ field, autoFocus, ...other }: QuestionProps) => {
+const Question = ({ field, ...other }: QuestionProps) => {
   if (field.type === "select_one" && Array.isArray(field.options)) {
     return <SelectOne {...other} field={field} />;
   } else if (field.type === "select_multiple" && Array.isArray(field.options)) {
     return <SelectMultiple {...other} field={field} />;
-  } else return <TextArea autoFocus={true} {...{ field, ...other }} />;
+  } else return <TextArea {...{ field, ...other }} />;
 };
 
 export default Question;
--- a/src/frontend/screens/ObservationDetails/Question.tsx
+++ b/src/frontend/screens/ObservationDetails/Question.tsx
@@ -1,4 +1,3 @@
-// @flow
 import React from "react";
 
 import SelectOne from "./SelectOne";
@@ -8,19 +7,18 @@ import TextArea from "./TextArea";
 import type { Field } from "../../context/ConfigContext";
 
 export type QuestionProps = {
-  field: Field,
-  value: any,
-  onChange: (value: any) => any,
+  field: Field;
+  value: any;
+  onChange: (value: any) => any;
+  autoFocus?: boolean;
 };
 
-const Question = (props: QuestionProps) => {
-  // This is needed for Flow to understand the type refinement
-  const { field, ...other } = props;
+const Question = ({ field, autoFocus = false, ...other }: QuestionProps) => {
   if (field.type === "select_one" && Array.isArray(field.options)) {
     return <SelectOne {...other} field={field} />;
   } else if (field.type === "select_multiple" && Array.isArray(field.options)) {
     return <SelectMultiple {...other} field={field} />;
-  } else return <TextArea {...props} />;
+  } else return <TextArea {...{ field, autoFocus, ...other }} />;
 };
 
 export default Question;
--- a/src/frontend/screens/ObservationDetails/Question.tsx
+++ b/src/frontend/screens/ObservationDetails/Question.tsx
@@ -13,12 +13,12 @@ export type QuestionProps = {
   autoFocus?: boolean;
 };
 
-const Question = ({ field, autoFocus = false, ...other }: QuestionProps) => {
+const Question = ({ field, autoFocus, ...other }: QuestionProps) => {
   if (field.type === "select_one" && Array.isArray(field.options)) {
     return <SelectOne {...other} field={field} />;
   } else if (field.type === "select_multiple" && Array.isArray(field.options)) {
     return <SelectMultiple {...other} field={field} />;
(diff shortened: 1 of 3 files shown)


src/frontend/screens/ObservationEdit/ObservationEditView.js
```diff
@@ -137,7 +137,7 @@ const DescriptionField = () => {
           placeholderTextColor="silver"
           underlineColorAndroid="transparent"
           multiline
-          autoFocus={false}
+          autoFocus={true}
```
gmaclennan: My understanding of this requirement was that the description / notes field should _not_ autofocus (retaining current behaviour), but the individual questions that appear under "Add details" _should_ autofocus.

src/frontend/screens/ObservationDetails/TextArea.tsx
```diff
@@ -16,6 +15,7 @@ const TextArea = ({ value, field, onChange }: QuestionProps) => (
       multiline
       scrollEnabled={false}
       textContentType="none"
+      autoFocus={autoFocus}
```
gmaclennan: I don't see where this is being set to `true`? Looks like it defaults to `false`

src/frontend/screens/ObservationDetails/Question.tsx
```diff
@@ -8,19 +7,18 @@ import TextArea from "./TextArea";
...
+  field: Field;
+  value: any;
+  onChange: (value: any) => any;
+  autoFocus?: boolean;
 };
 
-const Question = (props: QuestionProps) => {
-  // This is needed for Flow to understand the type refinement
-  const { field, ...other } = props;
+const Question = ({ field, autoFocus = false, ...other }: QuestionProps) => {
```
gmaclennan: Minor niggle, not sure this makes sense as a prop for this component, given that it has no meaning if the field is a select_one or select_multiple. Is there a need to expose it? I think we want all questions that are text fields to autofocus.

gmaclennan (changes requested):

src/frontend/screens/ObservationDetails/Question.tsx
achou11: +1, probably no need to expose if the requirement is to apply to it to all text field questions

ErikSin: I agree that not necessary to expose it for the select_one and the select_multiple but I did this for simplicity. So the text area take `QuestionProps`. If we want to expose it to ONLY the text field, I would need to create:
> interface TextFieldProp{
> questionProps:QuestionProps,
> autofocus: boolean
>}

The way I have it right now, by making autofocus optional seems a lot simpler to me.

ErikSin:
@gmaclennan I misunderstood which autofocus to turn on. I changed it to be the appropriate field. I left the questionProps with the autofocus, but I am happy to change it if people are against the way I did it.

src/frontend/screens/ObservationDetails/Question.tsx
achou11: I think given Gregor's suggestion, you'd update https://github.com/digidem/mapeo-mobile/blob/9051b94135f8551f422ee2824fb66da82d638f9c/src/frontend/screens/ObservationDetails/TextArea.tsx#L18 so that it's 

```jsx
  <>
    <QuestionLabel field={field} />
    <TextInput
      value={value}
      onChangeText={onChange}
      style={styles.textInput}
      underlineColorAndroid="transparent"
      multiline
      scrollEnabled={false}
      textContentType="none"
      autoFocus // <-- Adjusted line
    />
  </>
```

which wouldn't require you to update the `QuestionProps` in `Question.tsx`

ErikSin:
Upon further discussion with Andrew, I now see where the confusion lies. I exposed the Autofocus because I assumed that the `TextArea` component was being used in a few different places. With a search, I see that its actually only being used for the observations. So I will just set it to true in the component, and NOT expose it since we don't need different behaviour anywhere.

gmaclennan (approved):

src/frontend/screens/ObservationDetails/Question.tsx
```diff
@@ -8,19 +7,17 @@ import TextArea from "./TextArea";
 import type { Field } from "../../context/ConfigContext";
 
 export type QuestionProps = {
-  field: Field,
-  value: any,
-  onChange: (value: any) => any,
+  field: Field;
+  value: any;
+  onChange: (value: any) => any;
```
achou11:
```suggestion
  onChange: (value: any) => void;
```

achou11: ideally would fix up this `QuestionProps` type to be more specific about what `value` is (e.g. `string`, `number`, etc) but more of a nit than anything

src/frontend/screens/ObservationDetails/TextArea.tsx
```diff
@@ -16,6 +15,7 @@ const TextArea = ({ value, field, onChange }: QuestionProps) => (
       multiline
       scrollEnabled={false}
       textContentType="none"
+      autoFocus={true}
```
achou11: nit but the assignment here isn't necessary i.e.

```suggestion
      autoFocus
```

achou11 (approved):
left a couple of non-blocking comments

PR #749, opened 2021-09-22
chore: Convert shared components directory to TypeScript

achou11:
There's probably a better way to present this PR to make it easier to review and less risky (maybe in smaller batches).

Notes:
- Upgrades `react-native-progress` to new major version (`5.0.0`)
- Creates a `/frontend/@types` directory to place module type defs for those that don't have their own
- Some areas to be more mindful of:
  - [`CameraView`](https://github.com/digidem/mapeo-mobile/blob/ac/ts-conversion-shared-components/src/frontend/sharedComponents/CameraView.tsx) went through a lot of changes. Decently confident that it works as expected, but may be worth taking a closer look at
  - Still unsure about how well-ported the [`MapView`](https://github.com/digidem/mapeo-mobile/tree/ac/ts-conversion-shared-components/src/frontend/sharedComponents/MapView) component is. I converted it from a class component that had some `shouldComponentUpdate` optimizations to a functional one that attempts to reimplement those optimizations. Think it's pretty close if not identical but a little tricky to test confidently


src/frontend/hooks/useIsFullyFocused.ts
```diff
@@ -0,0 +1,51 @@
...
+import { NavigationEventSubscription } from "react-navigation";
+import { useNavigation } from "react-navigation-hooks";
+
+/**
+ * Similar to `react-navigation/core`'s [`useIsFocused`](https://reactnavigation.org/docs/use-is-focused/)
+ * but considers the screen to be focused only when:
+ * 1) the screen animation finishes (as opposed to when the animation starts)
+ * 2) the app is not in the background
+ */
+export const useIsFullyFocused = () => {
```
achou11, line 12: extracted from https://github.com/digidem/mapeo-mobile/blob/5cfc723e1faa6f19a18ee76326b304a103e1ca5b/src/frontend/hooks/useIsFullyFocused.ts

gmaclennan:
I didn't get to review this today, but will do on Monday. The main thing I want to check is the MapView component. I can hopefully remember the optimizations in there and the reasons for them, and perhaps we will be able to remove a few. I think some stuff in there was a workaround for pre-release versions of the react mapbox library that is now fixed upstream.

achou11:
> I didn't get to review this today, but will do on Monday. The main thing I want to check is the MapView component. I can hopefully remember the optimizations in there and the reasons for them, and perhaps we will be able to remove a few. I think some stuff in there was a workaround for pre-release versions of the react mapbox library that is now fixed upstream.

Sounds good! I think the main optimization was centered around the user's location coordinates that are received as a prop via the `LocationContext`. The location is used by the Mapbox Camera and according to [your comment](https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/sharedComponents/MapView.js#L168-L173), it can cause CPU usage to spike since the location changes relatively frequently. The primary optimization seemed to focus on only rerendering the MapView if the distance traveled since the last cached user location (also calculated from that location prop) exceeded the minimum distance we defined.

gmaclennan:
Oh great, I left a code comment! How helpful for me now 😀

ErikSin:
This is too big of a PR....breaking it up into smaller PR's doesnt seem useful though. Perhaps it would make sense to break up this PR, and merge the smaller Pr's in between unrelated PRs. im thinking that way if there is a bug that gets introduced we can determine where the bug was introduced? I'm just spitballing ideas and I am open to other suggestions

achou11:
Closing this since it's stale at this point. Still helpful for reference if needed though

PR #755, opened 2021-09-30
chore: Add Husky to project

achou11:
Notes:
- Similar to https://github.com/digidem/mapeo-desktop/pull/534 where a pre-commit hook is added to format relevant changed files and the `extract-messages` is run
  - Adds `pretty quick` (for formatting)


gmaclennan:
Just wondering if lint-staged and pretty-quick --staged might conflict with each other? They both do the same thing don't they? (i.e. just act on staged files) but pretty-quick I think has its own logic for this.

achou11:
> Just wondering if lint-staged and pretty-quick --staged might conflict with each other? They both do the same thing don't they? (i.e. just act on staged files) but pretty-quick I think has its own logic for this.

Yeah i was somewhat confused because I initially tried with just `pretty-quick` (without the `--staged`) but that didn't actually commit the changes. Had to actually add the flag to have it commit changes.

Alternatively, we could forgo using lint-staged if we're thinking that formatting is the only thing we really want to do with the staged files 🤷‍♂️ would kind of differ from the desktop PR but i don't have strong attachments to it

ErikSin:
> > Just wondering if lint-staged and pretty-quick --staged might conflict with each other? They both do the same thing don't they? (i.e. just act on staged files) but pretty-quick I think has its own logic for this.
> 
> Yeah i was somewhat confused because I initially tried with just `pretty-quick` (without the `--staged`) but that didn't actually commit the changes. Had to actually add the flag to have it commit changes.
> 
> Alternatively, we could forgo using lint-staged if we're thinking that formatting is the only thing we really want to do with the staged files 🤷‍♂️ would kind of differ from the desktop PR but i don't have strong attachments to it

I think having both is unnecessary And pretty-quick is doing what we want it to do already. I think for ease, lets get rid of `lint-stage`

ErikSin (changes requested):
Lets get rid of lint-stages

ErikSin (approved):

ErikSin:
Feel free to merge after rebase

PR #756, opened 2021-09-30
chore: Remove NDK 23 in CI when building

achou11:
Looks like GitHub updated the environment so NDK 23 is included (see [example failure](https://github.com/digidem/mapeo-mobile/runs/3757652905#step:12:1206)), which we need to remove prior to building the app.


achou11:
Closing in favor of #757

PR #757, opened 2021-09-30
chore: Patch nodejs-mobile to specify NDK version

gmaclennan:
Fixes failed builds if a later version of NDK is installed.
nodejs-mobile was not specifying which NDK version to use,
which caused it to use the latest installed version

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -63,9 +63,6 @@ jobs:
         with:
           java-version: "openjdk8"
           architecture: "x64"
-      # Our build will otherwise use NDK 22, which does not work.
-      - name: Delete newer NDK version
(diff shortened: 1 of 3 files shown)


achou11 (approved):
nice!

gmaclennan:
Looks like the build on Github is passing e2e tests (the issue before with NDK v22 was it built without error, but then the builds failed - the issue with NDK v23 is that they have [removed "GNU binutils"](https://developer.android.com/ndk/downloads/revision_history) which nodejs-mobile seems to depend on to build modules. I have no real idea about what these are!)

We should make a PR to `nodejs-mobile-react-native` to contribute back our changes: 

- Specify NDK version
- Use prebuilt toolchain

I can do it one day when I have free time.

PR #846, opened 2021-10-18
feat: Add Create Project flow

### LINKED ISSUE #841: (Front) Create a project
**Description :**
User is in Practice Mode
User goes to the settings/Project Configuration
User clicks on **Create a project**

**Screen is:**
(back button + create project )
Enter the name for the project
(create project) 

*If user attemps to create a project without adding a Project Name or Naming the device
Display error message as designed

*note :  we are removing the option of Advance project config

**Design**
[Here
](https://www.figma.com/file/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=46744%3A804)
**Points**

Points : 5 points

achou11:
Closes #841 

Notes:
- Only includes project and device name fields
- Had to fix the practice mode UI logic because it would cause a whole screen re-render when changing to routes that don't display it (i.e. Create Project screen)
- None of the backend supports this right now

Preview:

Kapture 2021-10-18 at 13 30 49


src/frontend/sharedComponents/PracticeMode.tsx
```diff
@@ -29,38 +29,37 @@ export const PracticeMode = ({
...
+      {showPracticeModeUi && !hideBar && (
+        <View style={styles.bottomBar}>
+          <MaterialCommunityIcons name="lightbulb-on" color={WHITE} size={28} />
+          <Text style={styles.text}>
+            <FormattedMessage {...m.title} />
+          </Text>
+        </View>
+      )}
+    </View>
+  );
```
achou11, line 52: Instead of relying on conditional returns to hide practice mode UI, we conditionally style a wrapping container. This prevents massive re-renders with the relatively minor trade-off of keeping some practice mode nodes present

ErikSin: Nice

src/frontend/screens/CreateProject/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+    validateDeviceName,
+  ] = useInputFieldValue();
+
+  const updateProjectName = React.useCallback(value => {
+    setProjectName(value);
+  }, []);
+
+  const updateDeviceName = React.useCallback(value => {
+    setDeviceName(value);
+  }, []);
```
ErikSin: Are these useCallbacks necessary?

ErikSin (approved):
Nice job

src/frontend/screens/CreateProject/index.tsx
achou11: probably not. gotten into the nasty habit of defaulting to that but it's not necessary in this case. will fix 👍

PR #848, opened 2021-10-20
fix: fix crash on Android 10 [WIP]

gmaclennan:
App currently crashes on Android 10. Hopefully this will be fixed by updating to latest React Native. Kicking this off with updating e2e tests to run on multiple versions of Android, to catch issues like this in the future. Will see how long the tests take, we may not be able to run on all platforms for every commit.

- [x] Update e2e tests to run on Android 8.1, 9.0, 10.0, 11.0
- [ ] Update to RN 0.66
- [ ] Apply patches to nodejs-mobile

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -67,8 +67,20 @@ android_test_task:
   artifact_cache:
     folder: android/app/build/outputs/apk
     fingerprint_script: echo $CIRRUS_BUILD_ID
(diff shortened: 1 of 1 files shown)


achou11:
@gmaclennan is this still relevant? I haven't tested on an Android 10 device but we've upgraded to React Native 0.66 already

Issue #301, opened 2020-02-05
java.lang.UnsatisfiedLinkError in MainActivity


gmaclennan:
## Error in Mapeo Mobile

**java.lang.UnsatisfiedLinkError** in **MainActivity**
couldn't find DSO to load: libv8executor.so caused by: couldn't find DSO to load: libfb.so caused by: couldn't find DSO to load: libc++_shared.so caused by: dlopen failed: "/data/data/com.mapeo/lib-main/libc++_shared.so" is 64-bit instead of 32-bit

[View on Bugsnag](https://app.bugsnag.com/digital-democracy/mapeo-mobile/errors/5e3a43baf82ef5001be6fbca?event_id=5e3a43ba0055f36d670a0000&i=gh&m=ci)

## Stacktrace

    SoLoader.java:738 - com.facebook.soloader.SoLoader.doLoadLibraryBySoName

[View full stacktrace](https://app.bugsnag.com/digital-democracy/mapeo-mobile/errors/5e3a43baf82ef5001be6fbca?event_id=5e3a43ba0055f36d670a0000&i=gh&m=ci)

*Created automatically via Bugsnag*

gmaclennan:
This is happening a lot, but only on one specific device type with manufacturer "OPPO". The device seems to be configured in Thai so this is probably related to @arky's trainings. According to https://github.com/facebook/react-native/issues/25490 this [might be a bug](https://github.com/facebook/react-native/issues/25490#issuecomment-521161256) in the "FunTouch OS" installer which can be fixed by updating the phone.

This also seems to be fixed with a workaround in https://github.com/facebook/SoLoader/pull/45 which requires us to force use of the latest SoLoader as described in https://github.com/facebook/react-native/issues/25490#issuecomment-534570333

Proposed resolution: Add resolution strategy to Mapeo `build.gradle`:

```java
configurations.all {
     resolutionStrategy {
       force "com.facebook.soloader:soloader:0.8.0"
     }
}
```

arky:
@gmaclennan Interesting! Oppo is quite popular brand in Southeast Asia. There were couple of these devices at Thailand trainings. 

It is really difficult to get folks to update the device OS. :(

arky:
@gmaclennan I quickly checked on Oppo F7 device, it runs mapeo without any issues. Will check if I could get my hands on other variants of this device.
[...]

PR #867, opened 2021-11-10
Upgrade React Native to 0.66

### LINKED ISSUE #840: Migrate to React Native 0.66
Note from Gregor:

React Native 0.66 is out, so we can maybe migrate directly to that: [](https://reactnative.dev/blog/2021/10/01/version-066?ck_subscriber_id=887776426). Related: When I had tried to upgrade to RN 0.65 before, nodejs-mobile was not building correctly. In my experiments with building Realm I figured out why.

Points: 13

achou11:
Closes #840 

### Open questions so far:
- I think `react-native-fs` may cause us problems after the bump to the sdk version we target (30). Several issues on the repo questioning whether it's actively maintained and there seem to be many issues with it on Android 11. More importantly, based on https://github.com/itinance/react-native-fs/issues/998 seems like submission won't even be allowed to the Play Store unless we patch something? **All in all, I think we may need to look into a better maintained alternative**
  - perhaps somewhat reassuring, but Manyverse also uses the dependency and I'm guessing they don't have any issues with submitting to the Play Store, so maybe we'll be okay for now. Could also possibly depend on the apis being used in the app code, but I guess we'll address the problem if we run into it 🤷‍♂️ 
- Do we want to enable Hermes? I briefly tried it and nothing seemed noticeably broken as far as I can tell. Some benefits of doing so:
  - apparent performance and size advantages over existing solutions
  - take advantage of the [Intl API work](https://reactnative.dev/blog/2021/08/17/version-065#whats-new-in-hermes-08) that was introduced in 0.65
  - allows us to use some other deps/more recent versions e.g. react-native-reanimated v2 (which is not really used heavily in our app)
  - remove configuration specific to incorporating react-native-v8 into the project
  
### Known issues:
- splash screen doesn't _always_ show correctly. there are various issues with `react-native-splash-screens` and seems like it's not maintained well enough for it to be resolved. general consensus has been to use `react-native-bootsplash` instead, which I have mostly working on a local branch. The only thing that needs to be addressed is getting the ICCA assets generated (I basically just need the Mapeo logo with the ICCA copy to generate the necessary assets) cc @gmaclennan 

### Remaining work:
- potentially need to update Detox
- fix splashscreen

### Dependency changes:
- Added
  - Primary:
    - `v8-android`
  - Dev (not including `@types`):
    - `metro-config`
    - `react-native-codegen`
- Upgraded
  - Primary:
    - `@react-native-mapbox/maps` (8.0.0 -> 8.5.0)
    - `nodejs-mobile-react-native` (0.6.1 -> 0.6.3)
    - `react` (16.13.1 -> 17.0.2)
    -  `react-native` (0.63.4 -> 0.66.4)
    - `react-native-device-info` (5.6.3 -> 8.4.8)
    - `react-native-v8` (0.63.2-patch.0 -> 0.66.3-patch.1)
  - Dev (not including `@types`):
    - `@babel/core` (7.11.1 -> 7.12.17)
    - `@babel/runtime` (7.11.2 -> 7.12.18)
    - `@react-native-community/cli` (4.12.0 -> 6.2.0)
    - `babel-jest` (25.3.0 -> 26.6.3)
    - `eslint` (7.6.0 -> 7.14.0)
    - `flow-bin` (0.122.0 -> 0.149.0)
    - `metro-react-native-babel-preset` (0.61.0 -> 0.66.2)
    - `react-dom` (16.13.1 -> 17.0.2)
    - `react-test-renderer` (16.13.1 -> 17.0.2)
- Removed
  - Primary:
  - Dev (not including `@types`):


android/app/src/main/java/com/mapeo/MainActivity.java
```diff
@@ -10,11 +10,11 @@
 import com.swmansion.gesturehandler.react.RNGestureHandlerEnabledRootView;
 /* End: react-native-guesture-handler */
 
-public class MainActivity extends ReactFragmentActivity {
+public class MainActivity extends ReactActivity {
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         SplashScreen.show(this);  // react-native-splash-screen
-        super.onCreate(savedInstanceState);
+        super.onCreate(null); // react-native-screens
```
gmaclennan: Why the change to `null`? I thought that the super needs to be called with the same argument (`savedInstanceState`).

android/app/build.gradle
```diff
@@ -333,24 +342,18 @@ android {
     }
 }
 
-configurations.all {
-    resolutionStrategy {
-        force "com.facebook.soloader:soloader:0.8.0"
-    }
-}
-
```
gmaclennan, line 359: This was a [fix for an error with OPPO phones](https://github.com/digidem/mapeo-mobile/issues/301). Can you check/confirm via the original issue whether this is now fixed in the latest React Native?

android/app/src/main/res/values/styles.xml
```diff
@@ -2,11 +2,22 @@
...
+
+    <!-- https://stackoverflow.com/a/42533536 -->
+    <style name="ThemeAlert" parent="Theme.AppCompat.DayNight.Dialog.Alert">
+        <!-- Used for the buttons -->
+        <!-- <item name="colorAccent"></item> -->
+        <!-- Used for the title and text -->
+        <item name="android:textColorPrimary">#000000</item>
+        <!-- Used for the background -->
+        <item name="android:background">#ffffff</item>
     </style>
```
gmaclennan, line 21: Does this change how alerts are displayed in the app?

android/build.gradle
```diff
@@ -2,11 +2,11 @@
...
     ext {
-        buildToolsVersion = "29.0.2"
+        buildToolsVersion = "30.0.2"
         // Android 5.0 - requirement for expo unimodules
         minSdkVersion = 21
-        compileSdkVersion = 29
-        targetSdkVersion = 29
+        compileSdkVersion = 30
+        targetSdkVersion = 30
         supportLibVersion = "29.0.0"
```
gmaclennan: Does this (`supportedLibVersion`) need to change? I don't really understand what this property is for.

package.json
```diff
@@ -68,17 +68,17 @@
     "lodash": "^4.17.19",
     "mapeo-offline-map": "^2.0.0",
     "mapeo-schema": "^2.0.2",
-    "nodejs-mobile-react-native": "^0.6.1",
+    "nodejs-mobile-react-native": "^0.6.3",
```
gmaclennan, line 69: I'm a bit worried about this update, which updates the Node version used by the backend to v12.19.0. Our mapeo-core tests currently fail for v12.19.0: https://github.com/digidem/mapeo-core/runs/4518408273?check_suite_focus=true
I understand we might require this update to support Android v11, which means we might need to investigate this mapeo-core error (seems to be related to leveldb, so might be a case of patching that). Can you look into this @achou11 ?

src/backend/index.js
```diff
@@ -128,7 +128,13 @@ rnBridge.app.on("resume", () => {
   // listening (because it wasn't paused)
   // https://github.com/janeasystems/nodejs-mobile/issues/177
   status.startHeartbeat();
-  startServer();
+
+  // (achou11): this check is not ideal but necessary because of how the 'resume' event
+  // is spawned after interacting with RN's permissions prompt on startup.
+  // It seems to be inconsistent based on OS, RN, and/or nodejs-mobile-react-native versions
```
gmaclennan, line 134: Interesting... it still fires again right so it does eventually start?

src/frontend/AppLoading.tsx
```diff
@@ -0,0 +1,85 @@
...
+  }, []);
+
+  React.useEffect(() => {
+    requestPermissions(REQUESTED_PERMISSIONS);
+
+    timeoutId.current = window.setTimeout(() => {
+      SplashScreen.hide();
+      timeoutId.current = undefined;
+      log("hiding splashscreen");
+    }, 1000);
```
gmaclennan: Better to move this into the same useEffect where the timeout is cleared (above) and avoid useRef for timeoutId. Otherewise, as it stands, a timeout could be left uncleared (`requestPermissions` changes, so the useEffect re-runs, adding a second timeout. Better to keep the creation of the timeout and clearing of the timeout in the same useEffect.

src/frontend/AppLoading.tsx
```diff
@@ -0,0 +1,85 @@
...
+ * server is up and running and listening.
+ * On initial app load it waits for the server before dismissing the native
+ * splash screen.
+ * If it doesn't hear a heartbeat message for timeout, it displays a screen to
+ * the user so they know something is wrong.
+ */
+export const AppLoading = ({ children }: React.PropsWithChildren<{}>) => {
+  const { requestPermissions } = usePermissions();
+
+  const [serverStatus, setServerStatus] = React.useState<ServerStatus | null>(
```
gmaclennan: Nitpick: better to initialize this with `Constants.IDLE`, which is the initial state of the server.
Even better: Add a synchronous `api.getServerState()` method, and initialize the state here with that.
Even even better: `useServerState()` hook that wraps the API call, keeps the logic in one place.

src/frontend/api.ts
```diff
@@ -485,20 +501,25 @@ export function Api({ baseUrl, timeout = DEFAULT_TIMEOUT }: ApiParam) {
...
-          nodejs.channel.removeListener("p2p-upgrade::state", onState),
+        // See comment in startServer
+        // @ts-expect-error
+        remove: () => stateSubscription.remove(),
       };
     },
     addP2pUpgradeErrorListener: (
       handler: (error: Error) => void
     ): Subscription => {
-      nodejs.channel.addListener("p2p-upgrade::error", handler);
```
gmaclennan, line 499: I see there was a bug here, glad you fixed it!

src/frontend/sharedComponents/MapView.js
```diff
@@ -190,14 +201,13 @@ class MapView extends React.Component<Props, State> {
     }
   }
 
-  handleObservationPress = (e: {
-    nativeEvent?: {
-      payload?: {
-        properties?: { id: string },
-      },
-    },
+  handleObservationPress = (event: {
```
gmaclennan: I guess this has probably been broken for a while - must have changed in a version of mapbox but we never tested this.

gmaclennan: I think we need to remove the `onPress` handler on the mapview (line 311 I think). We have the onPress handler on the source layer, so we don't want this to be called twice.

gmaclennan (changes requested):
Great work on this @achou11, it is *a lot*! This is partly a result of not updating the deps for a while - I think we need to have a regular schedule of keeping things updated to ensure this doesn't become such a hard task in the future. It looks great to me, just a few comments, some of which require action before we merge.

android/app/src/main/java/com/mapeo/MainActivity.java
achou11: this is from the `react-native-screens` [docs](https://github.com/software-mansion/react-native-screens#android) so generally an uninformed change on my part 😅 can try reverting that change and seeing if anything breaks, since I didn't specifically test for it. Original issue is [here](https://github.com/software-mansion/react-native-screens/issues/17)

android/app/build.gradle
achou11: React Native 0.66.4 is currently using `soloader` `0.10.1` as this point so I think it should be resolved if the fix was included in `0.8.0`?

https://github.com/facebook/react-native/blob/v0.66.4/ReactAndroid/gradle.properties#L16

achou11: for additional context, I needed to remove this to fix a separate build issue i was running into

android/build.gradle
achou11: good question. i don't have a great idea but based on https://developer.android.com/studio/build/gradle-tips#configure-project-wide-properties, maybe it should also be updated

package.json
achou11: ah yeah i forgot about that 🤦‍♂️ but yes i can look into it.

do you know which commit in https://github.com/nodejs-mobile/nodejs-mobile-react-native introduces the new node version being used? wondering if downgrading to `0.6.2` and adding the appropriate patches would suffice  (this assumes that 0.6.3 is the problematic release, moot otherwise)

src/backend/index.js
achou11: yep! i can't fully remember the nature of the issue i was running into, but i just found that on my device and OS, the event was being sent before the server was initiated on app startup, causing an error raised in `startServer`.

src/frontend/api.ts
```diff
@@ -256,12 +256,18 @@ export function Api({ baseUrl, timeout = DEFAULT_TIMEOUT }: ApiParam) {
...
-      ).then(async () => {
+      const serverStartPromise = new Promise<void>(resolve => {
+        // nodejs-mobile-react-native channel extends React Native's EventEmitter
+        //   - https://code.janeasystems.com/nodejs-mobile/react-native/bridge
+        const statusSubscription = nodejs.channel.addListener("status", () => {
+          // addListener returns an EventSubscription with the remove method
+          //   - https://github.com/facebook/react-native/blob/v0.66.3/Libraries/vendor/emitter/_EventEmitter.js#L57
+          //   - https://github.com/facebook/react-native/blob/v0.66.3/Libraries/vendor/emitter/_EventSubscription.js#L39
+          // @ts-expect-error
+          statusSubscription.remove();
```
achou11, line 267: i _think_ the TS typings for nodejs-mobile-react-native need to be updated since the documentation seems to be correct. Maybe worth adding issue/PR for it on the repo as a follow-up

src/frontend/lib/logger.ts
```diff
@@ -0,0 +1,22 @@
+import Bugsnag from "@bugsnag/react-native";
+// import semverPrerelease from "semver/functions/prerelease";
+// import { version } from "../../../package.json";
+
+// const prereleaseComponents = semverPrerelease(version);
+
+Bugsnag.start({
+  /**
+   * TODO(andrew): release stage cannot be set from JS layer and must be done at native layer
```
achou11: highlighting this as this PR still needs to address this issue (unless we're okay with not accounting for prerelease release stages?)

src/frontend/sharedComponents/MapView.js
achou11: > I guess this has probably been broken for a while - must have changed in a version of mapbox but we never tested this.

the previous code wasn't really broken. it was just using a now-deprecated API as of `8.0.0`. Figured it would just make sense to update the implementation to get rid of the deprecation notices

> I think we need to remove the onPress handler on the mapview (line 311 I think). We have the onPress handler on the source layer, so we don't want this to be called twice.

Yeah we can safely remove the handler on the map view. fwiw looks like it doesn't get called twice even if both of them have the handler

tsconfig.json
```diff
@@ -6,13 +6,28 @@
...
     "moduleResolution": "node",
     "strict": true,
     "target": "esnext",
-    "types": ["node", "jest", "detox"]
+    "types": [
+      "node",
+      "jest",
+      "detox"
+    ],
+    "resolveJsonModule": true
```
achou11, line 21: only added line in this file is here

android/app/src/main/res/values/styles.xml
achou11: yeah. i think react native made some internal changes such that some native UI elements have different interfaces based on the theme settings for the device. I'm currently on dark mode so without this change it showed up as this:

image

With this change, the alert looks like this:

image

Found the relevant React Native issue https://github.com/facebook/react-native/issues/31345 and seems like an artifact in the `styles.xml` file is causing the text color inconsistency. Will post a screenshot with that line removed shortly

Curious if you think we should stick to letting the OS theme determine the alert style or if we should override it as i have it here?

achou11: Here's what happens when removing `<item name="android:textColor">#000000</item>`, as suggested in the issue:

Light mode:

image

Dark mode:

image

---

As you can see, the contrast issues are fixed but since that line is removed, the app falls back to using the OS-specified text color for native elements instead of #000000. So in my phone's case, the action buttons are teal. Basically, this line change has wider consequences and I'm not entirely sure if we're okay with it or not (maybe we are for now?). It seems like the alerts are the only UI elements that we currently use that are heavily affected by this, so maybe that's okay?

src/frontend/AppLoading.tsx
achou11: Done, see https://github.com/digidem/mapeo-mobile/pull/867/commits/48dab34a8f17cee801a931bc9c8799546d738449

src/frontend/AppLoading.tsx
achou11: Did the even even better option in https://github.com/digidem/mapeo-mobile/pull/867/commits/2cb8d77a11d663a11cc0ef2c72064e886c16c821

src/frontend/sharedComponents/MapView.js
achou11: Done in https://github.com/digidem/mapeo-mobile/pull/867/commits/e52f7dab9b71e6215b5f564cb46627b985c58930

android/build.gradle
achou11: Updated in https://github.com/digidem/mapeo-mobile/pull/867/commits/9abfda5f10ab777e1b6ff1936bfe80f054524f3f

android/app/src/main/java/com/mapeo/MainActivity.java
achou11: the same change is also mentioned in the current version of the react-navigation docs https://reactnavigation.org/docs/getting-started/#installing-dependencies-into-a-bare-react-native-project

package.json
achou11: ah i see that 0.6.2 introduces the new node version. disregard previous comment then

android/app/src/main/java/com/mapeo/MainActivity.java
gmaclennan: Great, I'm going to assume that this is the way to do it, since [it's also referenced](https://github.com/expo/expo/tree/master/packages/expo-splash-screen#migrate-from-expo-splash-screen--0120) in the older version of `expo-splash-screen` (side-note, perhaps we should switch to expo-splash-screen in the future, since it should be better maintained).

android/app/build.gradle
gmaclennan: Great, let's go ahead with this removed, and make sure we get Arky to test this on an OPPO phone before we do a public release.

android/app/src/main/res/values/styles.xml
gmaclennan: I'm ok with either option (with or without specified text colour). We can check in the release QA if anything else is not looking right.

src/frontend/api.ts
gmaclennan: yes can you do that?

src/frontend/lib/logger.ts
gmaclennan: yes, let me know if you need help with this.

gmaclennan (changes requested):
So two outstanding issues to fix:

1. Ensuring that sync will work if we're using Node v12.19.0 (from nodejs-mobile-react-native@0.63.0)
2. Fixing version numbers in bugsnag

src/frontend/api.ts
achou11: yeah i added an issue there for now: https://github.com/nodejs-mobile/nodejs-mobile-react-native/issues/3

src/frontend/lib/logger.ts
achou11: I probably will honestly. Don't have the greatest familiarity around native android configurations 😅

achou11: Made an attempt to solve this but the confidence level is somewhere between low and middling 😅 . See https://github.com/digidem/mapeo-mobile/pull/867/commits/a873b72d3ac9a202044996e89bb88192e7f18d5d

achou11:
After internal discussion, going to do the following to make this PR less daunting and safer:

1. Extract non-upgrade changes to a separate PR
2. Merge other PR after ensuring changes don't break stuff
3. Rebase this PR with those changes

See #887

android/app/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,7 @@
   <uses-permission android:name="android.permission.INTERNET" />
   <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
   <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+  <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
```
achou11: for context, think this is automatically added via `expo-location` but explicitly added here for clarity

https://github.com/expo/expo/tree/master/packages/expo-location#configure-for-android

.github/workflows/android.yml
```diff
@@ -57,7 +57,9 @@ jobs:
           path: android/build/nodejs-native-assets
           key: ${{ runner.os }}-nodejs-mobile-${{ hashFiles('src/backend/package-lock.json') }}
       - name: Node
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v2
+        with:
+          node-version: "12.16.3"
```
achou11, line 62: the upgrade of `@react-native-community/cli` (which is used to run android build steps) requires Node >=12

gmaclennan: Interesting I had thought we were already using Node 12 for builds. It's set in `.nvmrc`, but I don't think Github actions pay attention to that file. The other two places this matters are in Bitrise and Cirrus CI. Bitrise respects `.nvmrc` because it uses nvm to manage Node (https://github.com/digidem/mapeo-mobile/blob/develop/bitrise.yml#L326). Cirrus CI uses a Docker container that we build ourselves, and that is currently set to `12.x` e.g. Node v12 latest (https://github.com/digidem/docker-android/blob/master/Dockerfile#L16). So I think we should be ok for this change, just wanted to make sure you are aware of all these pieces.

achou11: Certain parts of the build process were using Node 12, but they had `node-version` specified. Without it [version 10 was being installed](https://github.com/digidem/mapeo-mobile/runs/4718848720?check_suite_focus=true#step:6:7) (or used via the cache).

Regardless, the additional context is appreciated 👍

PR #869, opened 2021-11-21
Delete photo

### LINKED ISSUE #866: (Front) Add option to remove a photo from obvs
**Description:**
For first new pictures
Add option to remove a photo from an observation once captured 
There is currently no way to remove a photo that has been added to a new observation (before save) or delete a photo when editing an observation. At the very least, user should be able to discard/remove a captured photo before the observation is saved.

**Solution:**
Pair trash icon with text label "Delete Image". Preference is to have this action in line with back, keeping actions organized together. We also avoid worries about if the button is visible enough when overlaid on a photo, where we cannot control the contrast between button and image.

Design:
Frame 305

Points : 8

ErikSin:
## Delete Observation Photo
User can delete a photo in a draft observation.

Introduced a header which can be toggled on and off, and now includes a delete button.

Delete Button prompts a bottom sheet asking user to confirm that they would like to delete the photo

Kapture 2021-12-05 at 15 18 28

## Wireframes
https://www.notion.so/digidem/Add-option-to-remove-a-photo-from-an-observation-once-captured-603178a8cb2748a08f22943f1f93a710


src/frontend/context/DraftObservationContext.tsx
```diff
@@ -66,32 +65,37 @@ const usePersistedState = createPersistedState(
...
 
 const defaultContext: DraftObservationContextType = [
   {
     photos: [],
     value: null,
     photoPromises: [],
   },
-  () => {},
+  //@ts-ignore
+  null,
```
achou11: think it'd be fine to keep this as `() => {}`, to appease TS

src/frontend/context/DraftObservationContext.tsx
```diff
@@ -66,32 +65,37 @@ const usePersistedState = createPersistedState(
...
 
 export const DraftObservationProvider = ({ children }: Props) => {
   const [
     state,
     status,
     setState,
+    // @ts-ignore We know the first item in the array is correct
   ] = usePersistedState<DraftObservationContextState>(defaultContext[0]);
-  const contextValue = React.useMemo(() => [state, setState], [
+  const contextValue = React.useMemo(() => [state, setState] as const, [
```
achou11: nit, but could avoid the `as const` if you make `DraftObservationContextType` readonly i.e. `type DraftObservationContextType = readonly ...`. not super important though

src/frontend/hooks/useDraftObservation.tsx
```diff
@@ -33,38 +32,40 @@ const PREVIEW_QUALITY = 30;
...
     // Create a new draft observation
     newDraft: (
       id?: string,
       value?: ObservationValue | null,
       capture?: Promise<CapturedPictureMM>
-    ) => void,
-  |}
+    ) => void;
+
+    deletePhoto: (index: number) => void;
```
achou11: is it too difficult/awkward to use some unique identifier on the photo for the argument here? probably not fully necessary but always a little hesitant about using indices for this kind of thing (maybe unreasonably so 😅 )

src/frontend/hooks/useDraftObservation.tsx
```diff
@@ -141,13 +151,12 @@ export default (): UseDraftObservation => {
   const newDraft = useCallback(
     (
       id?: string,
-      value?: ObservationValue | null = { tags: {} },
+      value: ObservationValue | (null | undefined) = { tags: {} },
```
achou11: nit but could keep the `value?` here

src/frontend/screens/PhotosModal/index.tsx
```diff
@@ -1,24 +1,50 @@
...
+    openSheet();
+  }
 
-  const routes = [];
+  function closeModal() {
+    modalIsOpen.current = false;
+    closeSheet();
+  }
+
+  const routes: any[] = [];
```
achou11: based on implementation, think you can type this as `routes: { key: string, uri: string }[]`

src/frontend/screens/PhotosModal/ConfirmDeleteModal.tsx
```diff
@@ -0,0 +1,51 @@
...
+            text: "Cancel",
+          },
+          {
+            variation: "filled",
+            onPress: () => {
+              deletePhoto(photoIndex);
+              closeSheet();
+              navigationProp.pop();
+            },
+            text: "Yes, Delete",
```
achou11: feel like it should be `Yes, delete` but I'm just generally partial to sentence-casing for everything

src/frontend/screens/PhotosModal/index.tsx
```diff
@@ -1,24 +1,50 @@
...
+  const modalIsOpen = useRef<boolean>(false);
+
+  //We don't want to show the header when the modal is open, these functions
+  //check if the modal is open before trying to show header
+  function toggleShowHeader() {
+    if (modalIsOpen.current) {
+      setShowHeader(false);
+      return;
+    }
+    setShowHeader(!showHeader);
```
achou11: nit but I generally prefer to use the functional syntax for "toggle" interactions i.e. `setShowHeader(previous => !previous)`

src/frontend/screens/PhotosModal/index.tsx
```diff
@@ -1,24 +1,50 @@
...
 const PhotosModal = ({ navigation }: { navigation: NavigationProp }) => {
   const observationId = navigation.getParam("observationId");
   const [index, setIndex] = useState(navigation.getParam("photoIndex") || 0);
   const [{ observation }] = useObservation(observationId);
   const [{ photos: draftPhotos }] = useDraftObservation();
+  const [showHeader, setShowHeader] = useState(false);
+  const { sheetRef, openSheet, closeSheet } = useBottomSheetModal({
+    openOnMount: false,
+  });
+  const modalIsOpen = useRef<boolean>(false);
```
achou11: curious, but does changing this to state affect the interactions/UI here?

src/frontend/hooks/useDraftObservation.tsx
ErikSin: Idk, why but this is the error I am getting if I left `value?`
image

ErikSin: Oh yeah, it looks like because we are initializing a value, there is no point of making it optional

src/frontend/screens/PhotosModal/ConfirmDeleteModal.tsx
ErikSin: Its just a placeholder until I here back from Sabella

src/frontend/screens/PhotosModal/index.tsx
ErikSin: Im not sure what you mean. I didn't want to use state here because this check is already linked to the `showheader` state, so there is no reason to attach 2 state hooks to one state. It is a little hard to explain on here, but basically I dont need to rerender the component when the modal is open, i only need to rerender if the modal is open AND the header is showing. So having 2 state hooks would cause an unnecessary re-renders

ErikSin: What might be a better pattern, is to have a boolean in the `useBottomSheetModal` which returns true when the modal is open. But it looks like this is not something we can do in the `@gorhom/bottom-sheet`. So we would have to follow the pattern that I used here

src/frontend/hooks/useDraftObservation.tsx
achou11: oh yeah of course 🤦‍♂️ nvm then what you have there makes sense

src/frontend/hooks/useDraftObservation.tsx
ErikSin: Not unreasonable. I am just pulling the uri, using the index. Kind of an extra step, but it would probably be more robust in the long term!

src/frontend/screens/PhotosModal/index.tsx
ErikSin: Unfortunantly this was an annoying typing that was used by the tabview....So i ended up just extending `Routes` from tabview to include our custom attributes

src/frontend/sharedComponents/PhotoView.tsx
```diff
@@ -6,10 +6,10 @@ import { AlertIcon } from "./icons";
...
 
 type Props = {
-  uri?: string,
-  variant?: "photo" | "loading" | "error",
-  style?: ViewStyleProp,
-  resizeMode?: "cover" | "contain" | "stretch" | "center",
+  uri?: string;
+  variant?: "photo" | "loading" | "error";
+  style?: ViewStyleProp;
+  resizeMode?: "cover" | "contain" | "stretch" | "center";
```
ErikSin, line 12: caused by husky

src/frontend/hooks/useDraftObservation.ts
```diff
@@ -75,6 +76,33 @@ export const useDraftObservation = (): UseDraftObservation => {
...
+
+    const newPhotoPromiseArray = draft.photoPromises.map(async photo => {
+      const resolvedPhoto = await photo;
+      if (resolvedPhoto.originalUri === uri) {
+        const deletedPhoto: Promise<DraftPhoto> = new Promise((res, rej) => {
+          resolvedPhoto.deleted = true;
+          res(resolvedPhoto);
+        });
+        const cancelPhoto = deletedPhoto as CancellablePromise<DraftPhoto>;
+        cancelPhoto.signal = { didCancel: true };
```
ErikSin, line 92: Not really following functional programming here, seemed a little clunky to create so 2 new objects

src/frontend/screens/PhotosModal/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+}
+
+const m = defineMessages({
+  deleteImage: {
+    id: "screens.PhotosModal",
+    defaultMessage: "Delete Image",
+  },
+});
+
+const PhotosModal = ({ navigation }: { navigation: NavigationProp }) => {
```
achou11, line 33: spent some time trying this and i think this shouldn't be a "modal" in the technical sense. Modals inherently introduce navigation functionality by being part of the navigation stack, which doesn't make sense for features like this. A heuristic I use is "would it make sense to ever be able to deeplink to this modal sheet?" In this case, not really. You'll also notice that with the current modal sheet implementation, the background content flickers because there's a transition when the sheet comes up, which especially looks odd on this screen.

There may be some other parts of the app using the modal sheet that also fall into this trap (maybe not as evidently), but i think it'd be worth creating a variation of the bottom sheet that doesn't have any navigation implications. I have a WIP local branch that builds off this one that implements an identical bottom sheet (with some enhancements), so could be something that we could incorporate into this PR of as a follow up, depending on timing

src/frontend/screens/PhotosModal/index.tsx
achou11: > What might be a better pattern, is to have a boolean in the useBottomSheetModal which returns true when the modal is open. But it looks like this is not something we can do in the @gorhom/bottom-sheet. So we would have to follow the pattern that I used here

makes sense! i think at initial implemetation, i omitted that kind of flag in the hook since it wasn't too difficult to roll your own implementation based on the consumer's needs, but probably worth incorporating since it's helpful in most cases. As i mentioned in https://github.com/digidem/mapeo-mobile/pull/869/files#r763153991, i have some work on implementing a more generic bottom sheet component, which has a similar hook that does return a value about its opened state (which we can also easily port over for the BottomSheetModal component)

src/frontend/screens/PhotosModal/ConfirmDeleteModal.tsx
```diff
@@ -0,0 +1,82 @@
...
+  photoIndex,
+  closeSheet,
+  navigationProp,
+}: ModalProps) => {
+  const [{ photos }, { deletePhoto }] = useDraftObservation();
+  const { formatMessage: t } = useIntl();
+
+  function photoDelete() {
+    const photoToDelete = photos[photoIndex];
+    console.log(photoToDelete);
```
achou11: can remove

src/frontend/screens/PhotosModal/index.tsx
ErikSin: Let discuss in team meeting, Im not sure I fully understand your logic. I realize that adding the navigation prop is unnecessary here. I pass that prop to a child, but I can just use the hook. If I do that, I can just avoid that navigation prop completely

achou11: after discussion and rethinking this, it'd probably be enough to update the modal sheet component to just not have the custom background in this specific case (think adding a prop and some conditional code will do the trick). didn't realize this wasn't being connected to the navigation stack so i think my concerns there don't apply.

we probably will have to create the more generic bottom sheet at some point but not essential for the time being

ErikSin: Yeah, also on during that discussion I realized that I pass the navigation prop AND use it, but its not part of the stack so Im confused as to where it is getting the navigation prop. In my understanding, that prop was only passed to stack components, not all components

ErikSin: Nevermind, I realize that the photos Modal which shows the photo as on overlay (not the bottom sheet) is on the stack, and Im passing that navigation prop to the bottom sheet!

src/frontend/screens/PhotosModal/index.tsx
```diff
@@ -139,6 +136,7 @@ const PhotosModal = ({ navigation }: { navigation: NavigationProp }) => {
         renderTabBar={() => null}
       />
       <ConfirmDeleteModal
+        disableBackdrop={true}
```
achou11, line 139:
```suggestion
        disableBackdrop
```

src/frontend/sharedComponents/BottomSheetModal/index.tsx
```diff
@@ -22,16 +22,19 @@ export const useBottomSheetModal = ({
 }) => {
   const initiallyOpenedRef = React.useRef(false);
   const sheetRef = React.useRef<RNBottomSheetModal>(null);
+  const isOpen = React.useRef<boolean>(false);
```
achou11, line 25: i'm thinking that it would make sense for this to be a state variable. might require some fixing up in other parts of the PR but should be straightforward enough using effects (see https://github.com/digidem/mapeo-mobile/pull/869#discussion_r765281021)

src/frontend/screens/PhotosModal/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const observationId = navigation.getParam("observationId");
+  const [index, setIndex] = useState(navigation.getParam("photoIndex") || 0);
+  const [{ observation }] = useObservation(observationId);
+  const [{ photos: draftPhotos }] = useDraftObservation();
+  const [showHeader, setShowHeader] = useState(true);
+  const { sheetRef, openSheet, closeSheet, isOpen } = useBottomSheetModal({
+    openOnMount: false,
+  });
+
+  function toggleShowHeader() {
```
achou11, line 43: think i'd fix this up a bit by doing something along the lines of:

```ts
  const { sheetRef, openSheet, closeSheet, isOpen } = useBottomSheetModal({
    openOnMount: false,
  });

  function toggleShowHeader() {
    setShowHeader(previous => {
      if (isOpened) return false;
      return !previous;
    });
  }

  useEffect(() => {
    if (isOpened) {
      setShowHeader(false);
    }
  }, [isOpened]);
```

where `isOpen` from the hook is React state and not a ref

src/frontend/screens/PhotosModal/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+import { TouchableNativeFeedback } from "../../sharedComponents/Touchables/index.ios";
+
+interface customRoute extends Route {
+  uri?: string;
+  error?: boolean;
+  capturing?: boolean;
+}
+
+const m = defineMessages({
+  deleteImage: {
```
achou11, line 27: should a description be included here? idk if we settled on a convention moving forward to include whenever possible or not

ErikSin: Description are only necessary when the context of the translation is ambiguous. For example, and words or phrases that have more then one meaning

src/frontend/sharedComponents/BottomSheetModal/index.tsx
ErikSin: This will cause unecessary re-renders as a useState Hook. Everytime the state of a useState hook is changed, all the associated components re-render. In this case we dont need it to re-render because open sheet and close sheet is already causing that re-render, we just want to simply encapsulate that state of being open or close.

Lets say we did have a state hook: `[isOpen, setIsOpen] = useState()`
if we called `openSheet` this will cause the bottom sheet to rerender (desired behaviour),
Then we would call setIsOpen(true). This will then cause the bottom sheet to re-render again (which is not necessary, as none of the UI is changing)

That is why I am using a useRef, because it does not cause any rerenders when its state changes

src/frontend/screens/PhotosModal/index.tsx
ErikSin: This seems like a stylisic preference and the way I have it coded feels easier to read due to the explicit conditions that I am pointing out in my if statements.

I do see the utility of the useeffect here, but due to the simple mechanics it does seem like overkill

src/frontend/sharedComponents/BottomSheetModal/index.tsx
achou11: based on how you want use the opened state in this pr, it feels like the re-render caused by the open state changing is desirable (refer to the [other comment](https://github.com/digidem/mapeo-mobile/pull/869/files#r765281021) to understand why).

more tangentially, i'm remembering why i didn't include this in the hook – the BottomSheetModal was built with the assumption that it would always be it's own navigation screen, which is something i based on the design system at its current state. This made having the variable `isOpen` unnecessary (especially since it's easy to just implement from the consumer side, which is what you had before)

src/frontend/screens/PhotosModal/index.tsx
ErikSin: Again, stylistic preference, I personally prefer the explicit `true`

src/frontend/sharedComponents/BottomSheetModal/index.tsx
achou11: either way, fine to keep the current implementation that you have if it works

achou11: think my main point with this is that as a consumer, it'd potentially be nice to able to write effects that are based on the `isOpened` value i.e.

```ts
const { isOpened } = useBottomModalSheet()

React.useEffect(() => {
  // ...
}, [isOpened)

```

not entirely sure, but I don't think the above is possible if `isOpened` is represented as a ref value instead of state, but could be mistaken (would be easy to test but I don't have the energy at the moment 😅)

ErikSin: Okay, I see what you are saying about that rerender. But I think for the reusability of the component it, in most other use cases, we dont want that rerender.

Yeah, that makes sense why it wasnt originally included in the hook. Its also interesting that it is not included out of the box. But probably for the same reason as you stated!

achou11 (approved):

achou11:
might want to try re-running CI again before merging. there may be some bigger issue with the github e2e test though (but not technically blocking if it fails since we have Cirrus too)

ErikSin:
> might want to try re-running CI again before merging. there may be some bigger issue with the github e2e test though (but not technically blocking if it fails since we have Cirrus too)

Wierd, it looks like Bitrise isnt even running. I will take a look at it today!

ErikSin:
Cirrus CI is passing

PR #887, opened 2022-01-03
Upgrade some deps and refactor some code prior to React Native upgrade

achou11:
Extracts several changes from #867 to make it easier to review that PR and theoretically safer to merge. The changes in this PR do not strictly rely on React Native being upgraded, although a few of the files will have some follow up changes (reflected in the other PR) to account for upgrade-specific alterations

Summary of changes:
- Upgrades to some deps, many of which will address a new deprecation warning that later versions of React Native surface
- Ports new implementation of Bugsnag integration
- Conversions and/or refactors to `PermissionsContext`, `LocationContext` and `AppLoading`, which will be affected by the React Native upgrade
- Small changes to `MapView`, which will be affected by the React Native upgrade to a small extent

Notable potentially-relevant changes from other PR that aren't ported to this one:
- Upgrade `@react-native-mapbox-gl/maps`: the upgrade seems to reliant on using a more recent version of React Native, so will keep that and its relevant changes in the other PR
- any changes related to removing the use of `removeEventListener` in favor of `subscription.remove()`. This is dependent on upgrading React Native and thus those changes will remain in the other PR


ErikSin (approved):
Looks great to me (sorry for bombarding you with so many questions on slack, but I thought it would be faster then doing it through github!)

PR #888, opened 2022-01-05
chore: more small fixes prior to React Native upgrade

achou11:
Extracts a few more things from #867 (that I missed in #887) that'll make the former slightly easier to review

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -386,9 +386,6 @@ dependencies {
         implementation jscFlavor
     }
 
-    implementation 'androidx.appcompat:appcompat:1.1.0-rc01'
-    implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0-alpha02'
-
     // Detox
     androidTestImplementation 'com.wix:detox:+'
     androidTestImplementation 'com.linkedin.testbutler:test-butler-library:2.2.1'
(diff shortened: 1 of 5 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,7 @@
   <uses-permission android:name="android.permission.INTERNET" />
   <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
   <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+  <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
```
achou11, line 7: for context, think this is automatically added via expo-location but explicitly added here for clarity

https://github.com/expo/expo/tree/master/packages/expo-location#configure-for-android

android/app/build.gradle
```diff
@@ -386,9 +386,6 @@ dependencies {
         implementation jscFlavor
     }
 
-    implementation 'androidx.appcompat:appcompat:1.1.0-rc01'
-    implementation 'androidx.swiperefreshlayout:swiperefreshlayout:1.1.0-alpha02'
```
achou11, line 390: these are no longer needed since `react-native-screens` specifies them directly now

https://github.com/software-mansion/react-native-screens/blob/6c87d7749ec62fbb51fb4ec50af1fa8733ebae86/android/build.gradle#L58

https://github.com/software-mansion/react-native-screens/blob/6c87d7749ec62fbb51fb4ec50af1fa8733ebae86/android/build.gradle#L61

ErikSin (approved):

gmaclennan (approved):

PR #919, opened 2022-01-27
chore: Update gradle wrapper to use 6.9.2

achou11:
More of a WIP PR right now, but opening to test CI

Suspicion is that 6.9.0 contains a bug causing one of our CI steps to fail sporadically (e.g. https://github.com/digidem/mapeo-mobile/runs/4968297075?check_suite_focus=true). Potentially relevant issue was reported (https://github.com/gradle/gradle/issues/15536) and apparently has been addressed (https://github.com/gradle/gradle/pull/18572). According to https://github.com/gradle/gradle/pull/18572#pullrequestreview-837242221, looks like the fix was backported to 6.9.2

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -48,7 +48,7 @@ task:
     fingerprint_script: echo $CIRRUS_BUILD_ID
   build_detox_script: npm run build-detox-android
   cleanup_before_cache_script:
-    - rm -rf ~/.gradle/caches/6.2/
+    - rm -rf ~/.gradle/caches/6.9.2/
     - rm -rf ~/.gradle/caches/transforms-1
     - rm -rf ~/.gradle/caches/journal-1
     - rm -rf ~/.gradle/caches/jars-3/*/buildSrc.jar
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -48,7 +48,7 @@ task:
     fingerprint_script: echo $CIRRUS_BUILD_ID
   build_detox_script: npm run build-detox-android
   cleanup_before_cache_script:
-    - rm -rf ~/.gradle/caches/6.2/
+    - rm -rf ~/.gradle/caches/6.9.2/
```
gmaclennan, line 51: Based on what I see in this folder on my machine, this should be `~/.gradle/caches/6.9` - e.g. no patch number.

achou11: if you rerun the android project (e.g. `npm run android`), you should see a cache folder for `6.9.2` too. I think the `6.9` was created from previous runs of the android commands when we had the version set to 6.9 (i.e. before this PR's change)

gmaclennan (changes requested):
Good detective work. If CI passes lgtm, once you make the small change I highlight.

.cirrus.yml
gmaclennan: oh right! yeah, makes sense. Let's leave as-is then.

achou11: when you added this to Cirrus, we were on `6.2` so I assumed that this step should be removing the cache of the currently used gradle version. lmk if that's incorrect though

PR #923, opened 2022-02-11
Kill passcode

### LINKED ISSUE #896: (Front) Kill Passcode Screen
**Description :**
User has set their app passcode and now can access to the Kill Passcode via Settings

**Screen is:**
1-
Kill Passcode
What is Kill Passcode?
(Add text explaining feature)
Use Kill Pascode (Toggle)

2- 
Kill Passcode
What is Kill Passcode?
(Add text explaining feature)
Use Kill Pascode (Toggle)
*0000* Enter the code above to hide sensitive data in Mapeo

3-
Security
Device Security
App Passcode
Passcode is set

Kill Passcode
Kill Passcode is set

Device Backup
Paper Key
Reinstate your account if this device is lost

**Copydeck**
https://docs.google.com/spreadsheets/d/1od1MdXpS32VjTSGgQQmILp72YPXlvxbfuh95q_JxWTI/edit#gid=1060814760

**Design**
https://www.figma.com/file/ZOsM89L0xy0NUOCkuksJgq/OTF-%2B-UXFund?node-id=47974%3A1260

 
To Do
- [x]  Discuss, during sprint planning, State Management of how we know when kill passcode is active. (Security Context)
- [ ] Build [Screen]()

ErikSin:
# Kill Password

## Features:
Password Screen: Activated when app is set in background, and a password is set.

Kill State: Kill State can be turned on when password is set. If kill state is active, user can type "00000" as the password, kill state will be activated

In kill state, All observations are hidden. Kill state and Password settings are also hidden.

To get out of kill state the user can type in the normal password (currently set to "12345")

Kapture 2022-02-11 at 18 56 26

## To Do
Set Password by user


src/frontend/AppContainerWrapper.tsx
```diff
@@ -47,7 +51,7 @@ const loadSavedNavState = async (includeOnboarding: boolean) => {
   try {
     const navState = JSON.parse(
       (await AsyncStorage.getItem(getNavStoreKey(includeOnboarding))) as string
-    );
+    ) as NavigationState;
```
achou11, line 57: Probably worth also annotating the wrapping function here e.g.

```ts
const loadSavedNavState = async (includeOnboarding: boolean): Promise<NavigationState | null> => {
```

you'd need to add a `return null` in the catch branch in this function too

src/frontend/AppContainerWrapper.tsx
```diff
@@ -21,6 +22,9 @@ import bugsnag from "./lib/logger";
 import { PracticeMode } from "./sharedComponents/PracticeMode";
 import DefaultContainer from "./Navigation/DefaultContainer";
 import OnboardingContainer from "./Navigation/OnboardingContainer";
+import { SecurityContext } from "./context/SecurityContext";
+import { AuthStack } from "./Navigation/AuthStack";
+import { AppState, AppStateStatus, LogBox } from "react-native";
```
achou11: remove unused import `LogBox`

src/frontend/AppContainerWrapper.tsx
```diff
@@ -94,11 +98,43 @@ function hidePracticeModeTemporarily(route: string | null) {
...
   const [hidePracticeMode, setHidePracticeMode] = React.useState(false);
+  const {
+    killStateActive,
+    passcode,
+    checkFlag,
+    setCheckFlag,
+  } = React.useContext(SecurityContext);
+
+  React.useEffect(() => {
+    const StateListener = AppState.addEventListener(
```
achou11: nit: camel case and rename to something more descriptive e.g. `appStateListener`

src/frontend/AppContainerWrapper.tsx
```diff
@@ -94,11 +98,43 @@ function hidePracticeModeTemporarily(route: string | null) {
...
+  }, []);
+
+  const handleStateAndSetCheckPassFlag = (nextAppState: AppStateStatus) => {
+    if (
+      (appState.current.match(/active/) && nextAppState === "inactive") ||
+      "background"
+    ) {
+      setCheckFlag(true);
+    }
+  };
```
achou11: nit: maybe just inline this to the `addEventListener` in the effect above?

src/frontend/AppContainerWrapper.tsx
```diff
@@ -142,7 +178,7 @@ const AppContainerWrapper = () => {
                 experiments.onboarding
               );
 
-              const loadedRouteName = getRouteName(loadedNavState);
+              const loadedRouteName = getRouteName(loadedNavState!);
```
achou11: generally not a fan of this operator. I would either update `loadSavedNavState` so that it definitively returns `NavigationState | null` (right now it returns `NavigationState | null | void`)

src/frontend/AppContainerWrapper.tsx
```diff
@@ -177,22 +209,35 @@ const AppContainerWrapper = () => {
     }
   }, [inviteModalEnabled, queuedInvite, openInviteModal]);
 
+  React.useEffect(() => {}, []);
```
achou11: remove

src/frontend/constants.ts
```diff
@@ -1,5 +1,7 @@
 export const URI_PREFIX = "mapeo://";
 export const ERROR_STORE_KEY = "@MapeoError";
+export const PASSWORD_KEY = "PASSWORD";
+export const ASYNC_KILL_KEY = "KillPassKey";
```
achou11: nit: remove `ASYNC` here

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,128 @@
+import * as React from "react";
+import AsyncStorage from "@react-native-async-storage/async-storage";
+import { ASYNC_KILL_KEY, PASSWORD_KEY } from "../constants";
+import { useEffect } from "react";
+import { useRef } from "react";
```
achou11: remove repetitive react imports here

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+interface SecuritContextType {
+  killState: boolean;
+  setKillState: (killStateValue: boolean) => void;
+  passcode: string | null;
+  setPasscode: (passValue: string | null) => void;
+  checkFlag: boolean;
+  killStateActive: boolean;
+  setKillStateActive: (isActive: boolean) => void;
+  setCheckFlag: (checkFlag: boolean) => void;
+}
```
achou11: wondering if there's a cleaner way of structuring this - maybe with the help of useReducer 🤔 doesn't feel right to have to pass a ton of state variables and each of their setters in a context

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+  children: React.ReactNode;
+}) => {
+  const [killState, setKillState] = React.useState(false);
+  const [killStateActive, setKillStateActive] = React.useState(false);
+  const [passcode, setPasscode] = React.useState<string | null>(null);
+  const [checkFlag, setCheckFlag] = React.useState(false);
+  const firstRender = useRef(true);
+
+  useEffect(() => {
+    if (firstRender.current) {
```
achou11: is this check needed if this is an effect with a array dep of `[]`?

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+        if (!!val) {
+          setPasscode(val);
+          setCheckFlag(true);
+          AsyncStorage.getItem(ASYNC_KILL_KEY, (err, val) => {
+            if (!!val) setKillStateActive(val === "true");
+          });
+        } else {
+          firstRender.current = false;
+        }
+      });
```
achou11: prefer using the promise-based API, as the callback style is considered legacy https://react-native-async-storage.github.io/async-storage/docs/api#getitem

achou11: also, consider using [AsyncStorage.multiget](https://react-native-async-storage.github.io/async-storage/docs/api#multiget) instead e.g.

```TS
AsyncStorage.multiGet([PASSWORD_KEY, ASYNC_KILL_KEY])
  .then(([password, killState]) => { ... })
  .catch(err => { ... })
```

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export const SecurityProvider = ({
+  children,
+}: {
+  children: React.ReactNode;
+}) => {
+  const [killState, setKillState] = React.useState(false);
+  const [killStateActive, setKillStateActive] = React.useState(false);
+  const [passcode, setPasscode] = React.useState<string | null>(null);
+  const [checkFlag, setCheckFlag] = React.useState(false);
```
achou11: does this need to be its own state variable? is it different than it being derived from `passcode` + `klilStateActive` i.e. `passcode && killState`?

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+  DEFAULT_CONTEXT
+);
+
+export const SecurityProvider = ({
+  children,
+}: {
+  children: React.ReactNode;
+}) => {
+  const [killState, setKillState] = React.useState(false);
+  const [killStateActive, setKillStateActive] = React.useState(false);
```
achou11: what exactly is the difference between these? is `killState` for allowing the option to exist in the first place and `killStateActive` reflecting whether or not it's actually been enabled by the user? Either way, would update the variable names to make this clearer

at a glance, potentially could simplify it to a single variable

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+      return;
+    }
+
+    if (!passcode) {
+      setKillState(false);
+      setKillStateActive(false);
+      setCheckFlag(false);
+      AsyncStorage.setItem(PASSWORD_KEY, "");
+    } else {
+      AsyncStorage.setItem(PASSWORD_KEY, passcode);
```
achou11: although it makes sense to run this in an effect since it's dependent on `passcode`, wondering if it's safer and less prone to catastrophic runaway effects if you call setItem in the passcode updater function `setPasscodeWithCheck`. just thinking that it could simplify the implementation a bit and remove the need for this effect, which is doing a somewhat a non-ideal check based on react render cycles

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  return (
+    <SecurityContext.Provider value={contextValues}>
+      {children}
+    </SecurityContext.Provider>
+  );
+};
+
+function validPasscode(passcode: string): boolean {
+  return passcode.length === 5 && !isNaN(parseInt(passcode));
```
achou11: as a precaution, pass `10` as the radix in `parseInt` i.e. `parseInt(passcode, 10))` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#description

src/frontend/hooks/useAllObservations.ts
```diff
@@ -14,16 +15,16 @@ export const useAllObservations = (): [
...
-  const observationsArray = useMemo(
-    () =>
-      Array.from(state.observations.values()).sort((a, b) =>
-        // TODO: move sorting into component
-        a.created_at < b.created_at ? 1 : -1
-      ),
-    [state.observations]
-  );
+  const observationsArray = useMemo(() => {
+    if (killState) return [];
```
achou11: mentioned in an earlier comment, but not entirely sure what the difference between the two state variables you defined are. based on my limited knowledge, would this be `killStateActive` instead?

src/frontend/screens/AppPasscode/EnterPasscode.tsx
```diff
@@ -1,16 +1,24 @@
...
-  const { setPassIsSet } = React.useContext(SecurityContext);
+  const { passcode, setPasscode } = React.useContext(SecurityContext);
 
   return (
     <React.Fragment>
-      <Text>Under Construction</Text>
-      <Button onPress={() => setPassIsSet(false)}>Turn Off App Passcode</Button>
+      <Text>{passcode}</Text>
+      <Button
+        onPress={async () => {
```
achou11: remove `async` here

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }, [inputtedPass]);
+
+  React.useEffect(() => {
+    AsyncStorage.getItem(PASSWORD_KEY, (err, result) => {
+      if (!!err) return;
+
+      userPass.current = result || "";
+    });
+  }, []);
```
achou11: slightly confused about what this is for. would this be prefilling the input? (doesn't seem like desired behavior)

src/frontend/context/AppProvider.tsx
```diff
@@ -6,11 +6,11 @@ import { ObservationsProvider } from "./ObservationsContext";
 import { ConfigProvider } from "./ConfigContext";
 import { SettingsProvider } from "./SettingsContext";
 import { DraftObservationProvider } from "./DraftObservationContext";
-import { SecurityProvider } from "../screens/Security/SecurityContext";
+import { SecurityProvider } from "./SecurityContext";
 
 // This is a convenience wrapper for providing all App contexts to the tree,
 // apart from the Permissions Provider which is needed separately.
-const AppProvider = ({ children }: { children: React.Node }) => (
+const AppProvider = ({ children }: { children: React.ReactNode }) => (
```
achou11, line 12: Can remove the `// @flow` in line 1 in this file since it's now `.tsx`

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,197 @@
...
+  );
+};
+
+function validatePassword(password: string) {
+  if (password.length !== CELL_COUNT) return false;
+
+  const passAsArray = Array.from(password);
+  return passAsArray.reduce((acc, letter) => {
+    return onlyNumRegEx.test(letter);
+  }, false);
```
achou11: nit: would use ~[`Array.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) here instead~ sorry i meant [`Array.every`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)

also, the current implementation here only returns true if the last character passes the regex test, which i don't think was the intention

achou11: better yet, just regex the entire string as is i.e. `onlyNumRegEx.test(password)`

src/frontend/context/SecurityContext.tsx
ErikSin: So Killstate, is when the user is in Kill mode (aka, no observations, and looks like it is fresh download). KillStateActive indicated that the user CAN go into kill mode. I could reduce this to an object, which might be clearer. 

I do agree these variable names are not the best. Do you have any suggestion for clearer names (im strugging to think of anything)

src/frontend/context/SecurityContext.tsx
ErikSin: I thought of doing that orignally, but I do want to update these BASED on an updated passcode. So this gurantees that when passcode is updated then asyncStorage would be updated

src/frontend/screens/AuthScreen.tsx
ErikSin: This is for saving the users password to compare with the user input, so I need somewhere in memory to save the password that does not effect state.

src/frontend/context/SecurityContext.tsx
ErikSin: When a user sets the password, I do not want the user to be redirected to the auth page. I only want them to be redirected to the auth page, when they open the app from closing it, or when they open it again from the app being in the background. This is what this flag is for.

src/frontend/context/SecurityContext.tsx
ErikSin: So when a user open the app, the password must be set to the variable in memory (`passcode`). When `passcode` gets set, I have THIS listener which sets asyncstorage to the new `passcode`. This check basically just avoid the asyncstorage being unecessarily set the first time

src/frontend/screens/AuthScreen.tsx
ErikSin: Deleted, this was from previous code that!

src/frontend/context/SecurityContext.tsx
achou11: I see. Perhaps the states here can be modified to make this clearer then:

```ts
//   - setAppMode('kill'): when user enters kill code when logging in
//   - setAppMode('normal'): when user exists kill mode (however that's done)
const [appMode, setAppMode] = useState<'normal' | 'kill'>('normal');

```

That makes sense overall, I just really dislike ^when something only has 2 types instead of being a boolean...BUT I think this is one case where it makes sense to do that

src/frontend/context/SecurityContext.tsx
achou11: got it. see my other [comment](https://github.com/digidem/mapeo-mobile/pull/923/files#r807381318) for an idea on how to make this clearer

src/frontend/context/SecurityContext.tsx
achou11: that's reasonable. perhaps the async storage calls here can be moved to `setPasscodeWithCheck` since that's the only source of updates for the passcode state (aside from the initial render effect in this component). i like the idea of co-locating those operations to avoid indirection and potentially undesired updates to async storage 

think the firstRender check is kind of a code smell though so just wondering if there's a way to avoid it in this case

src/frontend/context/SecurityContext.tsx
achou11: think this becomes less of a problem if you move the setItem calls out of the effect below (see [comment](https://github.com/digidem/mapeo-mobile/pull/923/files#r807390019))

src/frontend/context/SecurityContext.tsx
ErikSin: Yeah, originally this was only going to have 4 values, and then it kind of blew up....Ill convert it with useReducer

ErikSin: Are you thinking one reducer function to manage all states in one object? Or just one useReducer for each state?

src/frontend/context/SecurityContext.tsx
ErikSin: Whoops sorry, I just edited your comment directly, meant to reply!

src/frontend/context/SecurityContext.tsx
ErikSin: Ok, I cleaned it up, check out the new implementation using useReducer. It is not completed, but I have the scaffolding done. I will finish it on thursday. Thanks for the initial review, very helpful!

src/frontend/AppContainerWrapper.tsx
```diff
@@ -94,11 +98,44 @@ function hidePracticeModeTemporarily(route: string | null) {
...
+  };
+
+  const AppContainer = React.useMemo(() => {
+    if (authState.authStatus === "pending") {
+      return AuthContainer;
+    }
+
+    return experiments.onboarding ? OnboardingContainer : DefaultContainer;
+  }, [experiments.onboarding, authState.authStatus]);
+
```
ErikSin: Our navigation stacks are becoming a little confusing, so I followed @achou11's lead here and took advantage of the functional component useMemo Hook save the current appContainer

src/frontend/AppContainerWrapper.tsx
```diff
@@ -94,11 +98,44 @@ function hidePracticeModeTemporarily(route: string | null) {
...
+    );
+
+    return () => appStateListener.remove();
+  }, [authState.authStatus]);
+
+  const handleStateAndSetCheckPassFlag = (nextAppState: AppStateStatus) => {
+    console.log("newAppStatus " + nextAppState);
+    console.log("appState");
+    if (
+      appState.current.match(/active/) &&
```
achou11: curious but do you need the ref setup in order to check the previous/current app state here? Wondering if you could remove the ref and have `AppState.currentState.match(/active/)` here instead.

Tangentially, do you need this first check at all? Wondering if the change event only goes to `background` if the previous state was something that wasn't background i.e. there's never a case where a `change` event is triggered due to `background -> background`

src/frontend/AppContainerWrapper.tsx
```diff
@@ -138,9 +175,8 @@ const AppContainerWrapper = () => {
         ? {}
         : {
             loadNavigationState: async () => {
-              const loadedNavState = await loadSavedNavState(
-                experiments.onboarding
-              );
+              const loadedNavState =
+                (await loadSavedNavState(experiments.onboarding)) || undefined;
```
achou11: would lean towards suggestion in https://github.com/digidem/mapeo-mobile/pull/923/files#r806082136 instead of the fallback here

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,130 @@
+import * as React from "react";
+import AsyncStorage from "@react-native-async-storage/async-storage";
+
+import { PASSWORD_KEY } from "../constants";
+
+type AppModeTypes = "normal" | "kill";
+type AuthStatusTypes = "pending" | "authenticated" | "notRequired";
```
achou11, line 7: nit: but slightly redundant to suffix with `Types` here

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,130 @@
...
+  appMode: AppModeTypes;
+  killModeEnabled: boolean;
+  authStatus: AuthStatusTypes;
+};
+
+type KillModeActions =
+  | { type: "setPasscode"; newPasscode: string | null }
+  | { type: "toggleAppMode"; newAppMode?: AppModeTypes }
+  | { type: "toggleKillModeEnabled"; newKillModeValue?: boolean }
+  | { type: "setAuthStatus"; newAuthStatus: AuthStatusTypes };
```
achou11, line 20: nit: usually like to create named types for each of these actions e.g.

```ts
type SetPasscodeAction = { type: 'setPasscode'; newPasscode: string | null }
type ToggleAppModeAction = ...

...

type KillModeActions = SetPasscodeAction | ToggleAppModeAction | ...
```

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+function killModeReducer(state: AuthState, action: KillModeActions): AuthState {
```
achou11: would rename to something more general and related to Security since it's not just handling kill mode state

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,130 @@
...
+     * If the auth status is set to 'pending' they will be navigated to the auth screen, to type in their password.
+     */
+    case "setPasscode":
+      const passcode = action.newPasscode;
+      if (passcode === null || validPasscode(passcode)) {
+        AsyncStorage.setItem(PASSWORD_KEY, passcode || "");
+        if (!passcode) {
+          return { ...state, passcode, authStatus: "notRequired" };
+        }
+        return { ...state, passcode, authStatus: "authenticated" };
```
achou11: nit but can simplify:

```ts
return { ...state, passcode, authStatus: passcode ? "authenticated" : "notRequired };
```

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,130 @@
...
+      }
+      throw new Error("Invalid New Password");
```
achou11: would move this to the beginning of the function and write the appropriate check to see if we should throw or not. something along the lines of

```ts
const passcodeIsValid = passcode === null || validPassCode(passcode);

if (!passcodeIsValid) throw new Error("Invalid New Password")

// state update logic here
```

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,130 @@
...
+     */
+    case "toggleAppMode":
+      return {
+        ...state,
+        appMode: !!action.newAppMode
+          ? action.newAppMode
+          : state.appMode === "kill"
+          ? "normal"
+          : "kill",
+      };
```
achou11: nit: i find the ternary a little hard to parse on quick reads here. definitely harder if we ever end up with more than two valid app modes. without getting into state machines, maybe something like this would be clearer despite being slightly more verbose:

```ts
let mode: AppModeTypes;

if (action.newAppMode) {
  mode = action.newAppMode;
} else {
  switch (state.appMode) {
    case "kill":
      mode = "normal";
      break;
    case "normal":
      mode = "kill";
      break;
  }
}

return {
  ...state,
  appMode: mode,
};
```

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,130 @@
...
+  passcode: string | null;
+  appMode: AppModeTypes;
+  killModeEnabled: boolean;
+  authStatus: AuthStatusTypes;
+};
+
+type KillModeActions =
+  | { type: "setPasscode"; newPasscode: string | null }
+  | { type: "toggleAppMode"; newAppMode?: AppModeTypes }
+  | { type: "toggleKillModeEnabled"; newKillModeValue?: boolean }
```
achou11: after reading the implementations for these toggle actions, it might be clearer to separate them each into two actions, one for toggling and the other for setting a specific value. this way you avoid more complex logic checks based on the payload and state when the intention is purely to toggle. e.g. for app mode:

```ts
// I like the idea of namespacing actions a bit but totally optional
| { type: 'appMode:toggle' }
| { type: 'appMode:set', newAppMode: AppModeTypes } // note that newAppMode is no longer optional here
```

then the reducer implementation becomes more straightforward:

```ts
case 'appMode:toggle':
  return { ...state, appMode: state.appMode === 'kill' ? 'normal' : 'kill' }
case 'appMode:set':
  // this check isn't technically necessary due to types, but could be good to have
  if (!isValidAppMode(action.newAppMode) throw new Error('Invalid app mode')
  return { ...state, appMode: actions.newAppMode }
```

src/frontend/context/SecurityContext.tsx
achou11: think the better suggestion is outlined in this [comment](https://github.com/digidem/mapeo-mobile/pull/923/files#r810353052) i.e. create separate actions + reducers for toggling vs setting state

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,130 @@
...
+      }
+
+      return {
+        ...state,
+        killModeEnabled: !!action.newKillModeValue
+          ? action.newKillModeValue
+          : state.killModeEnabled
+          ? false
+          : true,
+      };
```
achou11: refer to other [comment](https://github.com/digidem/mapeo-mobile/pull/923/files#r810353052) for suggestion on how to simplify this

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const contextValue: SecurityContextType = React.useMemo(() => {
+    return [state, dispatch];
+  }, [state, dispatch]);
+
+  return (
+    <SecurityContext.Provider value={contextValue}>
+      {children}
+    </SecurityContext.Provider>
+  );
+};
```
achou11: noice, think using `useReducer` here works well in this case 👍

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+function killModeReducer(state: AuthState, action: KillModeActions): AuthState {
+  switch (action.type) {
+    /***
+     * if passcode is set to null we need to set the auth status to 'notRequired'
+     *
+     * If a passcode is set, we want to set the authStatus to 'authenticated'
+     * If the auth status is set to 'pending' they will be navigated to the auth screen, to type in their password.
+     */
+    case "setPasscode":
```
achou11, line 32: does the `appMode` need to be updated here too? i.e. if you set `passcode = null`, that would mean that `appMode` should be 'normal' if it was 'kill', right?

achou11: like in theory, there wouldn't be a way to be able to set a passcode and be in kill mode i guess, but for the validity of state changes i think it's worth representing that case

src/frontend/hooks/useAllObservations.ts
achou11: need to update this component based on the new security context implementation

src/frontend/screens/AppPasscode/index.tsx
```diff
@@ -19,7 +19,7 @@ const m = defineMessages({
 });
 
 export const AppPasscode: NavigationStackScreenComponent = () => {
-  const { passIsSet } = React.useContext(SecurityContext);
+  const [authstate, setAuthState] = React.useContext(SecurityContext);
```
achou11: remove unused variable `setAuthState`

src/frontend/context/SecurityContext.tsx
ErikSin: These are being used in a few places, so typing them just makes the coding a little easier

src/frontend/context/SecurityContext.tsx
ErikSin: No, I think that makes alot of sense!

src/frontend/context/SecurityContext.tsx
ErikSin: In reference to this [comment](https://github.com/digidem/mapeo-mobile/pull/923/files#r810359036), I actually left it as an `if` statement because I am now changing 2 values based whether password is null, so ternary operator would have to be used twice

src/frontend/context/SecurityContext.tsx
ErikSin: Yeah, the state hooks really blew up....This is a much nicer architecture. I do think we fall into that trap of all our state being associated with one object...and therefore causing unnecessary re-renders...but I think it is fine for now since it is relatively simple

src/frontend/sharedComponents/PasswordInput.tsx
```diff
@@ -0,0 +1,141 @@
+import * as React from "react";
```
ErikSin, line 1: I made this its own component so we can use it on the other password inputs in the upcoming screens we are building

src/frontend/context/SecurityContext.tsx
ErikSin: This feels like we are adding alot of extra noise....I think I will just use if statments to clear up the double ternary if that is the problem.

src/frontend/context/SecurityContext.tsx
ErikSin: Adding more dispatchers feels like overkill, especially given out use cases, so I added explicit if statments for readibility

src/frontend/context/SecurityContext.tsx
achou11: the problem for me is that the intent of the action doesn't match the logic that happens within it. from a strict sense of "toggling", there shouldn't be a custom value to be provided like the current implementation has it as of right now. setting a value is a very different action than toggling to me. the optionality of a specific value makes the logic more unclear as a result.

having more actions if the actions are clearly defined and less logic to implement feels like a okay tradeoff

ErikSin: I see what you mean. I am worried about over complicating things to have 2 state setters, for the same state, especially when there is only 2 possible states. If we changed the name from `toggle` to `setOrToggle` would that be a good compromise? Or do you think have multiple state setters is the better solution?

achou11: still think having multiple setters is still the clearer solution in this case. there isn't much overhead when it comes to adding actions compared to writing logic to handle complicated and potentially unclear actions. this becomes more apparent when there are more than the 2 states we're starting off with (can imagine this going to 3+ states at some point)

src/frontend/screens/AppPasscode/NewPasscode/PasscodeIntro.tsx
```diff
@@ -26,12 +28,12 @@ interface SplashProps {
...
-  const { setPassIsSet } = React.useContext(SecurityContext);
+  const [authState, setAuthState] = React.useContext(SecurityContext);
   const { formatMessage: t } = useIntl();
   const { navigate } = useNavigation();
 
   function tempStateSet() {
-    setPassIsSet(prev => !prev);
+    setAuthState({ type: "setPasscode", newPasscode: "12345" });
     navigate("Security");
   }
```
achou11, line 38: couple of notes:

- would lean towards removing the navigation call here because with the screen transition, it feels broken/abrupt
- nit: if we keep the function here (which wouldn't be necessary if the first point is taken), can we rename it to something we wouldn't have to change later on?

src/frontend/screens/Security/index.tsx
```diff
@@ -64,52 +65,70 @@ const m = defineMessages({
...
-      <ListDivider style={styles.divder} />
+          <ListItem
+            onPress={() => {
+              if (!authState.passcode) {
+                highlightError();
+                return;
+              }
+              navigate("KillPasscode");
+            }}
+          >
```
achou11, line 118: not a blocker but feel like it would be more useful to disable the option to enable this in the actual kill passcode screen, rather in this security screen. in its current state, it's impossible to even find out what the kill passcode is, which i feel could be useful for someone who's considering the option but hasn't commited to creating an app passcode just yet

src/frontend/screens/AppPasscode/NewPasscode/PasscodeIntro.tsx
ErikSin: I am working on this part during this sprint, so this will be completely changed. This is just placeholder code!

src/frontend/screens/Security/index.tsx
ErikSin: Yeah, perhaps we should bring this up with Sabella, because this is her current design flow, but I do agree with your sentiment.

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+      AsyncStorage.setItem(PASSWORD_KEY, passcode || "");
+      if (!passcode) {
+        AsyncStorage.setItem(KILL_KEY, JSON.stringify(false));
+        return {
+          ...state,
+          passcode,
+          authStatus: "notRequired",
+          appMode: "normal",
+        };
```
achou11: wondering if this should also set `killModeEnabled` to `false` i.e. turning off the app passcode also resets the kill mode state. right now, the following occurs:

1. Turn on passcode
2. Turn on kill mode
3. Turn off passcode
4. Turn on passcode
5. `killModeEnabled` is `true`

in 5, i'd kind of expect that to be `false` i.e. the app passcode determines how that state resets. thoughts?

ErikSin: You are correct, good catch.

src/frontend/sharedComponents/PasswordInput.tsx
```diff
@@ -0,0 +1,134 @@
...
+const onlyNumRegEx = new RegExp("^[0-9]+$");
+
+interface PasswordInputProps {
+  handleError: () => void;
+  handleCorrectOrNewPass: (inputtedValue: string, clear: () => void) => void;
+  clearError: () => void;
+  stylesProps?: StyleProp<ViewStyle>;
+}
+
+export const PasswordInput = ({
```
achou11, line 24: couple of bugs/ux issues i'm noticing:

- the inputs aren't vertically centered, which is strangely distracting 😅 
- thoughts on autofocusing this on initial render? or providing a prop to have that option?
- can't delete the first digit using the delete button
- not a fan of clearing the input and blurring when it's determined that it's incorrect
- feeling mixed on the library's default behavior where pressing a cell clears the value of the cells after it - guess not as big of an issue but kind of annoying if i just wanted to fix a single digit

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const m = defineMessages({
+  enterPass: {
+    id: "screens.EnterPassword.enterPass",
+    defaultMessage: "Enter your passcode",
+  },
+  wrongPass: {
+    id: "screens.EnterPassword.wrongPass",
+    defaultMessage: "Incorrect password, please try again ",
+  },
```
achou11: replace `password` with `passcode` to keep consistency in wording?

src/frontend/AppContainerWrapper.tsx
```diff
@@ -92,13 +99,42 @@ function hidePracticeModeTemporarily(route: string | null) {
   return route !== null && TEMP_HIDE_PRACTICE_MODE_UI.includes(route);
 }
 
+const AuthContainer = createAppContainer(AuthStack);
```
achou11: remove unused variable + associated imports

src/frontend/AppContainerWrapper.tsx
```diff
@@ -92,13 +99,42 @@ function hidePracticeModeTemporarily(route: string | null) {
...
   const [queuedInvite, setQueuedInvite] = React.useState<string | null>(null);
   const [{ experiments }] = React.useContext(SettingsContext);
   const [hidePracticeBar, setHidePracticeBar] = React.useState(true);
   const [hidePracticeMode, setHidePracticeMode] = React.useState(false);
+  const [{ authStatus }, setAuthState] = React.useContext(SecurityContext);
+  const isMounted = React.useRef(false);
+
+  React.useEffect(() => {
+    isMounted.current = true;
+  }, []);
```
achou11: this doesn't seem to be used, okay to remove? also fyi but we have a hook for this too: `useIsMounted`.

in the future, if this is just to prevent that common React warning about setting state on an unmounted component, this check will be completely unnecessary. there's an issue explaining how that warning is misleading and isn't actually a problem for unmounted components in most cases

src/frontend/AppContainerWrapper.tsx
```diff
@@ -92,13 +99,42 @@ function hidePracticeModeTemporarily(route: string | null) {
...
+
+  React.useEffect(() => {
+    const appStateListener = AppState.addEventListener(
+      "change",
+      handleStateAndSetCheckPassFlag
```
achou11: may have already commented this before, but would move the function definition into this effect because:

1. it's not used elsewhere in the component
2. it's missing in the effects dep list, so just easier to avoid remembering to do that

src/frontend/AppContainerWrapper.tsx
ErikSin: ahhh really? good to know!

src/frontend/AppContainerWrapper.tsx
achou11: even simpler would be to inline the function i.e.:

```ts
  React.useEffect(() => {
    const appStateListener = AppState.addEventListener(
      "change",
      (nextAppState: AppStateStatus) => {
        if (authStatus !== "notRequired") {
          if (nextAppState === "inactive" || nextAppState === "background") {
            setAuthState({ type: "setAuthStatus", newAuthStatus: "pending" });
          }
        }
      }
    );

    return () => appStateListener.remove();
  }, [authStatus, setAuthStatus]);
```

src/frontend/AppContainerWrapper.tsx
```diff
@@ -92,13 +99,42 @@ function hidePracticeModeTemporarily(route: string | null) {
...
+    );
+
+    return () => appStateListener.remove();
+  }, [authStatus, navRef.current]);
```
achou11: remove `navRef.current` here because adding refs to effects deps is generally never needed/recommended (https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/)

src/frontend/AppLoading.tsx
```diff
@@ -6,6 +6,7 @@ import api, { Constants, ServerStatus } from "./api";
 import { PERMISSIONS } from "./context/PermissionsContext";
 import { usePermissions } from "./hooks/usePermissions";
 import ServerStatusScreen from "./screens/ServerStatus";
+import { SecurityContext } from "./context/SecurityContext";
```
achou11: remove unused import

src/frontend/screens/AppPasscode/EnterPasscode.tsx
```diff
@@ -1,16 +1,21 @@
...
     <React.Fragment>
-      <Text>Under Construction</Text>
-      <Button onPress={() => setPassIsSet(false)}>Turn Off App Passcode</Button>
+      <Text>{authState.passcode}</Text>
+      <Button
+        onPress={() => {
+          setAuthState({ type: "setPasscode", newPasscode: null });
+        }}
+      >
+        Turn Off App Passcode
```
achou11, line 17: Add translation string for this

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,134 @@
+import * as React from "react";
+import { defineMessages, FormattedMessage } from "react-intl";
+import { View, Image, Text, StyleSheet } from "react-native";
+import { NavigationStackScreenComponent } from "react-navigation-stack";
+import { DARK_BLUE, DARK_GREY, MEDIUM_GREY, WARNING_RED } from "../lib/styles";
+import AsyncStorage from "@react-native-async-storage/async-storage";
+
+import { KILL_PASSCODE, PASSWORD_KEY } from "../constants";
+import { AuthState, SecurityContext } from "../context/SecurityContext";
```
achou11: remove unused variable `AuthState`

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,134 @@
...
+    id: "screens.EnterPassword.wrongPass",
+    defaultMessage: "Incorrect password, please try again ",
+  },
+});
+
+export const AuthScreen: NavigationStackScreenComponent = () => {
+  const [error, setError] = React.useState(false);
+  const [authState, setAuthState] = React.useContext(SecurityContext);
+
+  const { navigate } = useNavigation();
```
achou11, line 26: remove unused variable and associated import

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  async function validatePass(inputtedPass: string, clearInput: () => void) {
+    console.log("input" + inputtedPass);
+    console.log("currentPass" + authState.passcode);
+    console.log(await AsyncStorage.getItem(PASSWORD_KEY));
```
achou11: remove

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,134 @@
...
+    setError(true);
+  }
+
+  function clearError() {
+    setError(false);
+  }
+
+  React.useEffect(() => {
+    console.log("Outside", authState.passcode);
+  }, []);
```
achou11: remove

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,134 @@
...
+const m = defineMessages({
+  enterPass: {
+    id: "screens.EnterPassword.enterPass",
+    defaultMessage: "Enter your password",
+  },
+  wrongPass: {
+    id: "screens.EnterPassword.wrongPass",
+    defaultMessage: "Incorrect password, please try again ",
+  },
+});
```
achou11: replace `password` with `passcode` for consistency

src/frontend/screens/AuthWrapper.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from "react";
+import { createAppContainer } from "react-navigation";
+import AppContainerWrapper from "../AppContainerWrapper";
+import { SecurityContext } from "../context/SecurityContext";
+import { AuthStack } from "../Navigation/AuthStack";
+
+const AuthContainer = createAppContainer(AuthStack);
+
+export const AuthWrapper = () => {
```
achou11, line 6: think the conditional rendering here is causing a moment where the screen is completely grey for about a second after successful passcode input. could the authscreen be implemented as a modal to avoid this issue? would have to play around with it otherwise to figure out how to fix because i don't have any great ideas off the top of my head

src/frontend/screens/KillPasscode.tsx
```diff
@@ -0,0 +1,124 @@
...
+      <View style={styles.switch}>
+        <Text>
+          <FormattedMessage {...m.toggleMessage} />
+        </Text>
+        <Switch
+          onChange={() => {
+            dispatch({ type: "killModeEnabled:toggle" });
+          }}
+          value={state.killModeEnabled}
+        />
```
achou11: think we discussed using a checkbox for this (not sure if that was gonna be a follow up), but if not, just wanted to note that it's kind of hard to see the switch in its off state (it just shows the knob but nothing else)

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,134 @@
...
+    console.log("currentPass" + authState.passcode);
+    console.log(await AsyncStorage.getItem(PASSWORD_KEY));
+    if (inputtedPass === KILL_PASSCODE && authState.killModeEnabled) {
+      setAuthState({ type: "appMode:set", newAppMode: "kill" });
+      setAuthState({
+        type: "setAuthStatus",
+        newAuthStatus: "authenticated",
+      });
+      return;
+    }
```
achou11: not a blocker, but would it make sense to always go to the map screen after entering the kill passcode instead of the last visited one? maybe it doesn't but just something i wanted to pose.

not sure where that logic would live but reading this got me thinking about that

src/frontend/context/SecurityContext.tsx
achou11: think there's a misunderstanding here. was purely referring to the names of the type variables i.e. `AppModeTypes -> AppMode` and `AuthStatusTypes -> AuthStatus`

src/frontend/screens/KillPasscode.tsx
ErikSin: We discussed using a checkbox instead of a switch for other thing, but the designs were never updated for this component, so we should check in with Sabella about that as well!

src/frontend/AppContainerWrapper.tsx
```diff
@@ -99,6 +105,26 @@ const AppContainerWrapper = () => {
...
+        if (authStatus !== "notRequired") {
+          if (nextAppState === "inactive" || nextAppState === "background") {
+            setAuthState({ type: "setAuthStatus", newAuthStatus: "pending" });
+          }
+        }
+      }
+    );
+
+    return () => appStateListener.remove();
+  }, [authStatus]);
```
achou11: sorry i forgot to add this in my previous suggestion initially but i _think_ you need to add `setAuthState` here too?

```suggestion
  }, [authStatus. setAuthState]);
```

src/frontend/screens/AuthWrapper.tsx
ErikSin: The problem with the modal, is android has no real way to tracking when it gets closed, so we are unable to add the modal to navigation state on close. So when the user closes the app, and opens again, the nav state persists to NOT the auth screen. The initial loading screen/app screen doesn have access to the navigation, so there is not way of dispatching the modal in that loading screen

src/frontend/screens/AppPasscode/EnterPasscode.tsx
ErikSin: This is just placeholder code! working on changing all this stuff this sprint

src/frontend/screens/AuthWrapper.tsx
achou11: makes sense. i think that brings up another question of if it's even necessary for this screen to be part of (or its own) navigation stack. unless there's something about deeplinking or something along those lines, it doesn't seem like we need to create a new stack navigator for this auth screen. the auth screen appearing is essentially a function of the security context state, and not sure if it'll ever need to be part of the navigation stack to accomplish its purpose. for example, i think if you did something like:

```tsx
// AuthWrapper.tsx
import * as React from "react";
import { createAppContainer } from "react-navigation";
import AppContainerWrapper from "../AppContainerWrapper";
import { SecurityContext } from "../context/SecurityContext";
// import { AuthStack } from "../Navigation/AuthStack";
import { AuthScreen } from "./AuthScreen";

// const AuthContainer = createAppContainer(AuthStack);

export const AuthWrapper = () => {
  const [{ authStatus }] = React.useContext(SecurityContext);

  if (authStatus === "pending") {
    // return <AuthContainer />;
    return <AuthScreen />;
  }

  return <AppContainerWrapper />;
};
```

with the accompanying style change in `AuthScreen.tsx`:

```tsx
  container: {
    paddingHorizontal: 20,
    alignItems: "center",
    // justifyContent: "center",
    // marginTop: 40,
    paddingTop: 40,
    backgroundColor: WHITE,
    flex: 1,
  },
```
you could accomplish the same thing without dealing with navigation complexity. let me know if there's something im missing though.

also, the change above wouldn't fix the original flickering issue, that just may be something related to how the normal app stack works and would need to improve on at some point

ErikSin: Yeah that makes sense, It was originally part of the nav stack because of converting tit to a modal! I looked into the flickering issue and I know how to fix it, so I will address it today!

ErikSin: I changed this to a modal and had to add an intermediate loading screen!

src/frontend/screens/KillPasscode.tsx
ErikSin: Converted to checkbox

src/frontend/sharedComponents/PasswordInput.tsx
ErikSin: Auto focus, vertically centered, and first digit error fixed. I didn't address the other 2, mainly because it is default behaviour and seems more personal preference, then technical :)

ErikSin:
Closed as work has been ported over to #941

achou11:
@ErikSin can we close this? this was the initial PR for the passcode but I think that's been addressed at this point, right?

ErikSin:
yup!

PR #927, opened 2022-02-21
chore: Edit copy for Experiments status and P2P Updates

emreiser:


achou11:
Thanks @emreiser! We'll need to update some parts of the app in this PR so that they reflect these changes first. cc @ErikSin guessing we want to merge this to have it as part of the release

PR #929, opened 2022-03-01
chore(deps): Update mapeo-default-settings to latest

achou11:
Notes:
- Updates `mapeo-default-settings` to latest version (v3.6.0), which most notably adds a default `color` field for various observation categories

Preview:

color-dots-default-config


ErikSin (approved):
Thank you

PR #930, opened 2022-03-01
chore: increase font size in experimental features

ErikSin:
Increasing font size of the new experimental features by request of programs team!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "mapeo-mobile",
-  "version": "5.4.0",
+  "version": "5.3.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


achou11 (approved):
Maybe try rerunning tests but otherwise lgtm

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mapeo-mobile",
-  "version": "5.4.0",
```
achou11, line 3: what's going on here?

achou11: should be 5.3.2 in this case, but curious why there's a diff here 🤔

achou11: ah i see this was probably accidentallly changed in https://github.com/digidem/mapeo-mobile/commit/5d3c965ac7cae21b6b9e3049f9edec6606e9a160

ErikSin:
> Maybe try rerunning tests but otherwise lgtm

yeah, something keeps breaking with our CI, its seems to not be working the first time around

PR #942, opened 2022-03-14
fix: fix regression in release apk build size

achou11:
Turns out that there's no easy way to override the `abiFilters` setting in `defaultConfig` via the `buildTypes` field: https://stackoverflow.com/questions/53322169/how-to-override-defaultconfig-abifilters-in-buildtypes

Specifying the field in `defaultConfig` causes the `buildType`-specific field to be completely ignored, which is why the release build was including all ABIs instead of the ones specified. The solution is to remove that default config specification and patch nodejs-mobile-react-native to properly handle the case where it's not specified in more recent versions of Gradle (handling an empty collection in addition to null)

Preview:

After running `gradlew assembleRelease`:

image

### DIFF
--- a/patches/nodejs-mobile-react-native+0.6.3.patch
+++ b/patches/nodejs-mobile-react-native+0.6.3.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/nodejs-mobile-react-native/android/build.gradle b/node_modules/nodejs-mobile-react-native/android/build.gradle
-index e54f25c..178b15a 100644
+index e54f25c..79b96e2 100644
 --- a/node_modules/nodejs-mobile-react-native/android/build.gradle
 +++ b/node_modules/nodejs-mobile-react-native/android/build.gradle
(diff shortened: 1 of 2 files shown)


patches/nodejs-mobile-react-native+0.6.3.patch
```diff
@@ -25,6 +25,15 @@ index e54f25c..178b15a 100644
          main.assets.srcDirs += '../install/resources/nodejs-modules'
      }
  
+@@ -239,7 +242,7 @@ if ("1".equals(shouldRebuildNativeModules)) {
+     GenerateNodeProjectAssetsLists.dependsOn "ApplyPatchScriptToModules"
+ 
+     def nativeModulesABIs = android.defaultConfig.ndk.abiFilters;
+-    if (nativeModulesABIs == null) {
++    if (!nativeModulesABIs) {
```
achou11, line 33: somewhat lazy approach as `null` and empty collections in groovy are both falsy http://groovy-lang.org/semantics.html#the-groovy-truth

achou11:
Following the hotfix release flow so closing in favor of https://github.com/digidem/mapeo-mobile/pull/943

PR #943, opened 2022-03-14
Hotfix v5.4.1

achou11:
Changes:

- Removes the release description generation step from Bitrise since it is flaky and isn't completely necessary
- Fixes a regression in apk size for the release build. Turns out that there's no easy way to override the `abiFilters` setting in `defaultConfig` via the `buildTypes` field: https://stackoverflow.com/questions/53322169/how-to-override-defaultconfig-abifilters-in-buildtypes
    - Specifying the field in `defaultConfig` causes the `buildType`-specific field to be completely ignored, which is why the release build was including all ABIs instead of the ones specified. The solution is to remove that default config specification and patch nodejs-mobile-react-native to properly handle the case where it's not specified in more recent versions of Gradle (handling an empty collection in addition to null)
    - Preview (after running `gradlew assembleRelease`):
      
      image

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -192,25 +192,13 @@ workflows:
                 set -x
 
                 node ./scripts/gh-pages-redirect.js /icca/latest https://mapeo-apks.s3.amazonaws.com/release-ICCA/${ICCA_RELEASE_APK_PATH##*/}
-      - script@1.1.6:
-          title: Generate release description
-          inputs:
-            - content: |-
-                #!/usr/bin/env bash
-                # fail if any commands fails
-                set -e
-                # debug log
-                set -x
-
-                RELEASE_DESCRIPTION=$(printf "$(npm run --silent release-notes | tail -n +3)\n\n### [Known Issues](KNOWN_ISSUES.md)\n")
-                envman add --key RELEASE_DESCRIPTION --value "$RELEASE_DESCRIPTION"
       - github-release@0.11.0:
           inputs:
             - username: $GITHUB_USERNAME
             - name: $BITRISE_GIT_TAG
             - draft: "no"
             - files_to_upload: $RELEASE_APK_PATH
-            - body: $RELEASE_DESCRIPTION
+            - body: "See [changelog](https://github.com/digidem/mapeo-mobile/blob/develop/CHANGELOG.md) for a detailed list of changes"
             - api_token: $GITHUB_TOKEN
       - google-play-deploy@3.0.2:
           inputs:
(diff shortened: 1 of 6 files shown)


patches/nodejs-mobile-react-native+0.6.3.patch
```diff
@@ -25,6 +25,15 @@ index e54f25c..178b15a 100644
          main.assets.srcDirs += '../install/resources/nodejs-modules'
      }
  
+@@ -239,7 +242,7 @@ if ("1".equals(shouldRebuildNativeModules)) {
+     GenerateNodeProjectAssetsLists.dependsOn "ApplyPatchScriptToModules"
+ 
+     def nativeModulesABIs = android.defaultConfig.ndk.abiFilters;
+-    if (nativeModulesABIs == null) {
++    if (!nativeModulesABIs) {
```
achou11, line 33: somewhat lazy approach as null and empty collections in groovy are both falsy http://groovy-lang.org/semantics.html#the-groovy-truth

bitrise.yml
```diff
@@ -192,25 +192,12 @@ workflows:
...
-
-                RELEASE_DESCRIPTION=$(printf "$(npm run --silent release-notes | tail -n +3)\n\n### [Known Issues](KNOWN_ISSUES.md)\n")
-                envman add --key RELEASE_DESCRIPTION --value "$RELEASE_DESCRIPTION"
       - github-release@0.11.0:
           inputs:
             - username: $GITHUB_USERNAME
             - name: $BITRISE_GIT_TAG
             - draft: "no"
             - files_to_upload: $RELEASE_APK_PATH
-            - body: $RELEASE_DESCRIPTION
```
gmaclennan, line 213: Maybe add a link to the change log here?

patches/nodejs-mobile-react-native+0.6.3.patch
gmaclennan: Huh, well that makes it quite easy. I like the idea of "groovy truth"

gmaclennan (approved):
Looks good. When you have some time (non-urgent) it would be good to submit a PR to nodejs-mobile-react-native.

achou11:
@gmaclennan yeah I created an issue there already (https://github.com/nodejs-mobile/nodejs-mobile-react-native/issues/11)

after addressing all feedback is it worth just merging this? not sure we have any more build minutes on bitrise so was gonna wait until we can discuss tomorrow

bitrise.yml
achou11: done 👍

gmaclennan (approved):

Issue #937, opened 2022-03-09
Add Background Map (via file)


ximenabb:
**Description** User clicked on add Background Map, User can import from file

**Screens**
Add Background Map
Import from File (.jp2,MV, GPX., PIX)]
Close

**Copydeck**
https://docs.google.com/spreadsheets/d/1Eco_G5hmAqnWW5NfzpAeeMgn8L5c61VO_PnkU7mSIRI/edit#gid=191621157

**Design**
https://www.figma.com/file/sxz2NvysPVab6ZTqs6f57K/Cobra-Collective

***

**Description** User clicked on add via import file. User can delete the file after import.

**Screens**
1- Import from file
Background Map Name
Import File
Accepted file types are XXXX

2- Import from file
Background Map Name
Import File
Accepted file types are XXXX
Delete

**Copydeck**
https://docs.google.com/spreadsheets/d/1Eco_G5hmAqnWW5NfzpAeeMgn8L5c61VO_PnkU7mSIRI/edit#gid=191621157
**Design**
https://www.figma.com/file/sxz2NvysPVab6ZTqs6f57K/Cobra-Collective?node-id=314%3A2177

ErikSin:
To ask in team meeting, what file extension?

Issue #936, opened 2022-03-09
Background Maps Screen


ximenabb:
**Description** 
User was on Map Settings and clicked on Background Maps:
-  user doesn’t have offline areas yet
- user has offline areas

**Screens**
1- No offline areas
Background Maps
Add background map
(Current Map)
No downloaded offline areas

2- has offline areas
No offline areas
Background Maps
Add background map
(Current Map)

**Copydeck**
https://docs.google.com/spreadsheets/d/1Eco_G5hmAqnWW5NfzpAeeMgn8L5c61VO_PnkU7mSIRI/edit#gid=191621157

**Design**
https://www.figma.com/file/sxz2NvysPVab6ZTqs6f57K/Cobra-Collective?node-id=183%3A751

ErikSin:
To ask Sabella, place holder images for now?

Issue #935, opened 2022-03-07
Settings


ximenabb:
**Description**
User access the map management screen via Settings

**Screens**
1/
Settings
XX
XX
XX
Map Settings
XXX

2/
Map Settings
Directional Arrow
Background Maps
Data Layers

**Copydeck**
https://docs.google.com/spreadsheets/d/1Eco_G5hmAqnWW5NfzpAeeMgn8L5c61VO_PnkU7mSIRI/edit#gid=207943684
**Design**
https://www.figma.com/file/sxz2NvysPVab6ZTqs6f57K/Cobra-Collective?node-id=183%3A663

PR #946, opened 2022-04-04
Feat: Adds map setting screens

### LINKED ISSUE #940: Delete Map
**Description** User clicked on a bg map, sees list of offline areas, wants to delete map

**Screens**
Are you sure you want to remove XXXX
This map and offline areas attached to it will be deleted. This cannot be undone
Remove Map
Cancel

**Copydeck**
https://docs.google.com/spreadsheets/d/1Eco_G5hmAqnWW5NfzpAeeMgn8L5c61VO_PnkU7mSIRI/edit#gid=1850696440

**Design**
https://www.figma.com/file/sxz2NvysPVab6ZTqs6f57K/Cobra-Collective

ErikSin:
# Map Settings and Map Background
User can pull access map settings from setting screen.

`Map background` will list all available map backgrounds. User will be able to click into each map background and delete map background.

This does not have any functionality, purely screens

Kapture 2022-04-04 at 19 33 30
 
Closes: #940, #935, #936, and #937


src/frontend/lib/styles.js
```diff
@@ -2,7 +2,7 @@
 
 export const BLACK = "#000000";
 export const WHITE = "#ffffff";
-export const MAPEO_BLUE = "#3366ff";
+export const MAPEO_BLUE = "#0066FF";
```
ErikSin, line 5: This is the updated mapeo blue according to the design specs

achou11: was curious about what + how much this would affect but the difference in the color is pretty limited, so not as concerned

messages/en.json
```diff
@@ -1039,6 +1083,19 @@
...
     "message": "Unexpected Error"
   },
+  "sharedComponent.OfflineAreaCard.lvlDetail": {
+    "message": "Level of Detail"
+  },
+  "sharedComponent.OfflineAreaCard.zoomLevel": {
+    "message": "Zoom Level"
+  },
+  "sharedComponents.BGMapCard.abbrevMegaBite": {
+    "description": "The abbreviation for mega bite",
```
achou11:
```suggestion
  "sharedComponents.BGMapCard.abbrevMegabyte": {
    "description": "The abbreviation for megabyte",
```

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -0,0 +1,180 @@
...
+  return (
+    <React.Fragment>
+      <ScrollView style={styles.container}>
+        <Button style={[styles.button]} variant="outlined" onPress={openSheet}>
+          {t(m.addBGMap)}
+        </Button>
+
+        {/* Default BG map card */}
+        <BGMapCard
+          mapId="D3f4U1t"
```
achou11, line 90: probably fine to make this something readable e.g. `mapeo-default`

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
```diff
@@ -0,0 +1,179 @@
...
+    id: "screens.Settings.MapSettings.removeMap",
+    defaultMessage: "Remove Map",
+  },
+  cancel: {
+    id: "screens.Settings.MapSettings.cancel",
+    defaultMessage: "Cancel",
+  },
+  clearDiagrams: {
+    id: "screens.Settings.MapSettings.clearDiagrams",
+    defaultMessage: "Are you sure you want to Clear diagrams",
```
achou11:
```suggestion
    defaultMessage: "Are you sure you want to clear diagrams?",
```

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
```diff
@@ -0,0 +1,179 @@
...
+  },
+  subtitle: {
+    id: "screens.Settings.MapSettings.subtitle",
+    defaultMessage:
+      "This map and offline areas attached to it will be deleted. This cannot be undone",
```
achou11, line 38:
```suggestion
      "This map and offline areas attached to it will be deleted. This cannot be undone.",
```

achou11: not entirely sure about the use of "attached" here but i think it's generally fine. just wanted to point out in case others had thoughts

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  const [listOfflineAreas, setListOfflineAreas] = React.useState<
+    OfflineArea[]
+  >();
+
+  const { getParam } = navigation;
+
+  React.useEffect(() => {
+    //Typesafety?
+    bgMapId.current = getParam("mapId", "");
```
achou11: context for this comment? is TS showing some kind of warning?

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
```diff
@@ -0,0 +1,179 @@
...
+          id: "2",
+          title: "Offline Area 2",
+          zoomLevel: 44,
+          levelOfDetail: "Small Roads",
+        },
+      ];
+    }
+
+    setListOfflineAreas(getAllOfflineAreas(bgMapId.current));
+  }, []);
```
achou11: do things get weird if you add `getParam` to this dep array? mostly just curious and wouldn't be surprised if omitting is the way to go

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
```diff
@@ -0,0 +1,179 @@
...
+  navigation,
+}) => {
+  const bgMapId = React.useRef("");
+  const { formatMessage: t } = useIntl();
+
+  const { closeSheet, isOpen, openSheet, sheetRef } = useBottomSheetModal({
+    openOnMount: false,
+  });
+
+  const [listOfflineAreas, setListOfflineAreas] = React.useState<
```
achou11: nit but would rename to sound less like an action e.g. `offlineAreas` or `offlineAreasList`

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+    setListOfflineAreas(getAllOfflineAreas(bgMapId.current));
+  }, []);
+
+  return (
+    <React.Fragment>
+      <ScrollView style={[styles.container]}>
+        {listOfflineAreas === undefined ? (
+          <Loading />
+        ) : (
```
achou11: at a glance (could be wrong based on how this component is navigated to), but may need to represent three states here:

1. No offline areas
2. Loading offline areas
3. Showing offline areas

src/frontend/screens/Settings/Settings.js
```diff
@@ -99,6 +99,11 @@ const m = defineMessages({
     defaultMessage: "App Passcode and Device Security",
     description: "Description of security button in settings",
   },
+  mapSettings: {
+    id: "screens.Settings.mapSettings",
+    defaultMessage: "Map Settings",
+    description: "Description of map settings button in settings",
```
achou11, line 103: nit but this should highlight that it's the option's title (as opposed to the subtitle/description)

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
ErikSin: This is going to change and be dynamically added, so its just a place holder for now

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
ErikSin: This is from the copy deck

achou11: yeah i'm aware of that, but my comment still stands 😄

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
ErikSin: No warning, just flagging that there is no typesafety and annoying complicated given that it is getting the nav item comes from the parent

achou11: ah got it. yeah there's a way to make it more type-safe but not worth looking into for this (think it's improved in more recent version of react navigation)

src/frontend/sharedComponents/BGMapCard.tsx
```diff
@@ -0,0 +1,127 @@
...
+    setZoomLevel(getZoomLevel());
+    setCenterCoordinate(getCenterCoordinate());
+  }, []);
+
+  return (
+    <TouchableOpacity onPress={onPress || onPressDefault}>
+      <View style={[styles.container, style]}>
+        <MapboxGL.MapView
+          // placeholder style URL
+          styleURL={styleUrl}
```
achou11, line 85: for the default style, you can use the enum provided by the MapboxGL library: `MapboxGL.StyleURL.Streets`

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
ErikSin: No, there is always an offline area in the current flow (you cant download a background map without offline areas for the current MVP, but down the line we will need that). At the moment Sabella has not made the design for no offline area, but we could potentially add it to the copy deck

src/frontend/sharedComponents/BGMapCard.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+const m = defineMessages({
+  currentMap: {
+    id: "sharedComponents.BGMapCard.currentMap",
+    defaultMessage: "Current Map",
+  },
+  abbrevMegaBite: {
+    id: "sharedComponents.BGMapCard.abbrevMegaBite",
+    defaultMessage: "MB",
+    description: "The abbreviation for mega bite",
```
achou11, line 21:
```suggestion
    id: "sharedComponents.BGMapCard.abbrevMegabyte",
    defaultMessage: "MB",
    description: "The abbreviation for megabyte",
```

achou11:
```suggestion
  abbrevMegabyte: {
    id: "sharedComponents.BGMapCard.abbrevMegabyte",
    defaultMessage: "MB",
    description: "The abbreviation for megabyte",
```

src/frontend/sharedComponents/BGMapCard.tsx
```diff
@@ -0,0 +1,127 @@
...
+        >
+          <MapboxGL.Camera
+            zoomLevel={zoomLevel}
+            centerCoordinate={centerCoordinate}
+          />
+        </MapboxGL.MapView>
+        <View style={[styles.textContainer]}>
+          <Text style={[styles.text, { fontWeight: "bold" }]}>{mapTitle}</Text>
+          <Text style={[styles.text]}>
+            {mapSize.toString() + t(m.abbrevMegaBite)}
```
achou11:
```suggestion
            {mapSize.toString() + t(m.abbrevMegabyte)}
```

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
ErikSin: yeah its much easier in react nav 6 (i believe they wrote nav 6 in ts)...right now its a whole bunch of boillerplate code and it makes it really confusing because it looks like it is the params for the parent component (because you cant use the hook to pass params)

src/frontend/sharedComponents/BGMapCard.tsx
```diff
@@ -0,0 +1,127 @@
...
+          compassEnabled={false}
+          zoomEnabled={false}
+          logoEnabled={false}
+          scrollEnabled={false}
+          style={[styles.map]}
+        >
+          <MapboxGL.Camera
+            zoomLevel={zoomLevel}
+            centerCoordinate={centerCoordinate}
+          />
```
achou11: think we want to disable the animation here, via `animationDuration={0}` and (maybe, but not as necessary?) `animationMode="linearTo"`. curious about the `allowUpdates` prop too as it might be something we'd want to set to `false` for this component

src/frontend/sharedComponents/BottomSheetModal/Backdrop.tsx
```diff
@@ -9,25 +9,24 @@ export const Backdrop = ({
...
 }: BottomSheetBackdropProps) => {
-  const animatedOpacity = React.useMemo(
-    () =>
-      interpolate(animatedIndex, {
-        inputRange: [0, 1],
-        outputRange: [0, 0.3],
-        extrapolate: Extrapolate.CLAMP,
-      }),
-    [animatedIndex]
+  //Use memo was causing a flickering effect when the backdrop was opened.
```
achou11: interesting. slightly worried that removing this might break something else (i think i took this from the library's example), so worth testing other components that use this if possible

src/frontend/lib/styles.js
ErikSin: Yeah, its very slight, that it won't be noticeable to the naked eye

image

src/frontend/sharedComponents/BottomSheetModal/Backdrop.tsx
achou11: The implementation is an adapted version of this example for animating the backdrop background: https://github.com/gorhom/react-native-bottom-sheet/blob/v2/example/src/components/customBackground/CustomBackground.tsx

i have a feeling that the changes here might break that but if you can verify that it doesn't then works for me

src/frontend/sharedComponents/BottomSheetModal/Backdrop.tsx
```diff
@@ -9,25 +9,24 @@ export const Backdrop = ({
...
     () => [
       style,
       {
         backgroundColor: BLACK,
-        opacity: animatedOpacity,
+        opacity: animatedOpacity.current,
       },
     ],
-    [style, animatedOpacity]
+    [style, animatedOpacity.current]
```
achou11: seems odd, but at least based on the example, i _think_ we want to keep this as the ref and not the current value of the ref since it's an animated value, which has weird rules and whatnot that i don't fully understand tbh. normally passing a ref in the dep array is not a good idea, so maybe this change is onto something (perhaps just removing from the dep array altogether?)

worth just qa-ing these changes overall since it's pretty sensitive to such changes 😅

ErikSin: I suppose it doesn't hurt, given that `.current` is a subset of the object. Animations are so finicky....

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
```diff
@@ -0,0 +1,179 @@
...
+    bgMapId.current = getParam("mapId", "");
+
+    //To Do Api call to get offline areas
+    function getAllOfflineAreas(mapId: string): OfflineArea[] {
+      return [
+        {
+          id: "1",
+          title: "Offline Area 1",
+          zoomLevel: 44,
+          levelOfDetail: "Small Roads",
```
achou11: is the idea that this field comes from the backend? i think this should be fully handled client-side and determined based on the zoom level that's retrieved

src/frontend/sharedComponents/OfflineAreaCard.tsx
```diff
@@ -0,0 +1,59 @@
...
+  zoomLevel: {
+    id: "sharedComponent.OfflineAreaCard.zoomLevel",
+    defaultMessage: "Zoom Level",
+  },
+});
+
+interface OfflineAreaCardProps {
+  title: string;
+  zoomLevel: number;
+  lvlOfDetail?: string;
```
achou11: based on https://github.com/digidem/mapeo-mobile/pull/946/files#r843175295, i think that we should generate the appropriate string based on the `zoomLevel` so basically have a map of zoom level ranges to a particular description

src/frontend/sharedComponents/BottomSheetModal/Backdrop.tsx
achou11: a h wait i see what's happening. you converted that value to a ref from a memoized animated value. I _think_ that's okay but then i think we may need to update the `animatedOpacity` value if the `animatedIndex` updates, which is why i used the memo. does the flickering still happen if you add this effect and remove the ref from the deps array here?
 
 ```jsx
 React.useEffect(() => {
   animatedOpacity.current = interpolate(animatedIndex, {
     // omitted for brevity
   })
 }, [animatedIndex])

achou11: alternatively, maybe just simplifying to something closer to the example would work? i can't remember if the extraction was done due to visual bugs or something 😅 e.g. something along the lines of this and removing the `animatedOpacity` value altogether

```js
  const containerStyle = React.useMemo(
    () => [
      style,
      {
        backgroundColor: BLACK,
        opacity: interpolate(animatedIndex, {
          inputRange: [0, 1],
          outputRange: [0, 0.3],
          extrapolate: Extrapolate.CLAMP,
        }),
      },
    ],
    [style, animatedIndex]
  );
```

ErikSin: i tried these, it did work. I dont want to spend too much time on this animation. so I am just turning it off and adding it to the backlog.

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -0,0 +1,180 @@
...
+              mapId={bgMap.mapId}
+              navigation={navigation}
+              mapSize={bgMap.size}
+              mapTitle={bgMap.title}
+            />
+          ))
+        )}
+      </ScrollView>
+
+      <BackgroundMapModal closeSheet={closeSheet} sheetRef={sheetRef} />
```
achou11: Might be easier to just inline this component here instead of creating a new one because you'll have to forward the `sheetRef` via `React.forwardRef` (as opposed to passing the prop you have here)

src/frontend/screens/Settings/MapSettings/OfflineAreas.tsx
```diff
@@ -0,0 +1,173 @@
...
+            />
+          ))
+        )}
+      </ScrollView>
+
+      <Button style={styles.button} onPress={openSheet}>
+        {t(m.removeMap)}
+      </Button>
+
+      <DeleteAreaModal closeSheet={closeSheet} sheetRef={sheetRef} />
```
achou11: Same comment as [this](https://github.com/digidem/mapeo-mobile/pull/946/files#r845363208). Either inline the component here or update the component to use `forwardRef`

src/frontend/sharedComponents/BGMapCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+    function getZoomLevel() {
+      //To do: API call to get zoom level
+      //This should be min zoom. Where is this coming from?
+      return 6;
+    }
+    function getCenterCoordinate() {
+      //To do: API call to get center coordinate
+      //I believe this should come from the styleJSON `center` property
+      return [position?.coords.longitude, position?.coords.latitude];
+    }
```
achou11: with the mapserver, this will all be a single api call that retrieves the offline stylejson, which contains the information about the zoom level and center coordinate too. You'd then pass this style into the `styleJSON` prop of the map

src/frontend/sharedComponents/BGMapCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+    setZoomLevel(getZoomLevel());
+    setCenterCoordinate(getCenterCoordinate());
+  }, []);
+
+  return (
+    <TouchableOpacity onPress={onPress || onPressDefault}>
+      <View style={[styles.container, style]}>
+        <MapboxGL.MapView
+          // placeholder style URL
+          styleURL={styleUrl}
```
achou11, line 85: See [comment](https://github.com/digidem/mapeo-mobile/pull/946/files#r845366546). WIth the mapserver, we want to fetch the stylejson and pass that into the `styleJSON` prop instead of the `styleURL` prop

src/frontend/sharedComponents/BGMapCard.tsx
```diff
@@ -0,0 +1,130 @@
...
+          compassEnabled={false}
+          zoomEnabled={false}
+          logoEnabled={false}
+          scrollEnabled={false}
+          style={[styles.map]}
+        >
+          <MapboxGL.Camera
+            zoomLevel={zoomLevel}
+            centerCoordinate={centerCoordinate}
```
achou11, line 94: these two will be derived from the fetched stylejson

src/frontend/sharedComponents/BottomSheetModal/Backdrop.tsx
```diff
@@ -13,7 +13,7 @@ export const Backdrop = ({
     () =>
       interpolate(animatedIndex, {
         inputRange: [0, 1],
-        outputRange: [0, 0.3],
+        outputRange: [0.1, 0.3],
```
achou11: i don't think we want to change this. having this would cause the backdrop to always be visible at 0.1 opacity, even if the sheet is closed

src/frontend/sharedComponents/BottomSheetModal/index.tsx
```diff
@@ -66,7 +66,7 @@ const useBackPressHandler = (onHardwareBackPress?: () => void) => {
 };
 
 const useSnapPointsCalculator = () => {
-  const [sheetHeight, setSheetHeight] = React.useState(0);
+  const [sheetHeight, setSheetHeight] = React.useState(-1);
```
achou11, line 70: Probably should write a comment about the context for this cursed value 😅 can do as a follow up but for posterity:

Having a value of 0 caused a flickering of the backdrop whenever the sheet was initially opened. This only occurs when the component is used within a component (as opposed to a standalone navigation screen). Switching it to a positive number causes the `onLayout` event in the content to report an inaccurate height value, which causes the sheet size to be miscalculated, affecting the content within the sheet (e.g. squishing, overflow, etc).

src/frontend/sharedComponents/OfflineAreaCard.tsx
```diff
@@ -0,0 +1,161 @@
+import * as React from "react";
+import { defineMessages, FormattedMessage, useIntl } from "react-intl";
+import { StyleSheet, View } from "react-native";
+import { LIGHT_GREY } from "../lib/styles";
+import Text from "./Text";
+
+//Lev
```
achou11: remove?

src/frontend/sharedComponents/OfflineAreaCard.tsx
```diff
@@ -0,0 +1,161 @@
...
+  },
+  lvlDetailVillage: {
+    id: "sharedComponent.OfflineAreaCard.lvlDetailVillage",
+    defaultMessage: "Village",
+    description: "Level of detail seen on map - can see villages on map",
+  },
+  lvlDetailSmallRoad: {
+    id: "sharedComponent.OfflineAreaCard.lvlDetailSmallRoad",
+    defaultMessage: "Small Road",
+    description: "Level of detail seen on map - cab see small roads on map",
```
achou11:
```suggestion
    description: "Level of detail seen on map - can see small roads on map",
```

src/frontend/sharedComponents/OfflineAreaCard.tsx
```diff
@@ -0,0 +1,161 @@
...
+  lvlDetailSmallCountry: {
+    id: "sharedComponent.OfflineAreaCard.lvlDetailSmallCountry",
+    defaultMessage: "Small Country",
+    description: "Level of detail seen on map - can see smaller countries",
+  },
+  lvlDetailLargeMetro: {
+    id: "sharedComponent.OfflineAreaCard.lvlDetailLargeMetro",
+    defaultMessage: "Large Metropolitan Area",
+    description:
+      "Level of detail seen on map - Can see details of large metropolitan areas",
```
achou11: for the sake of consistency: 

```suggestion
      "Level of detail seen on map - can see details of large metropolitan areas",
```

src/frontend/sharedComponents/OfflineAreaCard.tsx
```diff
@@ -0,0 +1,161 @@
...
+  },
+  lvlDetailCity: {
+    id: "sharedComponent.OfflineAreaCard.lvlDetailCity",
+    defaultMessage: "City",
+    description: "Level of detail seen on map - Can see most Cities on Map",
```
achou11: for consistency's sake:

```suggestion
    description: "Level of detail seen on map - can see most cities on map",
```

src/frontend/sharedComponents/OfflineAreaCard.tsx
```diff
@@ -0,0 +1,161 @@
...
+}
+
+export const OfflineAreaCard = ({
+  title,
+  zoomLevel,
+  position,
+}: OfflineAreaCardProps) => {
+  const { formatMessage: t } = useIntl();
+
+  const lvlOfDetail = React.useMemo(() => {
```
achou11, line 99: the memoization is probably not necessary here. would also extract this into a function outside the component

achou11: hm i guess a little bit more complicated to extract since it relies on the format message function. maybe extract the logic so that it returns the appropriate message key name e.g.

```jsx
// Not the best name but you get the idea
function getLevelOfDetailKey(zoomLevel: number) {
  switch (true) {
    case zoomLevel <= 2:
      return 'lvlDetailGlobal';
    // ...
  }
}

const Component = () => {
  // ...
  const lvlOfDetail = t(m[getLevelOfDetailKey(zoomLevel)])
  // ... 
}

File: src/frontend/sharedComponents/BGMapCard.tsx (hunk 21 above)
Comment (ErikSin) [reply src/frontend/sharedComponents/BGMapCard.tsx]: thats fine, I think we can deal with this when the api is built!

File: src/frontend/sharedComponents/BottomSheetModal/Backdrop.tsx (hunk 24 above)
Comment (ErikSin) [reply src/frontend/sharedComponents/BottomSheetModal/Backdrop.tsx]: right, i was just messing around, ill put this back

**Comment by achou11** [comment]
what's the plan for integrating this PR given that none of the API is available yet? Would it make sense to create a feature flag or was the plan to wait for the API to be built into the app before merging this?

File: src/frontend/sharedComponents/OfflineAreaCard.tsx (hunk 30 above)
Comment (ErikSin) [reply src/frontend/sharedComponents/OfflineAreaCard.tsx:99]: I think memoizing just optimizes it here since there are so many values.

**Comment by ErikSin** [comment]
> what's the plan for integrating this PR given that none of the API is available yet? Would it make sense to create a feature flag or was the plan to wait for the API to be built into the app before merging this?

good question, maybe lets discuss it today in the meeting. There are an increasing number of feature flags that i think it is becoming unmanageable.

File: src/frontend/sharedComponents/BGMapCard.tsx (hunk 22 above)
Comment (ErikSin) [reply src/frontend/sharedComponents/BGMapCard.tsx:85]: right, im going to leave this for now, mainly because i would have to build an entire styleJson as a dummy JSON, and this gives me the visual we need for just the screen!

File: src/frontend/AppContainerWrapper.tsx [hunk 31]
```diff
@@ -21,32 +21,40 @@ import bugsnag from "./lib/logger";
...
 
 // WARNING: This needs to change if we change the navigation structure
-const getNavStoreKey = (includeOnboarding: boolean) =>
-  `@MapeoNavigation@${includeOnboarding ? 9 : 8}`;
-
-const createNavigationStatePersister = (includeOnboarding: boolean) => async (
-  navState: NavigationState
-) => {
+//Dev experiments and include onbooarding cannot be turned on at the same time (as discussed with the whole team)
+const getNavStoreKey = (includeOnboarding: boolean, devExperiments: boolean) =>
```
achou11: Not sure if it will be an issue in reality, but i think this will be hard to maintain if, for example, onboarding-specific navigation changes. In that case, we'd have to increment the onboarding key to be `11` instead of `10` i.e. we'd have to remember to increment the version by 2 instead of 1 because of the variations, and for each variation

I wonder if it would make help to namespace the navigation keys based on the flags so basically starting new keys for the two flags. Thinking of something along these lines:

```ts
// If no variant is provided, default to the original nav key. Can provide a explicit key to specify this if desired
const getNavStoreKey = (variant?: 'DevExperiments' | 'Onboarding') {
  switch (variant) {
    case 'Onboarding':
      // Since this is namespaced, we can start at 1, and then increment this whenever onboarding-specific navigation changes
      return '@MapeoNavigation-Onboarding@1'
    case 'DevExperiments':
       // Since this is namespaced, we can start at 1, and then increment this whenever dev experiments specific navigation changes
      return '@MapeoNavigation-DevExperiments@1'
    default:
      // Maybe starting this at 10 would be the safer thing, since 9 was previously occupied by my intro of onboarding navgiation
      return  '@MapeoNavigation@10'
   }
  }
}
```

Not sure if there are unforeseen consequences of doing this - any thoughts?

achou11 (commented):
Added a potential suggestion to make navigation key management a little less prone to error. Would love to get your eyes on that before approving! (which i'd otherwise do 👍)

src/frontend/AppContainerWrapper.tsx
ErikSin: Yeah I think i prefer your way of name-spacing here. This does not solve the problem of if we mix these states, but definitely more elegant then nested ternaries.

achou11: yeah that's true. in theory maybe could combine them so it'd be something like `@MapeoNavigation-Onboarding-DevExperiments@1` but i think that's not necessary to worry about for now 😅

ErikSin: I think its clearer now, but im not 100% sure if thats true. I have been staring at this code for too long  😅

achou11 (approved):
great stuff! happy that we've made some improvements to handling feature flags with this too 👍

src/frontend/sharedComponents/BottomSheetModal/Backdrop.tsx
achou11: issue resolved by https://github.com/digidem/mapeo-mobile/pull/946/commits/976099d03e442449227918da50300797f6abe459

src/frontend/AppContainerWrapper.tsx
```diff
@@ -11,42 +11,39 @@ import {
...
+    await AsyncStorage.setItem(NAV_STORE_KEY, JSON.stringify(navState));
   } catch (err) {
     log("Error saving navigation state", err);
   }
 };
 
-const loadSavedNavState = async (includeOnboarding: boolean) => {
+const loadSavedNavState = async () => {
   try {
+    if (featureFlagOn) return null;
```
achou11, line 44: nit: move this outside the try statement since we don't want to potentially catch reference errors (i.e. `featureFlagOn` is not defined)

src/frontend/AppContainerWrapper.tsx
```diff
@@ -138,26 +135,23 @@ const AppContainerWrapper = () => {
...
 
               return loadedNavState;
             },
-            persistNavigationState: createNavigationStatePersister(
-              experiments.onboarding
-            ),
+            persistNavigationState: createNavigationStatePersister(),
           },
-    [experiments.onboarding, updateRouteBasedAppState]
+    [onboarding, updateRouteBasedAppState]
```
achou11, line 148: can `onboarding` be removed here? doesn't look like this effect uses it anymore

src/frontend/constants.ts
```diff
@@ -1,5 +1,6 @@
 export const URI_PREFIX = "mapeo://";
 export const ERROR_STORE_KEY = "@MapeoError";
+export const NAV_STORE_KEY = "@MapeoNav";
```
achou11, line 3: As long as we're persisting navigation state, it's probably safer to preserve the version number (8) here (thinking about nav changes that are added later on that aren't related to experiments or feature flags)

src/frontend/context/SettingsContext.tsx
```diff
@@ -14,32 +14,31 @@ export type ExperimentalP2pUpgrade = boolean;
 
 export type SettingsState = {
   coordinateFormat: CoordinateFormat;
+  // Experiments should only include experiments that are enabled in the app. AKA only experiments that can be turned on and off by the user
```
achou11, line 17: nice, this is a helpful comment 👍

src/frontend/context/SettingsContext.tsx
```diff
@@ -54,12 +53,30 @@ export const SettingsProvider = ({ children }: React.PropsWithChildren<{}>) => {
     [setState]
   );
 
+  /**
+   * @typedef {SettingsContextType[2]} setExperiment
+   * @param key The key of the experiment to toggle
+   * @param value The value to set the experiment to. If not provided the value will be toggled
+   * @description Sets the value of the experiment to the provided value or toggles it if no value is provided
+   */
+  const setExperiments: SettingsContextType[2] = React.useCallback(
```
achou11, line 62: just so i'm understanding this, this is so that you can toggle experiments more granularly, as opposed to having to work at the top-level settings?

src/frontend/context/SettingsContext.tsx
```diff
@@ -54,12 +53,30 @@ export const SettingsProvider = ({ children }: React.PropsWithChildren<{}>) => {
     [setState]
   );
 
+  /**
+   * @typedef {SettingsContextType[2]} setExperiment
+   * @param key The key of the experiment to toggle
+   * @param value The value to set the experiment to. If not provided the value will be toggled
+   * @description Sets the value of the experiment to the provided value or toggles it if no value is provided
+   */
```
achou11, line 61: Curious why the JS doc annotation here?

src/frontend/context/SettingsContext.tsx
```diff
@@ -54,12 +53,30 @@ export const SettingsProvider = ({ children }: React.PropsWithChildren<{}>) => {
...
+  );
+
   const contextValue: SettingsContextType = React.useMemo(() => {
     // If we add any new properties to the settings state, they will be
     // undefined in a users' persisted state, so we merge in the defaults
     const mergedState = merge({}, DEFAULT_SETTINGS, state);
-    return [mergedState, setSettings];
-  }, [state, setSettings]);
+    return [mergedState, setSettings, setExperiments];
+  }, [state, setSettings, DEFAULT_SETTINGS]);
```
achou11, line 79: `DEFAULT_SETTINGS` doesn't need to be in the deps array

src/frontend/context/SettingsContext.tsx
achou11: suggestion: move this to `useExperiments` hook, unless there are places that aren't using the hook and instead using this context directly

src/frontend/lib/DevExperiments.ts
```diff
@@ -0,0 +1,16 @@
...
+const checkFeatureFlagOn = (featureObject: { [key: string]: boolean }) => {
+  for (const k in featureObject) {
+    if (featureObject[k]) {
+      return true;
+    }
+  }
+  return false;
+};
+
+export const featureFlagOn = checkFeatureFlagOn(devExperiments);
```
achou11, line 16: This feels a little more verbose than necessary. Would make this a one-liner:

- using an explicit "or" condition i.e. `devExperiments.onboarding || devExperiments.appPasscode || ...`

- using `Array.some` i.e. `Object.values(devExperiments).some(flag => flag === true)`

src/frontend/screens/AppPasscode/index.tsx
```diff
@@ -21,12 +22,12 @@ const m = defineMessages({
...
   const { passIsSet } = React.useContext(SecurityContext);
-  const [{ experiments }] = React.useContext(SettingsContext);
   const { navigate } = useNavigation();
+  const { appPasscode } = devExperiments;
 
   React.useEffect(() => {
-    if (!experiments.appPasscode) navigate("Settings");
-  }, [experiments]);
+    if (!appPasscode) navigate("Settings");
+  }, [appPasscode]);
```
achou11, line 30: this won't change in the React lifecycle so it shouldn't need to be specified in the deps array. i think something like the following would suffice?

```js
React.useEffect(() => {
  if (!devExperiments.appPasscode) navigate("Settings")
}, [])

File: src/frontend/AppContainerWrapper.tsx [hunk 41]
```diff
@@ -96,9 +93,9 @@ const AppContainerWrapper = () => {
   const navRef = React.useRef<NavigationContainerComponent>();
   const [inviteModalEnabled, setInviteModalEnabled] = React.useState(true);
   const [queuedInvite, setQueuedInvite] = React.useState<string | null>(null);
-  const [{ experiments }] = React.useContext(SettingsContext);
   const [hidePracticeBar, setHidePracticeBar] = React.useState(true);
   const [hidePracticeMode, setHidePracticeMode] = React.useState(false);
+  const { onboarding } = devExperiments;
```
achou11, line 98: nit: i _think_ since this is a stable constant value, you can define this outside the React component. Can't fully remember, but extracting it within the component may have rendering implications, which can be avoided since it's a stable constant

src/frontend/screens/Security/index.tsx
```diff
@@ -66,12 +67,12 @@ const m = defineMessages({
...
   const { passIsSet } = React.useContext(SecurityContext);
-  const [{ experiments }] = React.useContext(SettingsContext);
   const { navigate } = useNavigation();
+  const { appPasscode } = devExperiments;
 
   React.useEffect(() => {
-    if (!experiments.appPasscode) navigate("Settings");
-  }, [experiments]);
+    if (!appPasscode) navigate("Settings");
+  }, [appPasscode]);
```
achou11, line 75: similar comment about not needing to specify `appPasscode` in the deps array, since it's a stable constant

src/frontend/context/SettingsContext.tsx
achou11: nit: the `set` in the naming feels a little unclear when actually using this for toggling a value. calling this via `setExperiments("name_of_experiment")` doesn't give a clear idea of _what_ i'm setting it to. personally try to avoid functions that just toggle state because of this lack of clarity when being called

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -0,0 +1,165 @@
...
+    const results = await DocumentPicker.getDocumentAsync();
+
+    if (results.type === "cancel") {
+      closeSheet();
+      return;
+    }
+
+    if (results.type === "success") {
+      // To do API call to import map
+      console.log(results.uri);
```
achou11, line 76: remove?

src/frontend/screens/Settings/ProjectConfig/index.tsx
```diff
@@ -62,8 +63,8 @@ export const ProjectConfig = () => {
   const [role] = React.useState<"participant" | "coordinator">("participant");
   const [status, setStatus] = React.useState<Status>("idle");
   const [config, { replace: replaceConfig }] = React.useContext(ConfigContext);
-  const [{ experiments }] = React.useContext(SettingsContext);
 
+  const { onboarding } = devExperiments;
```
achou11, line 67: can extract this value outside of the component since it's a stable constant

src/frontend/screens/Settings/Settings.tsx
```diff
@@ -99,12 +97,16 @@ const m = defineMessages({
...
+    defaultMessage: "Map Settings",
+    description: "Description of map settings button in settings",
+  },
 });
 
 const Settings = () => {
   const { navigate } = useNavigation();
-  const [{ experiments }] = useContext(SettingsContext);
-
+  const { appPasscode, mapSettings } = devExperiments;
```
achou11, line 109: can extract outside of component

achou11 (commented):
A lot easier to follow now that experiments vs feature-flags (or dev experiments) are more clearly defined 👍

src/frontend/AppContainerWrapper.tsx
achou11: also it's not needed in general since it's a stable const value

src/frontend/AppContainerWrapper.tsx
ErikSin: Right, i think ive just been stuck in the react world for so long, i forget how normal js works!

src/frontend/context/SettingsContext.tsx
ErikSin: I was curious about getting intellisense to pick up the description because i do think it is an odd use case, And there is the option to not set a value, so i just thought it would be useful to have that info while coding...But I don't know how to do that with react and jsdoc because this did not work

src/frontend/context/SettingsContext.tsx
ErikSin: I addressed the explicit settings boolean, by making it a mandatory parameter

ErikSin: Moved the logic to the hook

PR #947, opened 2022-04-07
fix: only support light theme

achou11:
### Notes:

- Via the [RN Diff helper tool](https://github.com/react-native-community/rn-diff-purge/blob/2b62aafb4ac53c02d6747dff571ddfca11354716/RnDiffApp/android/app/src/main/res/values/styles.xml#L4), I introduced a change that basically added support for theme colors to be dynamically switched based on the OS's theme (light, dark, day/night). this - in addition to changes I made to accompany this template change - caused certain text inputs and other ui elements (e.g. picker and alert dialogs) to have completely inconsistent and unusable colors set for them when set to the dark theme. This was most apparent in the manual coordinates input screen. If the OS was in light theme, the colors were usable
- this change reverts the template change so that we tell the app to only support the OS light theme, which is what all of our components and UI are designed for anyways.

---

### Preview (issues were only relevant when in dark theme):

- Text inputs are no longer "invisible". Note that only the inputs in this screen seem to be affected by the issue, not the other ones throughout the app 
    - Before: 
    
      image
    - After:
    
      image

- Alert dialogs use light theme
    - Before:
    
      image

    - After

      image

### DIFF
--- a/android/app/src/main/res/values/styles.xml
+++ b/android/app/src/main/res/values/styles.xml
@@ -1,23 +1,12 @@
 <resources>
 
     <!-- Base application theme. -->
+    <!-- If we want to automatic support for dark theme, change Light to DayNight -->
     <style name="AppTheme" 
-        parent="Theme.AppCompat.DayNight.NoActionBar">
+        parent="Theme.AppCompat.Light.NoActionBar">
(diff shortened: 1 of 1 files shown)


achou11:
update: plan is to coordinate this change with the p2p updater fix and release those as part of a hotfix release

Issue #961, opened 2022-04-27
Experiments Feature Flag


ximenabb:
**Description**
To release this feature and have it testable, we will nest the Map Management feature under Settings/ Experiments

**Screens**

1/ 
Experiments

P2P App Updates
Off

Directional Arrow
Off

Map Management (Copy TBD)
Off

2/
Map Management 
TEXT (Copy TBD)
Use Map Management (Checkbox)

(Button : Manage this is Map Management) (Copy TBD)

**Design**

PR #966, opened 2022-05-04
Map selector

### LINKED ISSUE #939: Map View / Bg Map Manager
**Description**
User clicked on the map button to open available maps, horizontal scroll

**Screens**
Background Maps
Manage Maps
Close

**Copydeck**
https://docs.google.com/spreadsheets/d/1Eco_G5hmAqnWW5NfzpAeeMgn8L5c61VO_PnkU7mSIRI/edit#gid=116684765

**Design**
https://www.figma.com/file/sxz2NvysPVab6ZTqs6f57K/Cobra-Collective?node-id=293%3A3004
Note from Sabella:
Dev might want to adjust the size of the thumbnails so the user sees a preview of the row and knows to scroll horizontally (we've talked about this in the past in other concepts with horizontal scrolling). This shows what I mean, one item is clipped to the right as an indicator that there is more to the right https://miro.medium.com/max/1400/1*2T03m1plHOiftBPo4UT0fg.png

ErikSin:
## Created Map Selector 
Map selector is a bottom sheet that gives the user the ability to change map background directly on the map view

image

image

image

## To Do
Connect to Map Server API

Closes: #939 #961


src/frontend/context/SettingsContext.tsx
```diff
@@ -18,27 +18,27 @@ export type SettingsState = {
   experiments: {
     p2pUpgrade: boolean;
     directionalArrow: boolean;
+    BGMaps: boolean;
   };
 };
 
-type SettingsContextType = readonly [
+export type SettingsContextType = [
```
achou11: nit: export doesn't seem necessary since it's not being used outside of this file?

src/frontend/context/SettingsContext.tsx
```diff
@@ -18,27 +18,27 @@ export type SettingsState = {
...
-  (key: keyof SettingsState["experiments"], value?: boolean) => void
+  (key: keyof SettingsState, value: any) => void
 ];
 
 const DEFAULT_SETTINGS: SettingsState = {
   coordinateFormat: "utm",
   experiments: {
     p2pUpgrade: false,
     directionalArrow: false,
+    BGMaps: false,
```
achou11: would avoid the abbreviation here and also change to something that can be camelcased e.g. `backgroundMaps`

src/frontend/screens/Settings/Experiments/BGMaps.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { defineMessages, FormattedMessage, useIntl } from "react-intl";
+import { StyleSheet, Text, View } from "react-native";
+import { ScrollView, TouchableOpacity } from "react-native-gesture-handler";
+import { NavigationStackScreenComponent } from "react-navigation-stack";
+import MaterialIcon from "react-native-vector-icons/MaterialIcons";
+
+import { LIGHT_BLUE, LIGHT_GREY, MEDIUM_GREY } from "../../../lib/styles";
+import HeaderTitle from "../../../sharedComponents/HeaderTitle";
+import { useExperiments } from "../../../hooks/useExperiments";
+import { color } from "react-native-reanimated";
```
achou11: remove unused export

src/frontend/screens/Settings/Experiments/BGMaps.tsx
```diff
@@ -0,0 +1,137 @@
+import * as React from "react";
+import { defineMessages, FormattedMessage, useIntl } from "react-intl";
+import { StyleSheet, Text, View } from "react-native";
+import { ScrollView, TouchableOpacity } from "react-native-gesture-handler";
+import { NavigationStackScreenComponent } from "react-navigation-stack";
+import MaterialIcon from "react-native-vector-icons/MaterialIcons";
+
+import { LIGHT_BLUE, LIGHT_GREY, MEDIUM_GREY } from "../../../lib/styles";
```
achou11: `MEDIUM_GREY` isn't used in this file

src/frontend/sharedComponents/BGMapSelector.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const { navigate } = useNavigation();
+
+  const [snapPoints, setSnapPoints] = React.useState<(number | string)[]>([
+    0,
+    "40%",
+  ]);
+
+  const { formatMessage: t } = useIntl();
+
+  React.useEffect(() => {
```
achou11: would create two separate effects here to separate functionality:

1. back press listener
2. fetching the styles list

src/frontend/sharedComponents/BGMapSelector.tsx
```diff
@@ -0,0 +1,227 @@
...
+      ];
+    }
+
+    getListStyles().then(stylesList => {
+      setBgMapList(stylesList);
+    });
+
+    return () =>
+      BackHandler.removeEventListener("hardwareBackPress", onBackPress);
+  }, [closeSheet, setBgMapList]);
```
achou11: `setBgMapList` doesn't need to be included here

src/frontend/sharedComponents/BGMapSelector.tsx
```diff
@@ -0,0 +1,227 @@
...
+  },
+});
+
+// To do: We should get this all from one central place
+// Perhaps as a Seperate Module published on npm
+interface mapServerStyle {
+  id: string;
+  styleUrl: string;
+  title?: string;
+}
```
achou11: yeah i the plan is for the map server to provide these types directly i.e. basically explaining the shape of the api

if helpful, the official shape for this (as it currently stands) is: `{ id: string; url: string; name?: string }`

src/frontend/hooks/useExperiments.ts
```diff
@@ -1,7 +1,17 @@
...
 export function useExperiments() {
-  const [{ experiments }, , setExperiments] = React.useContext(SettingsContext);
-  return [experiments, setExperiments] as const;
+  const [settings, setSettings] = React.useContext(SettingsContext);
+
+  const setExperiments = React.useCallback(
+    (key: keyof SettingsState["experiments"], val: boolean) => {
+      setSettings("experiments", { ...settings.experiments, [key]: val });
+    },
+    []
```
achou11: `settings` should be addd as a dep here

src/frontend/screens/MapScreen.tsx
```diff
@@ -0,0 +1,96 @@
+import * as React from "react";
+import { GestureResponderEvent, View } from "react-native";
```
achou11: `GestureResponderEvent` is unused

src/frontend/screens/MapScreen.tsx
```diff
@@ -0,0 +1,96 @@
...
+          />
+        </React.Fragment>
+      )}
+    </View>
+  );
+};
+
+interface BGMapButtonProps {
+  /** `openSheet()` should NOT come from `useBottomSheetModal` */
+  onPress: () => void;
```
achou11: a little confused - is the comment just misaligned with the prop name?

src/frontend/sharedComponents/BGMapSelector.tsx
```diff
@@ -0,0 +1,227 @@
...
+interface MapSelectorProps {
+  /** Should NOT come from `useBottomSheet()` */
+  closeSheet: () => void;
+  onMapSelected: (styleURL: string) => void;
+}
+
+/** `ref` should NOT come from - `useBottomSheet()` */
+export const BGMapSelector = React.forwardRef<
```
achou11: nit: probably wouldn't classify this as a shared component. seems like something that'd be specific to the `MapScreen` component

src/frontend/screens/Settings/Experiments/BGMaps.tsx
```diff
@@ -0,0 +1,137 @@
...
+    borderTopWidth: 2,
+    borderBottomWidth: 2,
+    display: "flex",
+    flexDirection: "row",
+    justifyContent: "space-between",
+  },
+  text: {
+    fontSize: 16,
+  },
+  linkContatiner: {
```
achou11:
```suggestion
  linkContainer: {
```

achou11 (changes requested):
First review but looks like a good start! Couple of notes:

- I'd avoid the "BG" abbreviation and just use "Background" when naming things, to be more explicit
- i didn't get to try this out on a device because jcenter is down and i can't build the project at the moment. there may be additional feedback that i have based on testing once i get a chance to do it

src/frontend/screens/MapScreen.tsx
ErikSin: Right, no, this was on old comment. I changed it to be a general press event, but its on only for opening sheet, so I will change it back to OpenSheet()

src/frontend/context/SettingsContext.tsx
ErikSin: addressed: [eae15c2](https://github.com/digidem/mapeo-mobile/pull/966/commits/eae15c281fa5579e7ffa19c82eafe5c36925c760)

src/frontend/context/SettingsContext.tsx
ErikSin: addressed: [eae15c2](https://github.com/digidem/mapeo-mobile/pull/966/commits/eae15c281fa5579e7ffa19c82eafe5c36925c760)

src/frontend/screens/Settings/Experiments/BGMaps.tsx
ErikSin: addressed: [eae15c2](https://github.com/digidem/mapeo-mobile/pull/966/commits/eae15c281fa5579e7ffa19c82eafe5c36925c760)

src/frontend/sharedComponents/BGMapSelector.tsx
ErikSin: addressed: [eae15c2](https://github.com/digidem/mapeo-mobile/pull/966/commits/eae15c281fa5579e7ffa19c82eafe5c36925c760)

src/frontend/screens/MapScreen.tsx
ErikSin: addressed: [eae15c2](https://github.com/digidem/mapeo-mobile/pull/966/commits/eae15c281fa5579e7ffa19c82eafe5c36925c760)

src/frontend/sharedComponents/BGMapSelector.tsx
ErikSin: addressed: [eae15c2](https://github.com/digidem/mapeo-mobile/pull/966/commits/eae15c281fa5579e7ffa19c82eafe5c36925c760)

src/frontend/screens/Settings/Experiments/BGMaps.tsx
ErikSin: addressed: [eae15c2](https://github.com/digidem/mapeo-mobile/pull/966/commits/eae15c281fa5579e7ffa19c82eafe5c36925c760)

src/frontend/screens/MapScreen.tsx
ErikSin: addressed: [eae15c2](https://github.com/digidem/mapeo-mobile/pull/966/commits/eae15c281fa5579e7ffa19c82eafe5c36925c760)

src/frontend/hooks/useExperiments.ts
ErikSin: Adressed: [4e40b49](https://github.com/digidem/mapeo-mobile/pull/966/commits/4e40b4921721c559731a257f50184fe608a426a9)

src/frontend/screens/Settings/Experiments/BGMaps.tsx
ErikSin: addressed: [eae15c2](https://github.com/digidem/mapeo-mobile/pull/966/commits/eae15c281fa5579e7ffa19c82eafe5c36925c760)

src/frontend/screens/Settings/Experiments/BGMaps.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+      <View style={[styles.switchContainer]}>
+        <Text style={[styles.text]}>
+          <FormattedMessage {...m.useBGMap} />
+        </Text>
+
+        <View
+          onTouchStart={() => {
+            setExperiments("backgroundMaps", !experiments.backgroundMaps);
+          }}
```
achou11, line 63: I get an error that freezes the app when pressing this (infinite callstack when setting state in useEffect)

src/frontend/screens/Settings/MapSettings/index.tsx
```diff
@@ -42,10 +42,10 @@ export const MapSettings: NavigationStackScreenComponent = () => {
   );
 };
 
-MapSettings.navigationOptions = {
+MapSettings.navigationOptions = ({ navigation }) => ({
   headerTitle: () => (
     <HeaderTitle>
-      <FormattedMessage {...m.mapSettings} />
+      <FormattedMessage {...m.backgroundMaps} />
```
achou11: is this supposed to be `m.mapSettings`?

src/frontend/screens/MapScreen/BGMapSelector.tsx
```diff
@@ -0,0 +1,228 @@
...
+          id: "7af323j1",
+          styleUrl: MapboxGL.StyleURL.Street,
+        },
+      ];
+    }
+
+    getListStyles().then(stylesList => {
+      setBgMapList(stylesList);
+    });
+  }, [setBgMapList]);
```
achou11: `setBgMapList` doesn't need to be listed here

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,151 @@
...
+        mapStyle: "loading",
+        mapStyleURL: null,
+      };
+  }
+}
+
+/**
+ * @param {string} styleId
+ * @returns `[mapStyle, setMapStyle(styleId:string)]` setMapStyle is a function that can be used to set the map style when map server is on
+ */
```
achou11: nit: probably not necessary to do a full jsdoc annotation for ts files. if you just want a description, i think a multiline comment with the text will work i.e.

```
/**
 * Description of function
 */
```

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,151 @@
...
+  const [{ backgroundMaps }] = useExperiments();
+
+  React.useEffect(() => {
+    if (state.mapStyle === "mapServer" || state.mapStyle === "customLegacy")
+      return;
+
+    if (ssid === null && state.mapStyle === "online") {
+      dispatch({ type: "setFallback" });
+    }
+  }, [ssid]);
```
achou11: probably need to add `state` as a dep here. better yet, to avoid adding the dep, you can do the logic above in the reducer instead

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  React.useEffect(() => {
+    if (!backgroundMaps && state.mapStyle === "mapServer") {
+      turnOffMapServer(ssid !== null);
+    }
+
+    if (backgroundMaps && state.mapStyle !== "mapServer") {
+      turnOnMapServer();
+    }
+  }, [backgroundMaps, ssid, turnOffMapServer, turnOnMapServer]);
```
achou11: not sure if this is needed. the map server should already be available (or starting up) when the app starts

achou11: looking through the implementation, i think the function names are misleading. this is more about fetching the style to use, not the on/off status of any map server.

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,151 @@
...
+        mapStyleURL: null,
+      };
+  }
+}
+
+/**
+ * @param {string} styleId
+ * @returns `[mapStyle, setMapStyle(styleId:string)]` setMapStyle is a function that can be used to set the map style when map server is on
+ */
+export function useMapStyle(styleId: string = "default") {
```
achou11: not exactly sure if the issue lies here, but i'm not getting any map when initially opening the app (experiment not enabled, just stuck on the loading map state). my guess is that this has something to do with it but haven't dug into it

achou11 (changes requested):
Haven't been able to fully test the changes on my device (see  https://github.com/digidem/mapeo-mobile/pull/966#discussion_r866356725 and review comments). Think the updated `useMapStyle` implementation needs some more work.

Seems like CI is failing (probably related to some UI change)

Also, thoughts on renaming the components/functions/variables that use `BG` to use `Background`?

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,151 @@
...
+  const [state, dispatch] = React.useReducer(mapStyleReducer, {
+    mapStyle: "loading",
+    mapStyleURL: null,
+  });
+
+  const { ssid } = useWifiStatus();
```
gmaclennan: This is not an indicator of whether the user is online or not. A user can be on wifi but not online, and they can be without wifi and online.

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,151 @@
...
+  }, [ssid]);
+
+  React.useEffect(() => {
+    if (!backgroundMaps && state.mapStyle === "mapServer") {
+      turnOffMapServer(ssid !== null);
+    }
+
+    if (backgroundMaps && state.mapStyle !== "mapServer") {
+      turnOnMapServer();
+    }
```
gmaclennan: It's a red flag to have useEffect without an async action. This should be in the main body of the function, and whether it is turned on should be derived from the state, not part of the state.

gmaclennan (changes requested):
Sorry I had very limited time to look at this so comments aren't very comprehensive. `useWifiState()` does not tell us whether the device is online or not - it can mean the opposite (4G, or offline wifi network for syncing).

The useMapState() hook needs re-working. All the useEffect() functions risk causing lots of loops and re-renders, and with all the dependencies the logic is really hard to follow. As much as possible this should be "given this state, this is what should render" (e.g. a reactive function), but right now it is full of imperative actions.

src/frontend/hooks/useMapStyle.ts
ErikSin: No, I am thinking about this in terms of the experimental feature flag. Turning on map server, allows the user to use a map style in map server. Turning off map server, defaults to any styles not available through map server (aka, `onlineDefaultMap` or `offlineMap` or `customMaps`

src/frontend/hooks/useMapStyle.ts
ErikSin: Ok, this is just an indicator of being connected to wifi, good to know!

ErikSin:
> Sorry I had very limited time to look at this so comments aren't very comprehensive. `useWifiState()` does not tell us whether the device is online or not - it can mean the opposite (4G, or offline wifi network for syncing).
> 
> The useMapState() hook needs re-working. All the useEffect() functions risk causing lots of loops and re-renders, and with all the dependencies the logic is really hard to follow. As much as possible this should be "given this state, this is what should render" (e.g. a reactive function), but right now it is full of imperative actions.

@gmaclennan I agree that declarative is the preferred style in react. I think the trouble is that everytime the user turns off the map server, we need to check:
1. If custom maps are available (this is done async, so not good practice to put in a reducer function).
2. If custom maps are NOT available then check if online maps are available.
3. If online maps are NOT available then set default offline map.

Given that these three checks have to happen every time, imperative programming seems much more streamline. That being said, ill try it with a declarative style and see if that clears up this hook.

ErikSin:
@gmaclennan I reworked things a bit, let me know what you think

src/frontend/screens/MapScreen/BGMapSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+        style={{ padding: 20 }}
+      >
+        {bgMapsList === null ? (
+          <Loading />
+        ) : (
+          <View style={{ backgroundColor: WHITE }}>
+            <Text style={styles.title}> {t(m.title)}</Text>
+            <TouchableOpacity
+              activeOpacity={0.99}
+              onPress={() => {
```
ErikSin: There is a slight opacity issue here where you can slightly see background map. Open to suggestions on how to solve this, but not a big deal. You can only see it if you are really looking!

src/frontend/AppContainerWrapper.tsx
```diff
@@ -145,13 +144,12 @@ const AppContainerWrapper = () => {
...
+    [updateRouteBasedAppState]
   );
 
-  const AppContainer = React.useMemo(
-    () => (onboarding ? OnboardingContainer : DefaultContainer),
-    [onboarding]
-  );
+  const AppContainer = devExperiments.onboarding
+    ? OnboardingContainer
+    : DefaultContainer;
```
achou11: since `devExperiments` is determined at build time now and never changes at runtime, i would move this outside of the component and also remove line 97

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,150 @@
...
+    return () => {
+      didCancel = true;
+    };
+  }, []);
+
+  React.useEffect(() => {
+    if (isOnline && state.type === "fallback") {
+      dispatch({ action: "setOnline" });
+    }
+  }, [isOnline]);
```
achou11: probably need `state.type` in the deps array here

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (id) AsyncStorage.setItem(MAP_STYLE_KEY, id);
+  }, [id]);
+
+  return React.useMemo(
+    () => ({
+      styleUrl: id ? api.getMapStyleUrl(id) : null,
+      styleType: "mapServer",
+      setStyleId: setId,
+    }),
+    [id, setId]
```
achou11, line 135: `setId` doesn't need to be in the deps array here

src/frontend/screens/MapScreen/BGMapSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+  },
+  manageMaps: {
+    id: "sharedComponents.BGMapSelector.manageMaps",
+    defaultMessage: "Manage Maps",
+  },
+});
+
+// To do: We should get this all from one central place
+// Perhaps as a Seperate Module published on npm
+interface mapServerStyle {
```
achou11:
```suggestion
interface MapServerStyle {
```

src/frontend/screens/Settings/ProjectConfig/index.tsx
```diff
@@ -105,7 +104,7 @@ export const ProjectConfig = () => {
 
   const loading = status === "loading" || config.status === "loading";
 
-  const isPracticeMode = onboarding && isInPracticeMode(config);
+  const isPracticeMode = devExperiments.onboarding && isInPracticeMode(config);
```
achou11, line 107: can't point directly to it in this PR but can remove line 66 which looks like

```js
const { onboarding } = devExperiments;
```

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -113,28 +118,51 @@ export const BackgroundMaps: NavigationStackScreenComponent = ({
...
+              <View style={styles.importTextAndIcon}>
+                <MaterialIcon
+                  name="file-upload"
+                  size={30}
+                  color={MEDIUM_GREY}
+                />
+                <Text style={styles.text}> {t(m.importFromFile)}</Text>
+              </View>
+              <Text style={[styles.text, { textAlign: "center" }]}>
+                {"( .mbtile )"}
```
achou11:
```suggestion
                {"(.mbtiles)"}
```

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -113,28 +118,51 @@ export const BackgroundMaps: NavigationStackScreenComponent = ({
...
+          </HeaderTitle>
+
+          <TouchableOpacity
+            onPress={handleImportPress}
+            style={styles.importButton}
+          >
+            <React.Fragment>
+              <View style={styles.importTextAndIcon}>
+                <MaterialIcon
+                  name="file-upload"
```
achou11, line 145: just noting that this is probably the correct icon to use, but it's different than the design (which uses the download icon)

src/frontend/screens/MapScreen/BGMapSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+  title?: string;
+}
+interface MapSelectorProps {
+  /** Should NOT come from `useBottomSheet()` */
+  closeSheet: () => void;
+  onMapSelected: (styleURL: string) => void;
+}
+
+/** `ref` should NOT come from - `useBottomSheet()` */
+export const BGMapSelector = React.forwardRef<
```
achou11, line 49: couple of small things here based on the designs. okay to do as follow ups:

- missing: name of map under each thumbnail
- missing: border to highlight currently selected map
- missing: description of currently selected map with info about offline areas. fine to skip this one since we're not dealing with offline areas for first iteration
- too much space between close button and bottom of screen

src/frontend/screens/MapScreen/BGMapSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+  map,
+  onMapSelected,
+}: {
+  map: mapServerStyle;
+  onMapSelected: (styleURL: string) => void;
+}) => {
+  const { position } = React.useContext(LocationContext);
+
+  return (
+    <TouchableHighlight
```
achou11, line 184: noting that the opacity comment [here](https://github.com/digidem/mapeo-mobile/pull/966/files#r879838411) should be pointing to here instead

src/frontend/screens/MapScreen/BGMapSelector.tsx
achou11: this comment should be pointing to the MapThumbnail component instead

src/frontend/screens/MapScreen/BGMapSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+          setSnapPoints([0, height]);
+        }}
+        style={{ padding: 20 }}
+      >
+        {bgMapsList === null ? (
+          <Loading />
+        ) : (
+          <View style={{ backgroundColor: WHITE }}>
+            <Text style={styles.title}> {t(m.title)}</Text>
+            <TouchableOpacity
```
achou11, line 124: nit: this container spans the entire width of the sheet, causing presses outside of what may be a reasonable area from the intended target text to be registered. adjusting the width and/or hitslop of this touchable would help

src/frontend/screens/MapScreen/BGMapSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+              activeOpacity={0.99}
+              onPress={() => {
+                navigate("MapSettings");
+              }}
+            >
+              <Text
+                style={{
```
achou11, line 130: nit: based on designs, should increase font weight

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  React.useEffect(() => {
+    AsyncStorage.getItem(MAP_STYLE_KEY).then(id => {
+      if (id) setId(id);
+    });
+  }, []);
+
+  React.useEffect(() => {
+    if (id) AsyncStorage.setItem(MAP_STYLE_KEY, id);
+  }, [id]);
```
achou11: maybe instead of relying on an effect here, create a callback so that this is handled at the same time as setting state? (the `useCallback` here is probably unnecessary)

```js
const setStyleId = React.useCallback((id?: string) => {
  // Do whatever combination of operations works here
  setId(id)
  if (id) AsyncStorage.setItem(MAP_STYLE_KEY, id)
}, [])
```

this way you avoid any chance of runaway effects and have the async storage operation more closely tied to the event that triggers it

src/frontend/screens/MapScreen/BGMapSelector.tsx
```diff
@@ -0,0 +1,231 @@
...
+    height: 80,
+  },
+  title: {
+    fontSize: 32,
+    textAlign: "center",
+    paddingBottom: 20,
+  },
+  divider: {
+    borderBottomColor: LIGHT_GREY,
+  },
```
achou11: no longer being used?

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -1,18 +1,17 @@
...
-  BottomSheetModal,
-  useBottomSheetModal,
-} from "../../../sharedComponents/BottomSheetModal";
+import BottomSheet, { BottomSheetBackdrop } from "@gorhom/bottom-sheet";
 import Button from "../../../sharedComponents/Button";
 import HeaderTitle from "../../../sharedComponents/HeaderTitle";
 import Loading from "../../../sharedComponents/Loading";
+import { BottomSheetMethods } from "@gorhom/bottom-sheet/lib/typescript/types";
+import MaterialIcon from "react-native-vector-icons/MaterialIcons";
+import { TouchableOpacity } from "react-native-gesture-handler";
```
achou11: nit: import from `sharedComponents/Touchables`

achou11 (commented):
Pointed out a few things that probably should be fixed in this PR, but a good amount of it can also be follow-up.

Slightly worried about the new implementation of the legacy useMapStyle hook, but from basic testing, didn't find anything that was noticeably broken. Think an edge case was fixed but there may be edge cases that weren't addressed in exchange, which is why I was leaning towards just keeping that part of the implementation the same as before.

Not sure how it's related to this PR, but noticing that on app startup, the map doesn't automatically zoom into my location. feel like i've seen this before (maybe from a local branch i was working on and updating the map screen in), but thought I should point that out in case it turns out to be a problematic regression

src/frontend/hooks/useMapStyle.ts
ErikSin: Why not? `setid` comes from the useState hook, so it is automatically memoized if thats what you are worried about

achou11: that's my point. it's a stable reference so there's no need to specify it in the deps array

ErikSin: For the sake of pure functional programming (and consistency), I think it is best practice to have all outside dependencies listed in the dependency array when possible.

achou11: it doesn't have anything to do with functional programming. state setters from the built-in useState and useReducer shouldn't be specified in a deps array. see https://reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often as an indirect example. if anything, it's more consistent to omit from the deps array

> (The identity of the setCount function is guaranteed to be stable so it’s safe to omit.)

src/frontend/screens/MapScreen/BGMapSelector.tsx
ErikSin: Whoops! good catch, that explains alot....

src/frontend/hooks/useMapStyle.ts
ErikSin: It is related to functional programming in terms of how we deal with side effects: https://dennyscott.io/use-effect-dependency-array/.

Regardless it does seem like a preference thing at this point

achou11: > Regardless it does seem like a preference thing at this point

yeah it is, feel free to keep as is 👍

src/frontend/screens/MapScreen/BGMapSelector.tsx
ErikSin: Im resolving this for now, but adding it to the ticket for future use

src/frontend/screens/MapScreen/BGMapSelector.tsx
ErikSin: Added to ticket for follow

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,154 @@
...
+} {
+  const [{ backgroundMaps }] = useExperiments();
+
+  const { isInternetReachable } = useNetInfo();
+
+  const legacyStyleInfo = useLegacyStyle(!!isInternetReachable, styleId);
+  const mapServerInfo = useMapServerStyle();
+
+  return backgroundMaps ? mapServerInfo : legacyStyleInfo;
+}
```
gmaclennan, line 154: Good solution to this. Conditions in hooks are really tricky (i.e. they don't generally work!) and this solves it well. It means redundant requests, but I think that's fine in this case and hopefully we don't be in this state for long (we can move it out of the feature flag soon hopefully). Another solution is to conditionally render a component that contains the hook, that accepts a function as children, if this kind of thing crops up in the future, e.g.

```js
<MapStyle>{(mapStyle) => <Map style={mapStyle} />}</MapStyle>
```

However I think your solution Erik is the right one here.

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,154 @@
...
+    },
+  };
+}
+
+function useMapServerStyle(): {
+  styleUrl: string | null;
+  styleType: MapTypes;
+  setStyleId: (id: string) => void;
+} {
+  const [id, setId] = React.useState<string | null>(null);
```
gmaclennan, line 113: This seems to make sense to be in context rather than a hook, because each instance of this hook (in the Map screen component and in the Map manager) would have a different id, setId which would be initialized from local storage. There is a (small) chance of a race condition here if different screens are calling a different setId() and each is trying to save to local storage.

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,154 @@
...
+    });
+  }, []);
+
+  const setIdWithAsync = React.useCallback(
+    (id: string) => {
+      setId(id);
+      AsyncStorage.setItem(MAP_STYLE_KEY, id);
+    },
+    [setId]
+  );
```
gmaclennan, line 127: We already have a `usePersistedStorage` hook for this, which catches some of the edge cases with this. Hopefully soon we can switch to something like https://github.com/mrousavy/react-native-mmkv and make a lot of the persisted state a lot simpler, since it's all synchronous.

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,154 @@
...
+export function useMapStyle(
+  styleId: string = "default"
+): {
+  styleUrl: string | null;
+  styleType: MapTypes;
+  setStyleId: (id: string) => void;
+} {
+  const [{ backgroundMaps }] = useExperiments();
+
+  const { isInternetReachable } = useNetInfo();
```
gmaclennan, line 148: Style nit: this is only used by the `useLegacyStyle` hook, so maybe include it in that hook.

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  const legacyStyleInfo = useLegacyStyle(!!isInternetReachable, styleId);
```
gmaclennan, line 150: We are loosing a state here: `onlineState = 'unknown' | 'online' | 'offline'`
In the previous hook this was reflected in a `loading` state.

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,154 @@
...
+    return () => {
+      didCancel = true;
+    };
+  }, []);
+
+  React.useEffect(() => {
+    if (isOnline && state.type === "fallback") {
+      dispatch({ action: "setOnline" });
+    }
+  }, [isOnline, state.type]);
```
gmaclennan, line 95: Using `React.useEffect()` like this is going to bite us because it is so easy to accidentally create an infinite render loop that freezes the app, and in addition is causes an unnecessary double render. I think it's safer and easier to read if this state is derived rather than set in a use effect, that we avoid side effects and it's easier to understand the state given the inputs e.g.

user is online & custom maps not available -> use online style
user is offline & custom maps not available -> use fallback style

In this implementation the logic is unclear: if the user is online and using the online style, but then goes offline, then this effect will fire, but it will not change anything.

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  React.useEffect(() => {
+    let didCancel = false;
+
+    // checks if custom maps are available
+    // If not, checks if online maps are available
+    // If not, sets to fallback maps
+    api
+      .getMapStyle(styleId)
```
gmaclennan, line 74: This api method returns a map style, not a style URL. The `api.getMapStyleUrl()` method is synchronous.

src/frontend/hooks/useMapStyle.ts
```diff
@@ -0,0 +1,154 @@
...
+    case "setCustom":
+      return { type: "custom", styleUrl: action.styleUrl };
+    case "setOnline":
+      return { type: "online", styleUrl: onlineStyleURL };
+    case "setFallback":
+      return { type: "fallback", styleUrl: fallbackStyleURL };
+    default:
+      return { type: "loading", styleUrl: null };
+  }
+}
```
gmaclennan, line 53: I don't think a reducer makes sense here, because we are mixing different things:

1. Is the online state online, offline, or unknown?
2. Are custom maps available

The styleUrl used should be based on the combination of those two states, but derived in a way you can write as a function, as opposed to a bunch of side effects, e.g.

```
const styleUrl = getStyleUrl(
  onlineState: 'unknown' | 'online' | 'offline',
  customMapState: 'unknown' | 'available' | 'unavailable'
)
```

src/frontend/screens/MapScreen/MapScreen.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  const handleObservationPress = React.useCallback(
+    (observationId: string) =>
+      navigation.navigate("Observation", { observationId }),
+    [navigation]
+  );
+
+  const handleAddPress = React.useCallback(() => {
+    log("pressed add button");
+    newDraft(undefined, undefined);
```
gmaclennan, line 45: Why are you changing this from `newDraft(undefined, { tags: {} })`. I can't remember why this was like this, but I want to be sure this change does not break things.

gmaclennan (changes requested):
Good work on this Erik, anything that touches some of my old map code is going to be a headache. I mainly focussed my review on the logic in the hooks, because that seems the most likely place for bugs to appear. I think the legacy style hook needs fixed before we merge (because it risks breaking existing behaviour). The new style hook we could merge as-is but should fix soon after.

For the legacy map style hook, I'm concerned that the re-write introduces new bugs. As I understand it one of the goals is to include the `isInternetReachable` value so that the map style can be responsive to online state (e.g. change if the user goes online/offlien) because this was a todo in the original hook. I don't think this re-write achieves that, and I think it potentially introduces new bugs. I think maybe stick with the original implementation (since we know it works) and leave the responsive `isInternetReachable` stuff to a second PR.

For the new style hook, I think the main issue is that I think this belongs in context, not in a hook, and there is a chance that we can get bitten by some race conditions (especially as a bit down the line we might forget that the `[id, setId]` returned by the hook are different in each component, and they each take their time to initialize. Also I think we should re-use the `usePersistedState()` hook, which has a few more checks for edge-cases.

I just did a quick lookover of all the components, and at a glance they all look good to me. I have not tested them visually though, I've just read through the code.

On a general note, I think it's important practice in React code to avoid `useEffect` when there is no async action (or side-effect), because it is easy to introduce bugs and render loops. Normally using it like this is a sign that you want a derived state, e.g. instead of something like

```js
const isOnline = useOnlineState()
const [icon, setIcon] = useState('offlineIcon')

useEffect(() => {
  if (isOnline) setIcon('onlineIcon')
}, [isOnline])

return icon
```

we should do something like:

```js
const isOnline = useOnlineState()
return isOnline ? 'onlineIcon' : 'offlineIcon'
```

(this is a simplified example of the code in the map style hook)


Also side-note, I saw the conversation about including `setId` in the dependency array of useEffect/useMemo/useCallback. Agreed that it's a stylistic preference. It would be nice to aim for consistency here though: there are a few useEffect/useMemo/useCallback in this PR that have a `setState` or `dispatch` in them and they do not include it in the dependency array. I don't think it's an issue for this PR, but we should just decide on one or the other as a team and stick to that (although noting that the default lint rules will often show a warning for including an unnecessary dependency)

src/frontend/screens/MapScreen/MapScreen.tsx
ErikSin: I changed it to TS, and TS was complaining. I originally changed it to null, which broke the e2e. Changing it to undefined seem to fix it, BUT i can change the `useDraftObservation()` hook to accomodate for that if we are worried

src/frontend/hooks/useMapStyle.ts
ErikSin: What i am trying to capture here is if the user is offline, and then goes online, we want the online maps to load. Right now mapeo has to restart completely in order for the map to go back online

gmaclennan: Yeah understood, it's best to do that with derived state rather than a useEffect, since bugs are easier to spot (e.g. the current bug if the user is online and goes offlien).

ErikSin: But I do see your concern about the infinite loop, but since it is attached to the reducer, we know that when it is `setOnline` the `state.type` is set to `online`

src/frontend/hooks/useMapStyle.ts
ErikSin: Right, good catch, that makes sense

src/frontend/screens/MapScreen/MapScreen.tsx
gmaclennan: Yeah just had a look at the code and this implementation is ok. May as well just write `newDraft()` rather than pass arguments. The Typescript in the hook is a bit weird and could do with a clean up (the reason this broke TS was because while the implementation here https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/hooks/useDraftObservation.ts#L170 correctly has the ClientGeneratedObservation type (which would allow `{ tags: {} }`), the `newDraft` definition here https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/hooks/useDraftObservation.ts#L64 does not. Also I don't think the type should include `null`, which means that the default value of `{ tags: {} }` is not set, which is why the e2e broke. Anyway, we can fix up the types in `useDraftObservation` later.

src/frontend/hooks/useMapStyle.ts
ErikSin: If custom maps are available, we should serve that, and the online state doesnt matter, correct? That is why im using a reducer here, because knowing the online state is not always necessary

ErikSin: ^or is that assumption wrong?

gmaclennan: I think that assumption is correct. But it means when maintaining this code you need to know that you should only call "setOnline()" when mapType !== 'custom', which can lead to bugs. I feel it kind of obfuscates what is happening, I think what you actual have here is a function like the one in my comment above. To expand that code above:

```ts
const {type, styleUrl} = getStyleUrl(onlineState, customMapState)

function getStyleUrl(
  onlineState: 'unknown' | 'online' | 'offline',
  customMapState: 'unknown' | 'available' | 'unavailable'
) {
  if (onlineState === 'unknown' || customMapState === 'unknown')  {
    return { type: 'loading', styleUrl: null }
  } else if (customMapState === 'available') {
    return { type: 'custom', styleUrl: customMapStyleUrl }
  } else if (onlineState === 'online') {
    return { type: "online", styleUrl: onlineStyleURL };
  } else {
      return { type: "fallback", styleUrl: fallbackStyleURL };
  }
}
```

I think writing it that way you avoid double-renders (render, then call useEffect, then render), you avoid the risk of render loops, you avoid needing to know which actions you can dispatch based on the current state, and in my opinion it is clearer for the reader what the result is given each state (e.g. I can easily see from the above that if the custom map is available then that will be what is used).

gmaclennan: (this is similar to what is in the original hook, just cleaned up a bit)

Issue #962, opened 2022-04-27
Add Map Server to Mobile


ErikSin:
# Integrate Map Server to Mobile

## Architecture
Background map management backend: [Map Server](https://github.com/digidem/mapeo-mapserver)

Front End Screens: [Pull Request](https://github.com/digidem/mapeo-mobile/pull/946)

### Terminology
- Background Map: Refers to the `styles` of maps

### To Connect
- [x] GET List of Background Maps
- [ ] ~~GET List of Offline Areas for each map~~
- [ ] DELETE Background Map
- [x] POST `mbtiles` files
- [x] GET Styles (to render new tiles for mapview)

### To do:
- [ ] Discuss how we manage error states

# Known issues:
- [x] [Map not be presented correctly](https://github.com/digidem/mapeo-mobile/issues/962#issuecomment-1139195924)
- [x] [Delete Style Returning 500 error](https://github.com/digidem/mapeo-mobile/issues/962#issuecomment-1139195924)
- [ ] Uploading 2 of the same map causes the map to become inaccessible to user and map cannot be uploaded again
- [ ] Mbtiles files that are not raster are not throwing errors, and causing a blank black map to be shown
- [ ] "Deleted" maps cannot be uploaded again
- [ ] Frontend cannot catch fastify errors

ErikSin:
## Update:
Two obvious errors:
- Map not being presented properly:
<img width="343" alt="image" src="https://user-images.githubusercontent.com/67773827/170610089-56aa22d9-7abf-4b98-bebe-33282e1a30ef.png">

- Delete Background Map `api.maps.deleteStyle(id)` returning httpError:500:
<img width="243" alt="image" src="https://user-images.githubusercontent.com/67773827/170610289-fd84ff7d-3220-41f6-bd3a-dbca51727758.png">

ErikSin:
@gmaclennan

gmaclennan:
The incorrect map view was caused by two errors in the mbtiles import code:

1. `tile_row` in the mbtiles sqlite database is the y (not the x) and `tile_column` is x
2. mbtiles are [stored in TMS scheme](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#content-1) not xyz.

Fixed in https://github.com/digidem/mapeo-map-server/commit/f8a5dadb025a953cb7970563cc8102c138865df0 and published as `@mapeo/map-server@1.0.0-alpha.5`

gmaclennan:
[...]

PR #969, opened 2022-05-17
chore: set up initial map server integration

achou11:
Towards #962

Notes:
- Introduces `@mapeo/map-server` into the backend which requires a few workarounds in the backend build process to get it working
- The frontend doesn't use any of the api methods implemented yet. Plan is to implement in a follow-up PR
- The api is available via the `maps` namespace, which is only available if the feature flag is enabled. example usage is `api.maps.getStyle(id)`
- Things to address:
  - tests. ideally would at least add some tests for the api layer (see `api.spec.js`)
  - error handling
    - server-side: mostly just mirrored a previous implementation for handling any errors that occur, but not entirely sure if that's the best way to approach it
    - client-side: the api client methods leaves error handling mostly to the user, similar to the existing api for the app. also not entirely sure if error status codes/messages are coming through properly; for some reason the messages we defined in the map server lib seem inaccessible client-side. also, there's nothing set up to listen for server states for the map server. not sure if this is necessary but seems like something we should have implemented


src/backend/mapServer.js
```diff
@@ -0,0 +1,69 @@
+const { promisify } = require("util");
+const getPort = require("get-port");
+const createServer = require("@mapeo/map-server").default;
+const log = require("debug")("map-server");
+
+const AsyncService = require("./upgrade-manager/async-service");
+
+// TODO: Does it make sense for this to extend AsyncService?
+class MapServer extends AsyncService {
```
achou11: using the async-service seems to make sense here, but open to other thoughts. didn't want to bother with moving files around so it's currently still under the upgrade manager directory, even though it's more general

src/backend/mapServer.js
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Start the server on the specified port. Listen on all interfaces.
+   *
+   * @returns {Promise<void>} Resolves with the port number when server is started
+   */
+  async _start() {
+    if (!this.#port) {
+      this.#port = await getPort();
+    }
```
achou11, line 35: only get an available port on initial server startup. subsequent calls will reuse the port, which comes with a small risk

src/frontend/api.ts
```diff
@@ -143,9 +152,113 @@ export const SERVER_START_TIMEOUT = 30000;
...
+  baseUrl?: string;
+  onReady?: () => Promise<void>;
+} = {}) {
+  const req = ky.extend({
+    prefixUrl: baseUrl,
+    timeout: false,
+    headers: {
+      "cache-control": "no-cache",
+      pragma: "no-cache",
+    },
```
achou11, line 169: the app server uses these headers so made an assumption that we'd want the same with the map server. not entirely sure if that's a good assumption though

src/frontend/api.ts
```diff
@@ -143,9 +152,113 @@ export const SERVER_START_TIMEOUT = 30000;
...
+function createRequestClient({
+  baseUrl,
+  onReady,
+}: {
+  baseUrl?: string;
+  onReady?: () => Promise<void>;
+} = {}) {
+  const req = ky.extend({
+    prefixUrl: baseUrl,
+    timeout: false,
```
achou11, line 165: there's no manual timeout code for the map server like there is for the app server at the moment, meaning that there's nothing we do to handle the case where the map server hangs. solution is to either:

- specify a timeout (i.e. let that be an option for the function)
- add proper timeout handling code specific to map server

src/backend/mapServer.js
gmaclennan: Yes that makes sense. I'm thinking of changing that API for this slightly, to make it into its own state machine, so that rather than extending the AsyncService class you do something like:

```js
  constructor() {
    #service = new AsyncStateMachine({ start, stop })
  }
```

We can do that in a follow-up PR though, I'll extract the start/stop service management from the project invite PR.

src/backend/mapServer.js
```diff
@@ -0,0 +1,69 @@
...
+
+    log(`${this.#port}: starting`);
+    if (!this.#fastifyStarted) {
+      log("first start, initializing fastify");
+      await this.#fastify.listen(this.#port);
+      this.#fastifyStarted = true;
+    } else {
+      log("second start, listening");
+      const { server } = this.#fastify;
+      await promisify(server.listen.bind(server))(this.#port);
```
gmaclennan:
```suggestion
      await promisify(server.listen).call(server, this.#port);
```
Nit pick, but I wonder if creating a new bound function each time has more of an overhead that a `call()`?

src/frontend/api.ts
gmaclennan: Maybe want to re-visit this in a further iteration (beyond this PR). Caching can help reduce CPU use, but causes issues when users update map styles etc.

src/frontend/api.ts
```diff
@@ -224,29 +326,19 @@ export function Api({ baseUrl, timeout = DEFAULT_TIMEOUT }: ApiParam) {
...
+  const { del, get, post, put } = createRequestClient({ baseUrl, onReady });
+
   // All public methods
   const api = {
+    /**
+     * Map server methods
+     */
+    ...(devExperiments.mapSettings
+      ? { maps: createMapServerApi() }
+      : undefined),
```
gmaclennan, line 436: Will this work? The exported api is a single instance created at app startup. If the user turns on the setting then this will not update. Maybe keep the methods available whether the experiment is turned on or not? Otherwise would require an app restart after turning it on/off.

src/frontend/api.ts
```diff
@@ -143,9 +152,113 @@ export const SERVER_START_TIMEOUT = 30000;
...
+      throw new Error(
+        "Map server client cannot be used because port is unknown"
+      );
+
+    if (!client) {
+      client = createMapServerClient(mapServerPort);
+    }
+
+    return client;
+  }
```
gmaclennan: Haven't completely processed this, but what will happen if the front-end requests something from the map server before the `map-server::start` event has fired? I think it's quite likely the user will start up in the map screen and the front end would immediately make requests (similarly on a `resume` when returning to the app). I think we might need a `await mapServer.ready()` type pattern with a status listener?

gmaclennan (commented):
I had a quick look through this. A couple of small things, but generally looks good. I think the startup code might cause errors when starting/resuming the app though when the front-end makes a request to the map server before it has started.

src/backend/mapServer.js
achou11: honestly i have no idea. i just copied this from the upgrade example and my JS knowledge isn't esoteric enough to know this off the top of my head 😄

src/frontend/api.ts
achou11: `devExperiments` is the build time feature flag that enables whether or not the user even has the option to use the map server or not. the setting you're thinking of belongs to the settings context, where using the map server can be toggled. so in this case, the api becomes available if the feature flag is on, regardless of if the user has actually enabled the map server or not

for the sake of simplicity, we could just ignore this check and just make sure that no components are attempting to call it when the map server option isn't available at all

src/frontend/api.ts
achou11: yeah in theory that's the correct way to go. in practice, it hasn't been an issue but i think that's because of luck and how the frontend waits for the app server before rendering the map

src/backend/mapServer.js
achou11: upon reading mdn seems like this change makes sense and would help a little

src/frontend/api.ts
achou11: might be a candidate to follow-up on after fixing up the server start/stop code, since it'll probably be related to that

src/frontend/api.ts
achou11: > it hasn't been an issue but i think that's because of luck and how the frontend waits for the app server before rendering the map

correction, it wasn't an issue because i was triggering api calls manually as opposed to some effect calling them on startup. so yeah, you're right that there's an issue with that in its current state

achou11: I added what I consider a temporary and naive solution for the issue here: https://github.com/digidem/mapeo-mobile/pull/969/commits/d0c2bcd1bda396611d071187a8cd8cbfabc2808b

There's no server status implementation for the map server specifically, which makes this feel incomplete. I figured that the cleanup you're doing (and potentially extracting) in https://github.com/digidem/mapeo-mobile/pull/748 will address a good portion of that so figured it could wait for that for the time being

src/frontend/api.ts
```diff
@@ -143,9 +153,207 @@ export const SERVER_START_TIMEOUT = 30000;
...
+      );
+
+      mapServerReadyPromise.catch(err => {
+        // @ts-expect-error
+        appServerStatusSubscription.remove();
+        bugsnag.notify(err);
+      });
+
+      return mapServerReadyPromise as Promise<void>;
+    },
```
gmaclennan, line 273: Messy, but looks like it will work! I accept full responsibility here for settings this up with this spaghetti tangle between backend and frontend! Will try to get this fixed up soon!

gmaclennan (approved):
Let's get this merged then start testing and iterating to improve it.

PR #973, opened 2022-05-25
Map selector

### LINKED ISSUE #939: Map View / Bg Map Manager

ErikSin:
## Created Map Selector 
Map selector is a bottom sheet that gives the user the ability to change map background directly on the map view

image

image

image

## To Do
Connect to Map Server API

Closes: #939 #961


gmaclennan:
Great Erik, that should fix things. Just did some cleanup of the hook, making the types stricter (which highlighted the need to handle the case of styleId = null when using the new map server - we can do that in a later PR), and I ended up just implementing the isInternetReachable check. For the legacy maps, we don't support any other styleId other than "default" so I just removed that to avoid that confusing anyone the next time we look at this code.

PR #974, opened 2022-05-26
chore: simplify nodejs startup & fix bugsnag

gmaclennan:
The backend requires values that are only available in the React Native process (storage folders, app version etc.)

Currently the nodejs process waits on an IPC message from RN containing the config before starting. This makes the server startup more complicated and it introduces edge cases for when the app might pause and resume before the server actually starts.

This PR fixes this with the new `nodejs.startWithArgs()` method in nodejs-mobile-react-native@0.8.1. This allows us to pass all the config arguments at the node process startup.

This PR also fixes the bugsnag config for the backend, which previously incorrectly reported the app version (it used the version in package.json, which is only correct for production releases).

### DIFF
--- a/patches/nodejs-mobile-react-native+0.8.1.patch
+++ b/patches/nodejs-mobile-react-native+0.8.1.patch
@@ -19,7 +19,7 @@ index 337de9b..636ef2b 100644
    private final ReactApplicationContext reactContext;
    private static final String TAG = "NODEJS-RN";
    private static final String NODEJS_PROJECT_DIR = "nodejs-project";
-+  private static final String NODEJS_ASSETS_DIR = "nodejs-assets"
++  private static final String NODEJS_ASSETS_DIR = "nodejs-assets";
    private static final String NODEJS_BUILTIN_MODULES = "nodejs-builtin_modules";
    private static final String TRASH_DIR = "nodejs-project-trash";
    private static final String SHARED_PREFS = "NODEJS_MOBILE_PREFS";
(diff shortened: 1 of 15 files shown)


patches/nodejs-mobile-react-native+0.8.1.patch
```diff
@@ -1,48 +1,25 @@
...
-index e882a0c..d5ed11a 100644
+index 337de9b..636ef2b 100644
 --- a/node_modules/nodejs-mobile-react-native/android/src/main/java/com/janeasystems/rn_nodejs_mobile/RNNodeJsMobileModule.java
 +++ b/node_modules/nodejs-mobile-react-native/android/src/main/java/com/janeasystems/rn_nodejs_mobile/RNNodeJsMobileModule.java
 @@ -32,6 +32,7 @@ public class RNNodeJsMobileModule extends ReactContextBaseJavaModule implements
    private final ReactApplicationContext reactContext;
    private static final String TAG = "NODEJS-RN";
    private static final String NODEJS_PROJECT_DIR = "nodejs-project";
-+  private static final String NODEJS_ASSETS_DIR = "nodejs-assets";
++  private static final String NODEJS_ASSETS_DIR = "nodejs-assets"
```
achou11: oyyy
```suggestion
+  private static final String NODEJS_ASSETS_DIR = "nodejs-assets";
```

achou11: fixed via https://github.com/digidem/mapeo-mobile/pull/974/commits/3efc786431201c791ce0071a899a589b914f0173

src/backend/index.js
```diff
@@ -16,162 +16,161 @@
...
+        server.close(() => {
+          status.setState(constants.CLOSED);
+          cb();
+        });
+      } else {
+        process.nextTick(cb);
+      }
+    }
+  } catch (error) {
+    console.log(error.stack);
```
achou11: is this intentional?

gmaclennan:
Sorry about pushing this with so many errors and not working, should test more thoroughly next time! I think this is all working now.

achou11 (approved):
great stuff - much easier to follow!

PR #975, opened 2022-06-01
chore: bugsnag error reporting improvements

gmaclennan:
Depends on #974

- Report same bugsnag release stage for frontend and backend
- Upload react native source maps
- Use bugsnag react error boundary
- Track feature flags with bugsnag


gmaclennan:
Merging this as tests pass and no major changes other than bug reporting, so that I can continue with PRs based on this one.

PR #989, opened 2022-06-24
feat: Add device id to About Mapeo screen

gmaclennan:
Not yet tested on a device.

Adds the device id to the About Mapeo screen in settings. The device ID is currently the public key of the writer hypercore for the device. This can be useful for debugging - matching the device to data in the hypercores


src/frontend/screens/Settings/AboutMapeo.js
```diff
@@ -87,6 +95,13 @@ const AboutMapeo = () => {
         />
         <DeviceInfoListItem label={t(m.androidBuild)} deviceProp="buildId" />
         <DeviceInfoListItem label={t(m.phoneModel)} deviceProp="model" />
+        <DeviceInfoListItem label={t(m.deviceId)} deviceProp="deviceId" />
+        <ListItem>
+          <ListItemText
+            primary={t(m.deviceId)}
+            secondary={displayDeviceId}
+          ></ListItemText>
+        </ListItem>
```
achou11, line 102: list item here doesn't seem necessary if you have the DeviceInfoListItem already

gmaclennan: `DeviceInfoListItem` is a specific wrapper for properties exposed by react-native-device-info.

achou11: ah got it 👍

achou11: tested this locally and somewhat confused about what we want to achieve with this. There are now two Device IDs being listed - one from react-native-device-info and the other from something in the backend? Question is what do we want to display in this case, and can we update the label for one of them to make it clearer about what it is?

image

gmaclennan: Looks like we need a better name. 

This is the public key of the device's writeable hypercore. It is useful for debugging when trying to establish which observations were created on which device. Useful when eg you are analysing a Mapeo project and find missing sparse data in particular hypercore and figuring out which device that is from.

achou11:
Updated this PR to catch up with develop branch and converted `AboutMapeo` to TS. Still need to come up with a better label name for the device hypercore id (see https://github.com/digidem/mapeo-mobile/pull/989#discussion_r988120831)

src/frontend/screens/Settings/AboutMapeo.js
achou11: got it! makes sense after reading the original PR description. few suggestions that come off the top of my head in case any may come relatively close:

- Device Core ID
- Device Data ID
- Device Hypercore ID
- Core ID
- Hypercore ID
- Core Data ID

PR #990, opened 2022-06-24
chore: git ignore random_seed file from git secret

achou11:
According to [git secret docs](https://git-secret.io/), we're not supposed to check in the `random_seed` file

> Note all the contents of the .gitsecret/ folder should be checked in, /except/ the random_seed file. This also means that of all the files in .gitsecret/, only the random_seed file should be mentioned in your .gitignore file.

Not entirely sure if it's safe to just remove the seed file at this point, or if it's somehow automatically generated locally when using git secret


gmaclennan (approved):
Fine (and good) to remove this. Ref. https://github.com/sobolevn/git-secret/issues/565

PR #992, opened 2022-06-27
React nav upgrade

ErikSin:
# Updates React Navigation v3 to v6

### To Do:
- [x] Update Header Component
- [x] Update `React-Navigation-Stack` to `@react-navigation/native-stack`
- [x] Update `React-Navigation-tabs` to `@react-navigation/bottom-tabs`
- [x] Add [Type safety](https://reactnavigation.org/docs/typescript) to Navigation
- [x] Update Onboarding Stack
- [x] Update Icca Stack
- [x] Update `AppContainer` with new navigation

## New Architecture to Note:
- new hook: `useSetHeader`: Header component can be directly edited inside the component (previously could only be set outside of the component). This hook reduces the boiler plate of `useLayoutEffect` and translation.
- `ICCA` nav stack and `Onboarding` Nav Stack have been merged into the main Navigation Stack. They are still only accessible if the environment feature flags are turned on, but are not their own separate navigation stacks that are nested. 
- `useNavigation` hook utilized the native navigation and are typesafe. The hooks should be imported for `useNavigationWithTypes` and NOT `@react-navigation/native`. It is impossible to determine statically what component is using the hook, they are not fully typesafe.
- Front end tests have been "updated". Currently our front end test are not developed. We are testing 2 pages, and don't seem to be testing anything of importance. I updated them so they pass with the new dependencies(aka i added new mocks)...but we should re haul and update them completely.


src/frontend/hooks/useSetHeader.tsx
```diff
@@ -0,0 +1,52 @@
+import { NavigationProp } from "@react-navigation/native";
```
ErikSin: In react nav 6, the header can be set directly in the component. This custom hooks reduces the boiler plate in translations and in `useLayoutEffect`

src/frontend/hooks/useNavigationWithTypes.ts
```diff
@@ -0,0 +1,4 @@
+import { useNavigation as useNavigationNative } from "@react-navigation/native";
+import { StackNavProp } from "../Navigation/AppStack";
+
+export const useNavigation = () => useNavigationNative<StackNavProp>();
```
ErikSin: the `useNavigation` hook utilizes native navigation AND is passed all the route name and params for typesafety. useNavigation should be imported from here, not `@react-navigation/native`

src/frontend/sharedTypes.ts
```diff
@@ -39,3 +41,9 @@ export type NavigationProp = NavigationStackProp<
 export type StackScreenComponent = NavigationStackScreenComponent<
   NavigationParams
 >;
+
+type NavKeys = keyof AppStackList;
+
+export type NativeNavigationProp<
+  ScreenName extends NavKeys
+> = NativeStackScreenProps<AppStackList, ScreenName>;
```
ErikSin: This is used to define the props of navigation screens (that are received from the navigator).

src/frontend/hooks/useSetHeader.tsx
```diff
@@ -0,0 +1,52 @@
...
+      headerRight: headerRight || undefined,
+      headerStyle: {
+        backgroundColor: backgroundColor || "#fff",
+      },
+      // For some reason, when headerTintColor is set, the back button is not getting the tint color. So I am setting it directly in custom header left.
+      headerLeft: props => (
+        <CustomHeaderLeft
+          headerBackButtonProps={props}
+          tintColor={headerTintColor}
+        />
```
ErikSin: header left should be recieving `headerTintColor` from `navigator.setOptions()`. The text is recieving the right tint color, BUT not the arrow. So to circumvent that, I am just re-rendering `CustomHeaderLeft` again with the `tintColor` param being updated

src/frontend/lib/styles.ts
```diff
@@ -1,5 +1,3 @@
-// @flow
-
 export const BLACK = "#000000";
```
ErikSin, line 1: Changed this file to typescript

src/frontend/Navigation/DefaultContainer.ts
```diff
@@ -1,4 +1,4 @@
-import { createAppContainer } from "react-navigation";
```
ErikSin, line 1: New Navigation does not use `AppContainer`. Instead a Navigation Container Component is used as a wrapper to all the navigation screens/children. All the `container` files can be deleted, just leaving it in here for now for importing purposes

package.json
```diff
@@ -89,17 +94,17 @@
...
+    "react-native-screens": "^3.13.1",
     "react-native-share": "^3.7.0",
     "react-native-splash-screen": "^3.2.0",
     "react-native-svg": "^12.1.1",
     "react-native-tab-view": "^2.15.1",
     "react-native-unimodules": "^0.14.10",
     "react-native-v8": "^0.66.3-patch.1",
     "react-native-vector-icons": "^9.1.0",
-    "react-navigation": "^4.4.0",
+    "react-navigation": "^4.4.4",
```
gmaclennan: What is this used for?

src/frontend/AppContainerWrapper.tsx
```diff
@@ -19,24 +19,33 @@ import {
...
+const createNavigationStatePersister = (navState?: NavigationState) => {
   if (featureFlagOn) return;
   try {
-    await AsyncStorage.setItem(NAV_STORE_KEY, JSON.stringify(navState));
+    if (!navState)
+      throw new Error("Navigation state is undefined while trying to save it");
+    AsyncStorage.setItem(NAV_STORE_KEY, JSON.stringify(navState));
   } catch (err) {
     log("Error saving navigation state", err);
   }
```
gmaclennan: I think during this refactor this might be a good time to switch away from this navigation persistence behaviour, and instead "if draft observation exists" -> navigate to observation edit screen, "else" -> don't load state, so would just start on home screen (map).
If you choose to keep the state persistence, then we need to ensure that NAV_STORE_KEY changes.

src/frontend/AppContainerWrapper.tsx
```diff
@@ -93,12 +92,41 @@ function hidePracticeModeTemporarily(route: string | null) {
...
+          if (state !== undefined) {
+            setInitialState(state);
+          }
+        }
+      } finally {
+        setIsReady(true);
+      }
+    };
+
+    if (!isReady) {
```
gmaclennan: I _think_ this might be safer to not reference the `isReady` value here, and not include it in the dependencies, so that this only runs once, but add a scoped `let cancelled = false` and in the useEffect return `cancelled = true`
I think it's ok as-is, but the code is a bit "loopy", settings something that causes the effect to re-run, and could lead to bugs in future maintenance.

src/frontend/AppContainerWrapper.tsx
```diff
@@ -108,45 +136,19 @@ const AppContainerWrapper = () => {
     []
   );
```
gmaclennan, line 50: This `updateRouteBasedAppState` should be derived from the routeName within the render, rather than through a callback and setting state. This is going to cause a lot of re-renders for each navigation change.

src/frontend/AppContainerWrapper.tsx
```diff
@@ -108,45 +136,19 @@ const AppContainerWrapper = () => {
...
-              return loadedNavState;
-            },
-            persistNavigationState: createNavigationStatePersister(),
-          },
+  const handleNavStateChange = React.useCallback(
+    (navState?: NavigationState) => {
+      if (!navState) return;
+      //TO DO: Check if prev route !== current route
+      updateRouteBasedAppState(getRouteName(navState));
+      createNavigationStatePersister(navState);
```
gmaclennan: Why do we want to create this (navigationStatePersister) every time this callback runs?

src/frontend/AppContainerWrapper.tsx
gmaclennan: e.g. store the routeName in state and derive everything else from that.

gmaclennan (commented):
I only just started this review looking at the `AppContainerWrapper` component, but thought I would share feedback so far, since I need to switch to other things today.

package.json
ErikSin: its not! when i was upgrading the navigation initially, i didnt realize that it was a completely different API. I initially just ran the upgrade command, which caused this. I will take this out once I remove it from all the pages.

src/frontend/AppContainerWrapper.tsx
ErikSin: Good point about the NAV_STORE_KEY.
 
I agree we should change the persistence behaviour, but I think we need to consult with programs team. I will get that process started (of requesting consultation from the programs team), but for now Ill just keep with our current navigation.

I think one thing we should do is move away from AsyncStorage, and during that refactor, we can move away from this persistence architecture

src/frontend/AppContainerWrapper.tsx
ErikSin: I agree, I actually just copied and pasted this directly from the [react-nav docs](https://reactnavigation.org/docs/auth-flow/#define-our-screens).

Im going to change it, i find it hard to read, and I prefer your method of using `let cancelled = false`

src/frontend/AppContainerWrapper.tsx
ErikSin: This is @achou11's architecture. Im curious what you both think. This app container needs the route name, but the route name is created inside the children of the appcontainer. So we could create another context, and pass the route name to that context....but then we are adding another context (which is not necessarily a problem). Is that what you had in mind @gmaclennan ?

src/frontend/AppContainerWrapper.tsx
ErikSin: I should have changed the name of this. `createNavigationStatePersister` should be changed to `persistNavState()`. I changed the architecture but did not change the function name

src/frontend/AppContainerWrapper.tsx
achou11: yeah gregor's right in that the state being set in that function should be probably be derived. Not exactly sure what my reasoning here was at the time - probably just too enamored with using effects 😅 but yeah, storing the route name in state and converting the states it updates to be derived should work?

ErikSin: @achou11 just for clarification, we would need to do this through a context, correct?

achou11: I dont think so? the route name state would live in this component and the calculated values are only needed here, right? i haven't looked at the updated implementation from your PR so apologies if i'm missing something

ErikSin: No, im confused, I misunderstood the original comment

ErikSin: But I think i understand it now. Instead of having `inviteModalEnabled`, `hidePracticeBar`, and `hidePracticeMode` hold state, we can just derive whether they are true or not directly from the route name (therefore avoiding unnecessary re-renders). Is that right?

achou11: yep!

src/frontend/AppContainerWrapper.tsx
gmaclennan: The code in the link you shared looks ok to me, I think the issues here come from including isReady as a dependency and having multiple setState calls.

src/frontend/AppContainerWrapper.tsx
gmaclennan: This is getting off topic (because of me) because this is not what this PR is about, but it's a chance to cleanup code. Does this definitely need to sit outside (e.g. a parent of) the AppContainer? Is there a way for practice mode and invite modal logic to be defined inside AppContainer, where routeName would be available as a prop / in context? That would avoid the slightly hairy logic of hoisting a navigation prop to the parent (which is essentially what this is doing). I don't think this needs to be resolved in this PR, but since this is my first proper review of this code I wanted to share thoughts on it.

src/frontend/AppContainerWrapper.tsx
ErikSin: Sorry Gregor, I shared the wrong link: https://reactnavigation.org/docs/state-persistence

I just copied and pasted from the docs. But I think I will change it regardless. It does feel very circular. I see why they did it this way (in order to have a ready state available outside of the function). But I can solve this problem with typescript

gmaclennan: Yeah, I see, it looks like it will _probably_ be ok, but it's got a bit of "code smell" to it to me.

ErikSin: Upgraded it here: https://github.com/digidem/mapeo-mobile/pull/992/files#diff-2cfb365dc18dfa0b63d430f262707c0a0ae6a3bb6a8cdffdfdc938754d1dfcb5R53

src/frontend/AppContainerWrapper.tsx
ErikSin: Unfortunately not. So the `AppContainer` is of type `Stack.Navigator`. `Stack.Navigator` can only have `Stack` Components as its children. As we soon as we put a non stack item(practice mode), it throws a run time error. We could maybe put the invite modal logic though.... @achou11 you designed that originally, what do you think?

src/frontend/Navigation/AppStack.tsx
```diff
@@ -32,115 +30,175 @@ import { AlreadyOnProj } from "../screens/AlreadyOnProject";
...
+      </RootStack.Group>
+    )}
+    {devExperiments.appPasscode && (
+      <RootStack.Group>
+        <RootStack.Screen name="NewPasscode" component={NewPasscode} />
+        <RootStack.Screen name="AppPasscode" component={AppPasscode} />
+      </RootStack.Group>
+    )}
+    {/* To do: Determine if ICCA */}
+    {false && (
```
ErikSin: How are we determining whether the ICCA flag variable is on? This needs to updated to reflect that, but I am not really sure how that's done

src/frontend/AppContainerWrapper.tsx
achou11: i haven't fully familiarized with the new navigation api but my underinformed understanding is that since practice mode is fundamentally a modification of the original stack i.e. same routes but with some UI adjustments and more limited access to certain screens, it may not make sense to create its own navigator for it and nest it due to potential duplication of screens. however, i do see the point of doing something like that to avoid having to account for some global state of "is in practice mode" for various screens, and it would make it easier to create the separation between modes from a development perspective i.e. a directory of screens that assume "normal" mode and another for "practice mode", where each can be built with the assumption of being under the correct navigation namespace/stack (e.g. `app/main/...` `app/practice/...`). only potential issues here may be 1) duplication and 2) sharing of screens (maybe this is not an issue though if properly architected). @ErikSin do you have a general sense of how many screens are explicitly affected by practice mode? if there are many, do you think creating a devoted stack navigator for it would be feasible? (maybe as a follow up PR)

Regarding the invite modal, i think that could be it's own navigation stack if it isn't already e.g. `app/invite/...`, so that there's less need for a global state to represent that. cant fully remember but are there things we disable that are unrelated to the invite flow when when we enter that flow?

achou11: > Stack.Navigator can only have Stack Components as its children

also as a note, you should be able to nest different kinds of navigators within each other https://reactnavigation.org/docs/nesting-navigators/. not sure if that comment was referring to something else though

ErikSin: Practice mode touches every page, so creating a new stack for it would just duplicate everything, so really not feasible. And yes, different navigators can be nested, but yeah not feasible for just a ui component like practice mode.

In regards to the invite modal....I was hoping you would remember, because I don't quite remember why we designed it the way we did 😬

achou11: reading through Gregor's comment again, realizing that he's referring to something slightly different than what i was expanding on. i suppose something like this would be feasible, although i'm not entirely sure if there are downsides to this approach:

```jsx
const AppContainerStack = createStackNavigator()
const AppStack = createStackNavigator()

const App = ({route, navigation}) => {
  // Use navigation and route params to do stuff related to invites, practice mode, etc
  return (
    <AppStack.Navigator>
      {/* All the screens for the main app */}
      <AppStack.Screen ... />
    <AppStack.Navigator
  )
}

const AppContainer = () => {
  return (
    <AppContainerStack.Navigator>
      <AppContainerStack.Screen 
        name="App"
        component={AppStack}
      />
    </AppContainerStack.Stack>
  )
}
```

EDIT: just realizing that this example as-is may not work since you probably wouldn't have granular route information about which route you're on in the `App` component, since it's based on the definition in the `AppContainer`. maybe there's a way of structuring it to something like this to make it work?

achou11: > Practice mode touches every page, so creating a new stack for it would just duplicate everything, so really not feasible

yeah sorry i meant **how many screens do something different because the app is in practice mode?** I understand that the practice mode UI is on most screens, but do all of the screens adjust internal logic because of the different mode? e.g. disabling or omitting certain UI or actions

achou11: but yeah i think it probably isn't reasonable to go that route of a practice mode stack vs just representing it in state. see https://github.com/digidem/mapeo-mobile/pull/992#discussion_r913971557 for a more pertinent response to gregor's question

ErikSin: Ohhh interesting! this might actually work...ill try it.

src/frontend/AppContainerWrapper.tsx
```diff
@@ -1,160 +1,88 @@
...
+        const initialUrl = undefined;
+
+        // if deeplinking, dont set initial state
+        if (initialUrl) {
+          setInitialNavState(undefined);
+          return;
+        }
+      } finally {
+        setInitialNavState(await loadSavedNavState(log));
+      }
```
achou11: can this be simplified to not use a try/finally?

src/frontend/AppContainerWrapper.tsx
```diff
@@ -1,160 +1,88 @@
...
+  const inviteModalEnabled = inviteModalDisabledOnRoute(currentRoute);
+  const hidePracticeBar = hidePracticeBarForRoute(currentRoute);
+  const hidePracticeMode = hidePracticeModeTemporarily(currentRoute);
+
   /**
    * TODO: Uncomment when project invites are supported
    */
   //   useProjectInviteListener(invite => {
-  //     if (inviteModalEnabled) {
+  //     if () {
```
achou11: can probably just remove this whole commented block. i kept it in here when i first implemented so that it'd be easier to remind myself whenever we got to it, but it's been so long that it's probably not worth keeping this :)

src/frontend/Navigation/AppStack.tsx
```diff
@@ -32,115 +30,175 @@ import { AlreadyOnProj } from "../screens/AlreadyOnProject";
...
+  Settings: undefined;
+  ProjectConfig: undefined;
+  AboutMapeo: undefined;
+  LanguageSettings: undefined;
+  CoordinateFormat: undefined;
+  Experiments: undefined;
+  PhotosModal: {
+    photoIndex: number;
+    observationId: string;
+    editing?: true;
```
achou11: just checking, but  is this supposed to be `boolean`?

src/frontend/Navigation/AppStack.tsx
achou11: yeah it's somewhat complicated the way it's done now, but i think it's set up so that ICCA is kind of its own target. so similar to how you can have target-specific files e.g. `AppStack.android.tsx` vs `AppStack.ios.tsx`, you'd have a file called `AppStack.icca.tsx` ([example](https://github.com/digidem/mapeo-mobile/blob/d3271092d75f6e8c82bc94af2bbec81fb7f3559e/src/frontend/Navigation/DefaultContainer.icca.ts)). Then when building/running the app, the [`RN_SRC_EXT` env variable is set to `icca`](https://github.com/digidem/mapeo-mobile/blob/d3271092d75f6e8c82bc94af2bbec81fb7f3559e/package.json#L17), which is used by [some logic in our metro config](https://github.com/digidem/mapeo-mobile/blob/d3271092d75f6e8c82bc94af2bbec81fb7f3559e/metro.config.js#L32-L42) to pull in that target-specific file and use that over the default file (e.g. `AppStack.tsx`)

not entirely sure it makes sense to follow that same approach, where you create an app stack specifically for iccas i.e. `AppStack.iccas.tsx` that includes the info+intro stacks/screens and let the build/metro process handle which one it uses. there was probably some important context that i'm forgetting that warranted creating a separate navigator for iccas specifically, but maybe with the updates to react navigation, it isn't necessary to do that anymore?

i'd probably wait for @gmaclennan's feedback on this since he has the most context, but the "safe" route would be to use the old approach i guess

src/frontend/Navigation/NavigationStateHelperFunctions.ts
```diff
@@ -0,0 +1,66 @@
...
+      log("Crashed on last open, skipping load of navigation state");
+      return undefined;
+    } else {
+      return (navState as InitialState) || undefined;
+    }
+  } catch (error) {
+    bugsnag.leaveBreadcrumb("Error loading nav state", { error });
+    log("Error reading navigation and error state", error);
+  }
+}
```
achou11: is the plan to move away from persisting/loading the nav state in a follow up PR? I think it's agreed that we don't really need or want to do this anymore and instead just implement some logic to navigate to the observation screen if a draft observation exists. if that's the case, these won't be needed (along with logic in the relevant navigators)

src/frontend/Navigation/NavigationStateHelperFunctions.ts
```diff
@@ -0,0 +1,66 @@
+import { InitialState, NavigationState } from "@react-navigation/native";
```
achou11: nit but would lower case the file title e.g. `navigationStateHelperFunctions.ts` to avoid potential confusion about this being a component. Not sure how big of a problem it is anymore, but i know some tools do/did assume that a capitalized file name in a React app means that it's a component

src/frontend/Navigation/OnboardingStack.tsx
```diff
@@ -16,57 +14,46 @@ import {
...
-    Main: MainStack,
-    JoinRequestModal,
-    ProjectInviteModal,
-  },
-  {
-    initialRouteName: "Main",
-    mode: "modal",
-    headerMode: "none",
-  }
+export const OnboardingStack = () => (
```
achou11: can we create a stack navigator for this and then nest it in the root stack? not sure if there is some complexity around that that i'm not aware of, but seems like a reasonable thing to do in this case.

src/frontend/screens/Observation/index.tsx
```diff
@@ -100,17 +104,17 @@ const Observation = ({ navigation }: Props) => {
...
+// }: {
+//   navigation: NavigationProp,
+// }) => ({
+//   headerTitle: () => (
+//     <HeaderTitle>
+//       <FormattedMessage {...m.title} />
+//     </HeaderTitle>
+//   ),
+//   headerRight: () => <ObservationHeaderRight navigation={navigation} />,
+// });
```
achou11: okay to remove?

src/frontend/AppContainerWrapper.tsx
ErikSin: I just tried this. unfortunately the parent navigator does NOT get the state of the child navigator (so it wont work for the practice mode) But It should work for the invite modal logic. Ill refactor the onboarding stack to be in a parent navigator.

src/frontend/screens/ObservationEdit/index.tsx
```diff
@@ -21,14 +20,23 @@ const m = defineMessages({
...
-      // new observation (which starts with the category chooser screen)
-      key: "fromObservationEdit",
-    });
+    navigation.navigate("CategoryChooser");
+    //   {
+    //   routeName: "CategoryChooser",
+    //   // Set a key here so we don't navigate back in the stack when creating a
+    //   // new observation (which starts with the category chooser screen)
+    //   key: "fromObservationEdit",
+    // });
```
achou11: do we no longer need this?

src/frontend/sharedComponents/BottomSheetModal/index.tsx
```diff
@@ -15,6 +16,16 @@ export const MODAL_NAVIGATION_OPTIONS = {
...
+export const useSetModalNavOptions = () => {
+  const { setOptions } = useNavigationFromRoot();
+  return React.useLayoutEffect(() => {
+    setOptions({
+      presentation: "transparentModal",
+      animation: "none",
+    });
+  }, []);
+};
+
```
achou11: unless this will change at some point, i don't think this is necessary. can instead set these options on the screens themselves e.g. when you use `WhateverStack.Screen`

similar comment about `useSetHeader` in general. that hook really only needs to be called if the component is setting something dynamic like a translated string. otherwise, it should be passed as an option when creating the screen in the stack component

achou11: i suggest updating `MODAL_NAVIGATION_OPTIONS` in this file to the options here and then using that export in the onboarding stack as the options you pass for the relevant screens (JoinRequest and ProjectInvite modals)

achou11 (changes requested):
Overall looks really good so far!

- left a comment about the ICCA variant as it's something that this PR doesn't fully address yet. i think we need to stick with the old approach of using a filename so that screens used in ICCA are included when bundled. there may be a way of doing that without using the extension approach though.
- there's a good number of `useSetHeader` calls that are probably unnecessary, namely when it's just for a static option that's being set e.g. `useSetHeader({ headerShown: false });`. I would lean towards setting options like that where `Stack.Screen` for the relevant screen is defined (usually in one of the stack components). Similar comment about the `useSetModalNavOptions` in the `BottomSheetModal` component
- note for other readers: the animation for moving between screens is different since we're using the native stack's default. it's possible to change this to something similar to the old one ([link](https://reactnavigation.org/docs/native-stack-navigator#animation), although adjusting the duration doesn't seem to work despite what the (confusing) [docs](https://reactnavigation.org/docs/native-stack-navigator#animationDuration) say

src/frontend/AppContainerWrapper.tsx
ErikSin: Are you meaning like getting rid of the await, and just using then, catch, and finally?

src/frontend/AppContainerWrapper.tsx
ErikSin: Will do

src/frontend/Navigation/AppStack.tsx
ErikSin: Yes, good catch

src/frontend/Navigation/NavigationStateHelperFunctions.ts
ErikSin: it is, BUT not for this PR because we need to get programs input

src/frontend/Navigation/NavigationStateHelperFunctions.ts
ErikSin: Wouldn't hurt to change it!

src/frontend/Navigation/OnboardingStack.tsx
ErikSin: I actually didnt create a separate stack because react navigation suggests only nesting navigators in cases where we want a separate ui NOT for breaking up the code, and suggests just using `stack.group`. That being said, I think i will make it its own navigator, to help with some of our state management for the invite modal

src/frontend/screens/ObservationEdit/index.tsx
ErikSin: Im not sure, i left it in here because i wanted to discuss this with you folks. I don't understand this comment, and I don't really know what they were trying to accomplish here because playing around with the UI, it seems fine even if we do no reset the stack

ErikSin:
> * there's a good number of `useSetHeader` calls that are probably unnecessary, namely when it's just for a static option that's being set e.g. `useSetHeader({ headerShown: false });`. I would lean towards setting options like that where `Stack.Screen` for the relevant screen is defined (usually in one of the stack components). 

So the reason I didn't do this, is because when we were using v3, we also had the option to turn off the header directly inside the app stack, but it was still programmed directly in the component. Im not sure if that was a preference thing OR if there was a reason to do that. But, I also would lean to statically setting it, than dynamically setting it in the hook.

src/frontend/AppContainerWrapper.tsx
achou11: no i mean getting rid of the wrapping try/finally. Doesn't seem like it's needed in this blocks current state, under the assumption that we don't want to set the initial nav state twice if we're handling a deeplink, which is what this is currently doing because of the finally. 

```js
const initialUrl = undefined; // or await Linking.getInitialUrl(); when it's working

if (initialUrl) {
  setInitialNavState(undefined);
  return;
}

setInitialNavState(await loadSavedNavState);
```

achou11: if the try statement was for when the linking api is available, then i wouldn't use finally and instead use catch

src/frontend/Navigation/OnboardingStack.tsx
achou11: that's good to know, thanks for the explanation!

src/frontend/Navigation/AppStack.tsx
ErikSin: Right, that makes sense. Happy to turn this back to its own file, but ill wait for gregors input just so we are doing this correctly

src/frontend/screens/ObservationEdit/index.tsx
achou11: my guess is that it's probably to account for this: https://reactnavigation.org/docs/navigation-prop#going-back-from-a-specific-screen

if that's the case, maybe keeping it will be safer?

src/frontend/AppContainerWrapper.tsx
ErikSin: ahhh yes, that is much clearer...the try/finally was remnants of the documentation, but I changed the architecture quite a bit....ill change it to how you have it!

src/frontend/Navigation/OnboardingStack.tsx
ErikSin: Just for visibility. As Andrew and I discussed on Slack, we actually are NOT going to create a separate stack for onboarding because it does not solve our problem of having the nav state available to the project invite modal. The nav state for children is only available to that child and not the parent or siblings.

 AND following react navigation suggestion, having a nested navigator is superflous unless the UI is different.

src/frontend/AppContainerWrapper.tsx
ErikSin: For visibility, its does not actually work for the invite modal logic: re: [this comment](https://github.com/digidem/mapeo-mobile/pull/992#discussion_r915255574)

achou11: just adding this snippet from a conversation with Erik that explains this rather convoluted setup right now. i had thought maybe using a [ref](https://reactnavigation.org/docs/navigation-container#methods-on-the-ref) could potentially solve the two problems at hand, but it only solves one:

> there are two problems that we’re trying to solve in an ideal situation:
>
> 1. we need to disable being able to open the invite modal flow depending on the screen in the root stack we’re in
> 2. we need to update the practice mode UI based on the route because it’s slightly different for certain screens
>
> 1 would be feasible to fix if we used the ref approach since we’re listening on an event listener for invites that are received. with the ref, we could determine the route the app is in when an processing an invite, and if the route is okay to show the invite modal, show it. in other words, the source of truth for our action is not the route, but the invite event
>
> 2 is harder and doesn’t benefit from the ref approach because the source of truth for our action is the route changing. using a ref wouldn’t cause state updates and thus rerenders that the practice mode UI depends on

achou11:
> > * there's a good number of `useSetHeader` calls that are probably unnecessary, namely when it's just for a static option that's being set e.g. `useSetHeader({ headerShown: false });`. I would lean towards setting options like that where `Stack.Screen` for the relevant screen is defined (usually in one of the stack components).
> 
> So the reason I didn't do this, is because when we were using v3, we also had the option to turn off the header directly inside the app stack, but it was still programmed directly in the component. Im not sure if that was a preference thing OR if there was a reason to do that. But, I also would lean to statically setting it, than dynamically setting it in the hook.

hm i see. i'd probably stick with doing it statically when possible if there isn't any difference in the UI (i can't really see any major upsides of keeping the previous approach). the only minor benefit for the old approach i can think of is colocation of the screen and its options. entirely possible that under the hood, it's doing the same thing anyways so maybe none of this is really a big deal 😄

src/frontend/screens/ObservationEdit/index.tsx
ErikSin: I left it in. They key is not being used anywhere... but ill keep it in for now

ErikSin:
> > > * there's a good number of `useSetHeader` calls that are probably unnecessary, namely when it's just for a static option that's being set e.g. `useSetHeader({ headerShown: false });`. I would lean towards setting options like that where `Stack.Screen` for the relevant screen is defined (usually in one of the stack components).
> > 
> > 
> > So the reason I didn't do this, is because when we were using v3, we also had the option to turn off the header directly inside the app stack, but it was still programmed directly in the component. Im not sure if that was a preference thing OR if there was a reason to do that. But, I also would lean to statically setting it, than dynamically setting it in the hook.
> 
> hm i see. i'd probably stick with doing it statically when possible if there isn't any difference in the UI (i can't really see any major upsides of keeping the previous approach). the only minor benefit for the old approach i can think of is colocation of the screen and its options. entirely possible that under the hood, it's doing the same thing anyways so maybe none of this is really a big deal 😄

I agree, i just changed it to be static!

src/frontend/sharedComponents/BottomSheetModal/index.tsx
ErikSin: updated!

src/frontend/AppContainerWrapper.tsx
```diff
@@ -1,189 +1,105 @@
...
+// import { Linking } from "react-native";
+import Loading from "./sharedComponents/Loading";
+import {
+  getRouteName,
+  hidePracticeBarForRoute,
+  hidePracticeModeTemporarily,
+  inviteModalDisabledOnRoute,
+  loadSavedNavState,
+  persistNavigationState,
+} from "./Navigation/navigationStateHelperFunctions";
```
gmaclennan, line 23:
```suggestion
} from "./Navigation/NavigationStateHelperFunctions";
```

src/frontend/AppContainerWrapper.tsx
```diff
@@ -1,189 +1,105 @@
...
-          if (nav) {
-            navRef.current = nav;
-          }
-        }}
-        uriPrefix={URI_PREFIX}
-      />
+      <NavigationContainer
+        initialState={initialNavState}
+        onStateChange={handleNavStateChange}
+        linking={{ prefixes: [URI_PREFIX] }}
```
gmaclennan, line 101: [Don't we need](https://reactnavigation.org/docs/navigation-container#ref) `ref={navRef}` here?

src/frontend/AppContainerWrapper.tsx
```diff
@@ -1,189 +1,105 @@
...
-
-              return loadedNavState;
-            },
-            persistNavigationState: createNavigationStatePersister(),
-          },
-    [updateRouteBasedAppState]
+  const handleNavStateChange = React.useCallback(
+    (navState?: NavigationState) => {
+      if (!navState || IS_E2E) return;
+      setCurrentRoute(getRouteName(navState));
```
gmaclennan: Rather than doing this in a callback, can we not do, in the main body of the function:

```js
const currentRoute = getRouteName(navRef?.getRootState())
```

this way we avoid the extra render (caused by the set state) for every navigation change.  Could also use the `navRef.getCurrentRoute()` method, might be simpler.

src/frontend/AppContainerWrapper.tsx
ErikSin: Refer to @achou11's [comment](https://github.com/digidem/mapeo-mobile/pull/992#discussion_r914064655)

gmaclennan: Fine with that logic, but the filename needs changed then. I changed this because CI is failing because the case of the import does not match the case of the filename (it works on MacOS because the default filesystem is case-insensitive). If you need help changing case in git with a case-insensitive filesystem, give me a holler (it's a pain!)

gmaclennan:
> * left a comment about the ICCA variant as it's something that this PR doesn't fully address yet. i think we need to stick with the old approach of using a filename so that screens used in ICCA are included when bundled. there may be a way of doing that without using the extension approach though.

The reason for doing the ICCA variant via the file extension is to have a way to choose the variant at build time, rather than runtime, so that ICCA resources are not included in the mainline version of the app. Other techniques I'm aware of would result in ICCA resources being included in the non-ICCA variant app bundle.

ErikSin:
> The reason for doing the ICCA variant via the file extension is to have a way to choose the variant at build time, rather than runtime, so that ICCA resources are not included in the mainline version of the app. Other techniques I'm aware of would result in ICCA resources being included in the non-ICCA variant app bundle.

In this case, I would just need to create an `AppStack.icca.tsx` file? Or is there anything else I would need to do?

src/frontend/Navigation/AppNavigator.tsx
```diff
@@ -0,0 +1,13 @@
+import { createNativeStackNavigator } from "@react-navigation/native-stack";
+import * as React from "react";
+import { AppStack } from "./AppStack";
+
+export const AppNavigator = () => {
+  const RootStack = createNativeStackNavigator();
```
gmaclennan: This should be created outside the component.

src/frontend/Navigation/AppNavigator.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+  return (
+    <RootStack.Navigator screenOptions={() => ({ headerShown: false })}>
+      <RootStack.Screen name="App" component={AppStack} />
+    </RootStack.Navigator>
```
gmaclennan: I think it might be best to avoid nested stack navigators (here we have RootStack, and AppStack nested inside it). I understand the challenge with Mapeo for ICCAs. Is it possible to do something similar using `RootStack.Group` - e.g. exporting a `RootStack.Group` from the `AppStack` component, and similarly exporting a group from the IccaStackNav? See https://reactnavigation.org/docs/nesting-navigators#best-practices-when-nesting

src/frontend/hooks/useIsFullyFocused.ts
```diff
@@ -32,18 +30,12 @@ export const useIsFullyFocused = () => {
...
-        setIsNavigationFullyFocused(true)
-      ),
-      navigation.addListener("didBlur", () =>
-        setIsNavigationFullyFocused(false)
-      ),
+      navigation.addListener("focus", () => setIsNavigationFullyFocused(true)),
+      navigation.addListener("blur", () => setIsNavigationFullyFocused(false)),
     ];
-
-    return () => navSubscriptions.forEach(s => s.remove());
```
gmaclennan, line 46: Need to keep this unsubscribe in the new effect

src/frontend/sharedComponents/CustomHeaderLeft.tsx
```diff
@@ -86,27 +88,39 @@ const HeaderBackIcon = ({ tintColor }: { tintColor: string }) => (
...
+  onPress?: () => void;
+  tintColor?: string;
+  headerBackButtonProps: HeaderBackButtonProps;
+}
+
+const CustomHeaderLeft = ({
+  onPress,
+  tintColor,
+  headerBackButtonProps,
+}: CustomHeaderLeftProps) => {
```
gmaclennan, line 102: I can't actually leave a comment on it, but it looks like there is a new way to prevent navigation (e.g. to show a "are you sure" warning) in RN6: https://reactnavigation.org/docs/preventing-going-back - this could be an alternative to our current implementation of `useBackHandler`, but I think it's ok to leave this to a follow-up PR.

gmaclennan (commented):
Great work on this @ErikSin, I can see a lot of work has gone into this and the code is in a really good place now.

I've left a few comments in the code, at least one of which can be done in a follow up PR. The other more general question I had was about our custom header implementation:

1. I know we use a custom back button, I think the original reason was for styling purposes and to improve the behaviour of the back button. However, for cases where all we want for the top-left icon is the back arrow, I think it might be better to stick with the default? With our custom header left the touch ripple is appearing in the wrong place anyway, and when testing locally it felt like the default header left responded better to touch events.
2. I see you have the `useSetHeader()` hook in most components. I think this is a really nice abstraction for `navigation.setOptions`, however it's only necessary for components where the screen needs access to the components in the header (e.g. in the case of a save button). Because it is on every screen we have an extra `useLayoutEffect` firing for each screen, which will likely be a performance hit for low-end devices. I think it might be better to default to using the standard screen options, and only using the `useSetHeader` hook for screens where we need to have the header components in the scope of the screen component. The downside of that approach however is in terms of code organization: it would move the header titles etc. into the AppStack component, rather than being co-located with the screen. I can think of two solutions to this:
    1. export the header options (e.g. title) as static properties on each screen component (e.g. for the `ObservationsList` screen you would add the line `ObservationsList.navigationOptions.title = m.observationsListTitle` just before the export, and then use this static property in the AppStack.
    2. export the whole stack screen from each screen component, e.g. using the same `ObservationsList` screen example, rather than export `ObservationsList`, you would export:
        ```
        export default () => <RootStack.Screen
          name="ObservationList"
          component={ObservationsList}
          options={{ headerTitle: m.observationsListTitle }}
        />
        ```
  However, I think that point (2) about useSetHeader is a performance optimization and can be postponed to a follow up PR, since this seems to work as is right now.

src/frontend/hooks/useIsFullyFocused.ts
ErikSin: good catch!

ErikSin: Following the pattern suggested by React Nav: https://reactnavigation.org/docs/navigation-events/#navigationaddlistener

src/frontend/Navigation/AppNavigator.tsx
ErikSin: I tried this, and ran into compile time errors. Exporting/Importing a `NativeNavStack` and `NativeNavStack.group` are being imported/exported as general `jsx.component`. The `navigator` won't take the general `jsx.component`, and therefore won't compile...Ive tried a few things, but if you have any suggestions on how to override this, let me know.

ErikSin: Refer to this issue: https://github.com/react-navigation/react-navigation/issues/9578

ErikSin: I changed it to be these nested groups... the syntax is very odd, but it work!

src/frontend/AppContainerWrapper.tsx
```diff
@@ -34,6 +35,8 @@ const AppContainerWrapper = () => {
     InitialState | "loading" | undefined
   >("loading");
 
+  const [config] = React.useContext(ConfigContext);
+
```
gmaclennan: I don't see where this is used?

src/frontend/hooks/useIsFullyFocused.ts
```diff
@@ -30,12 +30,14 @@ export const useIsFullyFocused = () => {
...
-    let navSubscriptions: (() => void)[] = [];
+    // let navSubscriptions: EventListenerCallback<any, any>  [] = [];
 
-    navSubscriptions = [
+    const navSubscriptions = [
       navigation.addListener("focus", () => setIsNavigationFullyFocused(true)),
       navigation.addListener("blur", () => setIsNavigationFullyFocused(false)),
     ];
+
+    return navSubscriptions.forEach(s => s);
```
gmaclennan: This is returning an array of functions. As far as I know, `useEffect` expects a single function to be returned, which is called when the dependencies change.

```suggestion
    return () => navSubscriptions.forEach(s => s());
```

src/frontend/Navigation/AppNavigator.tsx
```diff
@@ -1,13 +1,17 @@
...
-      <RootStack.Screen name="App" component={AppStack} />
+    <RootStack.Navigator
+      initialRouteName="Home"
+      screenOptions={route => ({
+        ...NavigatorScreenOptions,
+        headerShown: route.route.name !== "Home",
+      })}
+    >
+      {/* Refer to this issue for this odd syntax: https://github.com/react-navigation/react-navigation/issues/9578 */}
+      {(() => AppStack())()}
```
gmaclennan: Yes this is odd! This is actually functionally equivalent to just `{AppStack()}`. What this is doing is just calling the `AppStack` component as a function, returning the React Element(s) that is(are) returned by that function. As a consequence Hooks will not work inside AppStack because it is not actually a component, it is just being used as a function that returns a React Element.

All of this is a bit messy and confusing! I made a commit to make this more explicit and adds a comment explaining this to help future devs. I moved the `devExperiments` check into the `Navigator` component, since there it can be swapped with a hook if we want to define this at runtime in the future (you can't use hooks in the "createScreenGroup" functions, because they are functions that return elements, not components).

src/frontend/hooks/useSetHeader.tsx
```diff
@@ -0,0 +1,71 @@
...
+          ),
+      headerTintColor: headerTintColor,
+      headerShown,
+    });
+  }, [navigation, t, titleOrOptions]);
+};
+
+function isMessageDescriptor(value: any): value is MessageDescriptor {
+  return !value ? false : "defaultMessage" in value;
+}
```
gmaclennan: This throws if `value` is a string. Use `unknown` so that Typescript catches this kind of error.

```suggestion
function isMessageDescriptor(value: unknown): value is MessageDescriptor {
  return typeof value === "object" && !!value && "defaultMessage" in value;
}
```

src/frontend/Navigation/AppNavigator.tsx
```diff
@@ -0,0 +1,41 @@
...
+// update them efficiently.
+
+export const AppNavigator = () => {
+  return (
+    <RootStack.Navigator
+      initialRouteName="Home"
+      screenOptions={route => ({
+        ...NavigatorScreenOptions,
+        headerShown: route.route.name !== "Home",
+      })}
```
gmaclennan: Consider setting this as the `options` prop on the Home `<RootStack.Screen>` component (same in ICCA navigator). I'm not sure the legacy reason for this being here, but it is not required with RNav6

gmaclennan (changes requested):
Looks very close, just a couple of changes needed before we merge.

ErikSin:
>  I know we use a custom back button, I think the original reason was for styling purposes and to improve the behaviour of the back button. However, for cases where all we want for the top-left icon is the back arrow, I think it might be better to stick with the default? With our custom header left the touch ripple is appearing in the wrong place anyway, and when testing locally it felt like the default header left responded better to touch events.


I believe this was for size? The default one is very small, and cant be made any bigger.

ErikSin:
> > I know we use a custom back button, I think the original reason was for styling purposes and to improve the behaviour of the back button. However, for cases where all we want for the top-left icon is the back arrow, I think it might be better to stick with the default? With our custom header left the touch ripple is appearing in the wrong place anyway, and when testing locally it felt like the default header left responded better to touch events.
> 
> I believe this was for size? The default one is very small, and cant be made any bigger.

Here are screenshot for comparison: 

### Custom Header:
image

### Default Header:
image


ALSO,

The ripple was distorted due to padding, I changed it to margin and seems to have fixed the problem

src/frontend/hooks/useSetHeader.tsx
```diff
@@ -39,9 +38,31 @@ export const useSetHeader = (
...
+        headerLeft: props => (
+          <CustomHeaderLeft
+            headerBackButtonProps={props}
+            tintColor={headerTintColor}
+          />
+        ),
+        headerTintColor: headerTintColor,
+        headerShown,
+      });
+    }, [navigation, t, titleOrOptions]);
```
gmaclennan: You [can't have a hook inside a conditional](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level), it will break.

gmaclennan (changes requested):
The hook inside a conditional statement needs changed, but otherwise this looks good.

src/frontend/hooks/useSetHeader.tsx
```diff
@@ -0,0 +1,83 @@
...
+) => {
+  const { formatMessage: t } = useIntl();
+  const navigation = useNavigationFromRoot();
+
+  if (isMessageDescriptor(titleOrOptions)) {
+    return React.useLayoutEffect(() => {
+      navigation.setOptions({
+        headerTitle: t(titleOrOptions),
+      });
+    }, [navigation, t, titleOrOptions]);
```
gmaclennan: This also won't work inside a conditional

ErikSin:
@gmaclennan got rid of the `useSetHeader()` and added the translatable objects as static properties of the components so that they could be used in the `AppStack`!

src/frontend/screens/ObservationDetails/QuestionNavBar.js
```diff
@@ -1,63 +0,0 @@
-import React from "react";
```
gmaclennan, line 1: From what I gather, this whole file is deleted because it was unused? Just checking.

src/frontend/hooks/useNavigationWithTypes.ts
```diff
@@ -0,0 +1,19 @@
...
+  useNavigation<NativeStackNavigationProp<AppStackList>>();
+
+export function useNavigationFromHomeTabs() {
+  return useNavigation<
+    CompositeNavigationProp<
+      BottomTabNavigationProp<HomeTabsList>,
+      NativeStackNavigationProp<AppStackList>
+    >
+  >();
+}
```
gmaclennan, line 19: Just for my own learning, can you explain what is happening here and why two different hooks?

e2e/firstTest.e2e.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-env detox/detox, jest/globals */
 
-import { byId, byText } from "./matcher";
+import { byId, byLabel, byText } from "./matcher";
```
gmaclennan:
```suggestion
import { byId, byText } from "./matcher";
```

lint fix - unused var

src/frontend/screens/GpsModal.tsx
```diff
@@ -133,26 +128,17 @@ const GpsModal = ({ navigation }: Props) => {
...
   if (!location.savedPosition && !location.position) return "None";
-  const lastTimestamp = location.position
+  const lastTimestamp = !!location.position
     ? location.position.timestamp
-    : // $FlowFixMe
-      location.savedPosition.timestamp;
+    : !!location.savedPosition
+    ? location.savedPosition.timestamp
+    : "";
   return new Date(lastTimestamp).toLocaleString();
```
gmaclennan: This logic gets slightly confusing. Suggestion:

```js
  const { savedPosition, position } = location
  if (position) return new Date(position.timestamp).toLocaleString();
  if (savedPosition) return new Date(savedPosition.timestamp).toLocaleString();
  return "None"
```

src/frontend/screens/ManualGpsScreen/index.tsx
```diff
@@ -75,46 +75,55 @@ const ManualGpsScreen: NavigationStackScreenComponent<{
...
+    } catch (err) {
+      if (err instanceof Error) {
+        ToastAndroid.showWithGravity(
+          err.message,
+          ToastAndroid.LONG,
+          ToastAndroid.TOP
+        );
       }
     }
+  }, [convertedData, updateDraft, navigation]);
```
gmaclennan:
```suggestion
  }, [convertedData, updateDraft, navigation, value]);
```

src/frontend/screens/Observation/index.tsx
```diff
@@ -51,12 +49,20 @@ const ObservationNotFound = () => (
...
+}) => {
+  const { observationId } = route.params;
+
+  React.useLayoutEffect(() => {
+    navigation.setOptions({
+      headerRight: () => (
+        <ObservationHeaderRight observationId={observationId} />
+      ),
+    });
+  }, [navigation, observationId, ObservationHeaderRight]);
```
gmaclennan:
```suggestion
  }, [navigation, observationId]);
```

Unnecessary dep (out of context variable)

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -167,14 +165,6 @@ export const BackgroundMaps: NavigationStackScreenComponent = ({
   );
 };
 
-BackgroundMaps.navigationOptions = {
```
gmaclennan, line 170: Missing:

```js
BackgroundMaps.navTitle = m.title;
```

src/frontend/screens/__tests__/Settings.test.tsx
```diff
@@ -33,7 +33,8 @@ jest.mock("react-navigation-hooks", () => ({
 // };
 
 test("Settings Page will navigate", () => {
-  const { getByText } = render(<Settings />);
-  fireEvent(getByText("Project Configuration"), "press");
-  expect(mockNavigate).toHaveBeenCalledWith("ProjectConfig");
+  const settings = render(<Settings />);
+  // fireEvent(getByText("Project Configuration"), "press");
+  expect(settings).toBeTruthy();
+  // expect(mockNavigate).toHaveBeenCalledWith("ProjectConfig");
```
gmaclennan, line 39: Make sure to open an issue to put this test back.

gmaclennan (changes requested):
Good work on this, this is a big amount of work, but I think the improvements of the  new navigation are worth it. I went over all files just to be sure nothing had slipped through (since this is such a big change) and there are a couple of small issues, but happy for you to merge without further review once those issues are fixed.

src/frontend/hooks/useNavigationWithTypes.ts
ErikSin: In simplest terms, the use navigation hooks takes the screen names as a type in order to provide type-safety when using the hooks. So when you use the `useNavigationFromRoot`, you can only navigate to a screen in `AppStackList`.

The reason we have 2 is for the nested navigators. If your in the home tabs: you can either navigate to `camera`, `map` OR `root, {screen:screen-name}` screen. Since the root is nested, this second hook just makes it so the developer has to explicitly code that they are navigating into a nested navigator (`root`) and which screen they are navigating into

src/frontend/screens/ObservationDetails/QuestionNavBar.js
ErikSin: Yeah, someone had hardcoded a different one that is being used directly into the obverstionDetails/index!

PR #1000, opened 2022-09-07
Feat: app passcode screens

ErikSin:
## App Passcode (Screens only)

Introduces the screen for user to set app passcode, confirmation of passcode, and several error messages.

Screens can be found under settings>security>app passcode

This feature is under the feature flag `FEATURE_PASSCODE`

image

image

image

## To Do
- Create State Management of app passcode
- Create Authentication flow with navigation


achou11:
more of a meta topic but since this isn't the actual feature being incorporated just yet, maybe it makes sense to not name the eventually merged commit here according to conventional commit standards? i.e. leave off the `feat:` prefix.

thinking about this in the context of a useful changelog entry. I'd imagine that when this feature is actually usable is when we'd want this to show up in the changelog, not necessarily all the parts that lead to this. Thoughts?

src/frontend/screens/AppPasscode/InputPasscode.tsx
```diff
@@ -0,0 +1,198 @@
...
+import Text from "../../sharedComponents/Text";
+
+const m = defineMessages({
+  titleSet: {
+    id: "screens.AppPasscode.NewPasscode.InputPasscodeScreen.TitleSet",
+    defaultMessage: "Set App Passcode",
+  },
+  initialPassError: {
+    id: "screens.AppPasscode.NewPasscode.InputPasscodeScreen.initialPassError",
+    defaultMessage: "Password Must be 5 numbers",
```
achou11: This should probably be sentence-cased

src/frontend/screens/AppPasscode/InputPasscode.tsx
```diff
@@ -0,0 +1,198 @@
...
+export const InputPasscode = ({
+  screenState,
+  setScreenState,
+}: SetPasscodeProps) => {
+  const [error, setError] = React.useState(false);
+  const initialPassword = React.useRef("");
+  const { navigate } = useNavigationFromRoot();
+
+  const [inputtedPass, setInputtedPass] = React.useState("");
+  const ref = useBlurOnFulfill({ value: inputtedPass, cellCount: CELL_COUNT });
```
achou11: maybe use a more descriptive variable name here? e.g. `inputRef`

src/frontend/sharedComponents/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+export const CELL_COUNT = 5;
+const onlyNumRegEx = new RegExp("^[0-9]+$");
+
+interface PasswordInputProps {
+  stylesProps?: StyleProp<ViewStyle>;
+  inputValue: string;
+  onChangeTextWithValidation: (newVal: string) => void;
+}
+
+export const PasswordInput = React.forwardRef<TextInput, PasswordInputProps>(
```
achou11: nit: Since this only allows numbers, maybe rename this component (and related types, files, etc) to `PasscodeInput`

src/frontend/sharedComponents/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+          onLayout={getCellOnLayoutHandler(index)}
+        >
+          {textChild}
+        </Text>
+      );
+    }
+
+    return (
+      <CodeField
+        key={2}
```
achou11: what's this key for?

src/frontend/sharedComponents/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+  },
+  focusCell: {
+    borderColor: DARK_GREY,
+    textAlignVertical: "top",
+  },
+  codeFieldRoot: {
+    maxWidth: 289,
+    alignSelf: "center",
+  },
+});
```
achou11: a few magic numbers here are generally not necessary (and may not work well depending on varying screen widths). think the goal with a shared component like this should be to allow its consumer to define its layout without it causing visual issues within the component itself, so improving responsiveness will make this more reusable/adaptable.

I have some local changes to help out with this is you're interested.

src/frontend/screens/AppPasscode/InputPasscode.tsx
```diff
@@ -0,0 +1,198 @@
...
+        style={styles.button}
+        onPress={() => validateAndSetScreen(screenState)}
+      >
+        <Text style={styles.buttonText}>
+          <FormattedMessage {...m.button} />
+        </Text>
+      </Button>
+    </View>
+  );
+};
```
achou11: some style changes here can help reduce the number of magic numbers used and make this more layout-responsive. hard to elaborate on the specific changes but happy to set up a branch to provide an example

src/frontend/sharedComponents/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+interface PasswordInputProps {
+  stylesProps?: StyleProp<ViewStyle>;
+  inputValue: string;
+  onChangeTextWithValidation: (newVal: string) => void;
+}
+
+export const PasswordInput = React.forwardRef<TextInput, PasswordInputProps>(
+  ({ stylesProps, inputValue, onChangeTextWithValidation }, ref) => {
+    const [props, getCellOnLayoutHandler] = useClearByFocusCell({
```
achou11: nit: maybe rename `props` to something more descriptive? e.g. `codeFieldProps`

src/frontend/screens/AppPasscode/InputPasscode.tsx
achou11: also i thought i had set up a shared component that provides a template for these kinds of layouts (basically big action button at the bottom), but I can't find it. maybe i only did it specifically for the bottom sheet content, so maybe future todo item is to extract it so other screens can use it too

src/frontend/sharedComponents/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+        </Text>
+      );
+    }
+
+    return (
+      <CodeField
+        key={2}
+        ref={ref}
+        autoFocus={true}
+        {...props}
```
achou11: nit: to avoid subtle bugs related to jsx props and object spreading, maybe move this so that it's the first prop passed?

src/frontend/sharedComponents/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+        keyboardType="number-pad"
+        textContentType="oneTimeCode"
+        renderCell={renderCell}
+      />
+    );
+  }
+);
+
+const styles = StyleSheet.create({
+  cell: {
```
achou11: the input cursor isn't vertically centered but adjusting the `lineHeight` here can fix that. adjusting to `24 * 1.8` seems to work well (24 is font size here)

achou11 (changes requested):
Mostly noted changes around styling and simplifying that in certain areas.

Tangential usability concern that may not be within the scope of this PR:

I feel like it would be helpful to have an option to "unmask" the input here. as it's currently implemented/designed, if i make any mistakes related to what i input, the only way i can feel confident about not having made a mistake is to basically clear it and start over again, which can then lead to me repeating this again and again. for something that can potentially lock you out and is pretty prone to mistakes when setting up, think this could be a helpful addition to the design.

ErikSin:
> Mostly noted changes around styling and simplifying that in certain areas.
> 
> Tangential usability concern that may not be within the scope of this PR:
> 
> I feel like it would be helpful to have an option to "unmask" the input here. as it's currently implemented/designed, if i make any mistakes related to what i input, the only way i can feel confident about not having made a mistake is to basically clear it and start over again, which can then lead to me repeating this again and again. for something that can potentially lock you out and is pretty prone to mistakes when setting up, think this could be a helpful addition to the design.

I would bring this up with Sabella to discuss at a ui weekly.

src/frontend/sharedComponents/PasswordInput.tsx
ErikSin: What numbers are you referring to specifically?

ErikSin:
> more of a meta topic but since this isn't the actual feature being incorporated just yet, maybe it makes sense to not name the eventually merged commit here according to conventional commit standards? i.e. leave off the `feat:` prefix.
> 
> thinking about this in the context of a useful changelog entry. I'd imagine that when this feature is actually usable is when we'd want this to show up in the changelog, not necessarily all the parts that lead to this. Thoughts?

No, that makes sense. I will rename my commits

src/frontend/sharedComponents/PasswordInput.tsx
achou11: i was referring to `49` and `289`.

After playing around with it, the 49 seems fine. The 289 didn't feel necessary to define here when I played around with it. Defining that as part of the styleProps or just letting the consumer determine that width (potentially relying on flex behavior) made this number feel less necessary

ErikSin: Right! So the 49 came directly from figma. I think it is directly proprtional to font size, aka 2*24+1. Being a fixed size feels fine to me (like you mentioned). TBH i think the 289 is a typo (....i built this a long time ago). But according to [Sabella's Design Specs](https://www.notion.so/digidem/d2b2903e77fc4a8cae6d6a8467e25b59?v=def381515abe45578065598c05682c75&p=eec60615a2ac4317b52ca2ea01f2465d&pm=c), button should have a fixed with of 280. Why i chose max width, Im unsure....

ErikSin: OHHH nvm, im confusing myself, this is not a button

ErikSin: Ill play around with it and figure out why i chose that number, and try taking it out....

src/frontend/screens/AppPasscode/InputPasscode.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+import { RED, WHITE } from "../../lib/styles";
+import Button from "../../sharedComponents/Button";
+import {
+  CELL_COUNT,
+  PasswordInput,
+} from "../../sharedComponents/PasscodeInput";
+import Text from "../../sharedComponents/Text";
+
+const m = defineMessages({
```
achou11, line 17: think we should be consistent with the wording and stick to "passcode" for everything, as opposed to "password"

ErikSin: Lets wait...we are submitting the copy deck to the programs team next week....so lets finalize that first before we change any of the language

PR #1004, opened 2022-09-08
feat: add colored border to categories in category chooser

luandro:
Currently using `2px` as border width for presets with the `color` field, and `0` for those without.

Screenshots:

Screenshot_20220908-091718

Example with `1.5px` border width:

Screenshot_20220908-083215


src/frontend/sharedComponents/icons/Circle.js
```diff
@@ -4,16 +4,26 @@ import { StyleSheet, View } from "react-native";
...
+  radius = 25,
+  style,
+  children,
+}: CircleProps) => (
   <View
     style={[
       styles.circle,
       {
+        borderWidth: color === "#EAEAEA" ? null : borderWidth,
+        borderColor: color,
```
achou11: small likelihood and relatively minor, but this is prone to a bug where if the color is defined in the preset as `#EAEAEA`, the border width will be set to null as opposed to `borderWidth`. Would suggest removing the default assignment in the `color` prop and updating this part to something like this:

```suggestion
        borderWidth: color ? borderWidth : null,
        borderColor: color || `#EAEAEA`,
```

src/frontend/sharedComponents/icons/Circle.js
```diff
@@ -4,16 +4,26 @@ import { StyleSheet, View } from "react-native";
...
+  borderWidth = 2,
+  radius = 25,
+  style,
+  children,
+}: CircleProps) => (
   <View
     style={[
       styles.circle,
       {
+        borderWidth: color === "#EAEAEA" ? null : borderWidth,
```
achou11: May be worth removing the `borderWidth` from `styles.circle` and letting this be the source of truth for setting that value i.e. changing `null` to `1`. Just a thought to potentially make this a little easier to follow

src/frontend/sharedComponents/icons/Circle.js
achou11: Additionally, may be worth validating that `color` is valid using `validate-color` (which is already used elsewhere in the project) and set the border color based on that e.g.

```js
borderColor: color && validateColor(color) ? color : `#EAEAEA`
```

achou11 (commented):
thanks for this! left a couple of comments worth looking at although not hard-blocking

rudokemper:
TODO: add colored borders to some of the other screens where the icons are used, like `ObservationEdit`, `ObservationsList`, `ObservationDetails`, `Observation`, ...

luandro:
Thanks for the great review and code tips @achou11! Did some refactoring, lemme know what you think.

Thank you for scoping out the missing components @rudokemper. They should all have borders now:

src/frontend/sharedComponents/icons/Circle.js
luandro: Don't understand what you mean here.

achou11: sorry. to explain, initially l didn't realize that the `styles.circle` sets `borderWidth: 1`, so it seemed like your change was setting the width to `null`, when really it would fallback to using `1` since styles.circle is used. My suggestion was to avoid setting that in two different locations and instead only specifying it in this inline object. may be irrelevant after recent commits

src/frontend/sharedComponents/icons/Circle.js
```diff
@@ -2,28 +2,42 @@
...
+  children,
+}: CircleProps) => {
+  const validColor = color && validateColor(color);
+  return (
+    <View
+      style={[
+        styles.circle,
+        {
+          borderWidth: validColor && borderWidth,
+          borderColor: validColor && color,
```
achou11: non-blocking but technically these properties don't accept a potential boolean - guessing that React Native handles this okay but if this component were written in TS, the linter may complain about this 😄 if it rendered okay when testing it out, wouldn't worry about fixing this right now - can be done whenever we migrate this component to TS

achou11 (approved):
Non-blocking comment but lgtm! Make sure you squash-and-merge when ready

jencastrodoesstuff:
[nature category icons.zip](https://github.com/digidem/mapeo-mobile/files/9531809/nature.category.icons.zip) I tried to build a set of configs (defailt v4 branch beta 5) with scaled to fit artboared nature SVG's  but ran into build issues and got frustrated. I was hoping to see a them in a build so I could determine if the artboard was determining icon size withing the Mm icon circles compared to the threats icons
[threats icon not rescaled.zip](https://github.com/digidem/mapeo-mobile/files/9531850/threats.icon.not.rescaled.zip)



[nature category icons.zip](https://github.com/digidem/mapeo-mobile/files/9531826/nature.category.icons.zip)

jencastrodoesstuff:
Was able to fix my config build issue to create a beta of future defaults to test.  The config with resized icons is [on Gdrive](https://drive.google.com/file/d/1Hkv0mbFeyzLSglc-qaX_XS5A-EZKcbAe/view?usp=sharing) and can be played with on are on the [jen branch](https://github.com/digidem/mapeo-default-config/tree/Jen-v.4.0.0)

src/frontend/sharedComponents/icons/Circle.js
achou11: update on this: this turns out to be a blocking issue. Suggestion to remove the `borderColor` and `borderWidth` values in the `style` object and update this section to be:

```suggestion
          borderWidth: validColor ? 1.5 : 1,
          borderColor: validColor ? color : "#EAEAEA",
```

src/frontend/sharedComponents/icons/Circle.js
```diff
@@ -2,28 +2,42 @@
...
-      },
-      style,
-    ]}
-  >
-    {children}
-  </View>
-);
+const Circle = ({
+  color,
+  borderWidth = 2,
```
achou11: Remove this prop since it's somewhat redundant with the `style` prop and because if a valid color is passed, safer to be opinionated and use a specific value defined within this component for the border width

achou11 (dismissed):
Discussed this change in the UX meeting and there are a couple of blocking issues with the code changes. I can update the PR today and play around with potential edge cases discussed during the meeting

src/frontend/sharedComponents/icons/Circle.js
achou11: > safer to be opinionated and use a specific value defined within this component for the border width

ah this part doesn't hold for this use case because it's possible for a category to not have a color defined while the others do, and in that case we want a consistent border width.

src/frontend/sharedComponents/icons/Circle.js
achou11: The suggested `borderWidth` adjustment needs to be amended because of https://github.com/digidem/mapeo-mobile/pull/1004#discussion_r976849914. My mistake for overlooking that case

macandrow:
@luandro Hello! We talked about this during the UX weekly and it looks good. Could we go with the 1.5 border for width? And I believe Andrew will add/ has added notes about the default color for observations with no chosen color.

luandro:
Hey @achou11, thank you and the tech team for diving into this. I won't have time to work on this until mid next week, so if there's a hurry to have these changes made, feel free to ajust. I'll also try to get some final feedback from the program's team today at our meeting.

achou11:
@luandro no worries! started making adjustments locally yesterday. haven't finished just yet but will push later today

src/frontend/sharedComponents/icons/CategoryIcon.tsx
```diff
@@ -45,13 +47,25 @@ export const CategoryIcon = React.memo<IconProps>(
...
+
   return (
-    <Circle radius={radii[size]} style={style}>
-      <CategoryIcon {...props} size={size} />
+    <Circle
+      color={color}
+      radius={radii[size]}
+      style={presetsUseColors ? { borderWidth: 1.5 } : undefined}
+    >
+      <CategoryIcon iconId={iconId} size={size} />
```
achou11: Since the variable border width is only applicable to the circle icon when used for categories, figured it would make more sense to do that calculation here and provide as a style override to the default circle border width if needed, removing any of that logic in the Circle component itself

src/frontend/screens/ObservationsList/ObservationListItem.tsx
```diff
@@ -84,14 +85,16 @@ const ObservationListItem = ({
...
-              size="small"
-              style={styles.smallIcon}
-            />
+            <View style={styles.smallIconContainer}>
+              <CategoryCircleIcon
+                iconId={iconId}
+                color={iconColor}
+                size="small"
+              />
+            </View>
```
achou11, line 94: More of a refactor change, but this was the only place that used the `style` override prop for `CategoryCircleIcon` (note, _not_ `Circle`) and it only adjusts the positioning of the icon, so decided to wrap it in order to remove that prop altogether (see https://github.com/digidem/mapeo-mobile/pull/1004/commits/d9a065046b0cc11efc098ad0cde3a1f8d5c4252f)

src/frontend/sharedComponents/icons/CategoryIcon.tsx
achou11: @gmaclennan is the `presetsUseColors` calculation here the proper way of determining if the preset has a color defined that's applicable to categories?

achou11:
@luandro see recent commits for updates to the implementation. Screenshots for `mapeo-default-settings v3.6.1` below:

- Category Chooser:

  category-chooser

- Observations list:

  observations-list

- Observation details:

  observation-details

- Edit observation:

  edit-observation

jencastrodoesstuff:
Offering a quick peek at how the latest shows up on my Google pixel 2 with the cleaned up SVG's (which go to the edge of a vector circle with no padding 
Current defaults
photo_2022-09-29 17 14 36
photo_2022-09-29 17 14 27

 Future defauilts (using a seperate branch  on beta 6 for icon testing)
Note that boundary, airstrip, and community need had the color definition removed to compare appearence.


photo_2022-09-29 17 14 42
photo_2022-09-29 17 14 44
photo_2022-09-29 17 14 46

PR #1007, opened 2022-09-08
Update map server integration

achou11:
Notes
- updates `@mapeo/map-server` to latest version (alpha 10)
- updates the backend build script to account for the `sdf` directory introduced by the map server, as a well as excluding `perf_hooks` which is introduced by map server via `piscina`
- adds a patch to `hdr-histogram-js`, which is introduced by map server via `piscina`. original code was using an `Intl` api, which is not supported by the js runtime we use, thus causing the app to crash on startup. I don't think we use any part of piscina that relies on this dep, so potentially an unecessary patch to that dep instead of piscina, but this works too 😄 
  - fun to see the potential [perf improvement](https://perf.link#eyJ0aXRsZSI6IkZvcm1hdHRpbmcgYSBmbG9hdGluZyBwb2ludCBudW1iZXIgdG8gbiBwbGFjZXMiLCJiZWZvcmUiOiJsZXQgZGF0YSA9IG5ldyBBcnJheSgxMDApLmZpbGwobnVsbCkubWFwKE1hdGgucmFuZG9tKVxuXG5sZXQgZGlnaXRzID0gM1xuXG5sZXQgZm9ybWF0dGVyID0gbmV3IEludGwuTnVtYmVyRm9ybWF0KFwiZW4tVVNcIiwge1xuICAgIG1heGltdW1GcmFjdGlvbkRpZ2l0czogZGlnaXRzLFxuICAgIG1pbmltdW1GcmFjdGlvbkRpZ2l0czogZGlnaXRzLFxuICAgIHVzZUdyb3VwaW5nOiBmYWxzZSxcbiAgfSk7XG5cbiAgXG5sZXQgZm9ybWF0SW50bCA9IChuKSA9PiBmb3JtYXR0ZXIuZm9ybWF0KG4pXG5sZXQgZm9ybWF0VG9GaXhlZCA9IChuKSA9PiBuLnRvRml4ZWQoZGlnaXRzKSIsInRlc3RzIjpbeyJuYW1lIjoiVXNpbmcgSW50bC5OdW1iZXJGb3JtYXQiLCJjb2RlIjoiZGF0YS5mb3JFYWNoKGZvcm1hdEludGwpIiwicnVucyI6WzI1MDAsMTAwMCw0MDAwLDUwMDAsODAwMCw1MDAwLDUwMCw1MDAsODUwMCwzMDAwLDcwMDAsNTUwMCw2MDAwLDQwMDAsNDAwMCw3NTAwLDU1MDAsMTAwMCwzNTAwLDc1MDAsNTAwMCw5MDAwLDU1MDAsMTAwMCwxNTAwLDEwMDAsNDAwMCw0NTAwLDY1MDAsMTUwMCw2NTAwLDI1MDAsMzUwMCw3NTAwLDg1MDAsNzAwMCw5NTAwLDIwMDAsOTAwMCwzMDAwLDY1MDAsNTAwLDU1MDAsNjUwMCwzNTAwLDI1MDAsMzUwMCw3NTAwLDUwMDAsNDUwMCw3MDAwLDc1MDAsNjAwMCw1MDAsMzUwMCw1MDAwLDgwMDAsNDUwMCwxNTUwMCwzMDAwLDUwMDAsMzUwMCwxNDUwMCw1MDAsOTUwMCwzNTAwLDE1MDAsOTUwMCw1NTAwLDk1MDAsNzUwMCwzNTAwLDY1MDAsMzUwMCw0MDAwLDgwMDAsNTAwMCw1MDAsNTUwMCw5MDAwLDIwMDAsOTAwMCw0NTAwLDgwMDAsNDUwMCwxMDAwLDI1MDAsNjUwMCw0MDAwLDc1MDAsNTAwMCw1NTAwLDcwMDAsMjUwMCw1MDAsNDUwMCw1MDAwLDQwMDAsNTAwMCwyNTAwXSwib3BzIjo1MDA1fSx7Im5hbWUiOiJVc2luZyB0b0ZpeGVkIiwiY29kZSI6ImRhdGEuZm9yRWFjaChmb3JtYXRUb0ZpeGVkKSIsInJ1bnMiOls5MzAwMCwzMjUwMCw2MTAwMCw3MDUwMCw5MjUwMCw2MjUwMCwxNTAwMCw1MDAwLDkwMDAwLDUxMDAwLDc5NTAwLDc1NTAwLDc4MDAwLDQ5NTAwLDY2MDAwLDkwMDAwLDc2NTAwLDMyNTAwLDQ3NTAwLDkyMDAwLDc1NTAwLDEwNTAwMCw3MzUwMCwzMjUwMCw0MDUwMCwzMjUwMCw2MTAwMCw2OTAwMCw3NDUwMCwzMjUwMCw4MzUwMCw0NjUwMCw0OTUwMCw4NjAwMCw5MDUwMCw4MzUwMCwxMDk1MDAsMzUwMDAsMTA0NTAwLDQ4MDAwLDkyNTAwLDI5MDAwLDM2NTAwLDg1MDAwLDQ2MDAwLDQyNTAwLDQ5NTAwLDg5MDAwLDY2NTAwLDYyNTAwLDgzNTAwLDEwNDUwMCw2OTAwMCwyNDUwMCw1MzUwMCw3MDUwMCw5NDAwMCw2MDAwMCwxNTIwMDAsNDUwMDAsNjkwMDAsNDcwMDAsMTQ1NTAwLDIwMDAwLDExMTUwMCw0OTUwMCwzNDAwMCwxMDcwMDAsNzI1MDAsMTAxNTAwLDg1MDAwLDU1NTAwLDgwNTAwLDQ2MDAwLDUzNTAwLDg5NTAwLDYwNTAwLDE2MDAwLDU5NTAwLDEwNzUwMCwzOTUwMCwxMDM1MDAsNjU1MDAsOTcwMDAsNjQ1MDAsMzE1MDAsMzc1MDAsOTUwMDAsNTEwMDAsOTE1MDAsNjM1MDAsNzUwMDAsODcwMDAsNDM1MDAsMTYwMDAsNTc1MDAsNzEwMDAsNTI1MDAsNjY1MDAsNDE1MDBdLCJvcHMiOjY2MTAwfV0sInVwZGF0ZWQiOiIyMDIyLTA5LTA4VDIxOjE3OjE5Ljk5OVoifQ%3D%3D) 
- in similar fashion to what we do with the worker file that the map server uses, we had to process a worker file that piscina uses using similar methods. not ideal and incredibly fragile, but works for now.
  - this requires a [patch to noderify](https://github.com/digidem/mapeo-mobile/pull/1007/commits/fdebace88bfa4ca070a9836a36c71747f08db301) to fix bundled files that are eventually used as worker entries (thanks Gregor!)
- implemented a [hacky](https://github.com/digidem/mapeo-mobile/pull/1007/commits/14a2e6cb730f88fc5d6265b7a51117a9c24f23eb) solution to a db migration problem that i introduced. context is that since the map server is started on initial app load, we created a map server db for each device that upgraded to the app version that has this initial integration. unfortunately, i made schema changes to the map server db but didn't generate proper migrations for them (instead i just replaced the "initial" migration completely). this presents an issue for users with the first version of the db already created on their device, as updating the app to include the new work in this pr would cause the map server initialization to crash (basically incompatible schemas). 
  - i detailed the context and consequences as a comment, but basically i'm justifying this hack based on the fact that the map server is still incredibly experimental and hasn't been officially released for public use. only mentioning this because the implementation causes users with the old map server db to lose the data in it and no migration from that to the new one.
- Another concern worth noting is that the inclusion of an offline font (`sdf/` directory) means that the app size is significantly larger. the dev version of the app is a couple of hundred mb in size, which is quite large. not sure if that poses as a problem or not, but wanted to point that out.


ErikSin:
What do you think about merging this (just the updated dependency and patch) and having the what is left to do as seperate Prs? In the spirit of smaller Prs

achou11:
> What do you think about merging this (just the updated dependency and patch) and having the what is left to do as seperate Prs? In the spirit of smaller Prs

happy to work on the front end-specific changes that are feature additions in a follow-up. there may be some small frontend changes that are included here though because otherwise the current implementation will break.

the last item I mention about the db should be handled in this PR, so still some work to do here before merging

src/backend/server.js
```diff
@@ -179,9 +179,33 @@ function createServer({
...
+      // confidently remove this code if we feel that enough existing users have installed this version.
+      try {
+        const legacyMapsDbPath = path.join(privateStorage, "maps.db");
+        fs.unlinkSync(legacyMapsDbPath);
+        fs.unlinkSync(legacyMapsDbPath + "-shm");
+        fs.unlinkSync(legacyMapsDbPath + "-wal");
+        log("Successfully removed legacy map server db");
+      } catch {
+        log("Could not find legacy map server db to remove");
+      }
```
achou11, line 204: would love to know if anyone has alternate ideas or if this is the best solution for now

scripts/build-backend.sh
```diff
@@ -46,28 +46,38 @@ echo -en "Minifying with noderify..."
   --filter=async_hooks \
   --filter=utf-8-validate \
   --filter=bufferutil \
-  --filter=worker_threads \
```
achou11, line 49: `noderify@5.1.1` filters this by default now: https://github.com/staltz/noderify/commit/ac8b095c56c045880fde5933ac1d02f88b20ef6a

scripts/build-backend.sh
```diff
@@ -46,28 +46,38 @@ echo -en "Minifying with noderify..."
...
   # not at the project root. If the path name of the worker changes in the
   # original module then this replacement needs to be updated. Apologies for the
   # fragility, but this was the best solution I could find in limited time.
-  sed 's|\./lib/mbtiles_import_worker\.js|../../../../mbtiles_import_worker.js|' > \
+  sed 's|\.\./lib/mbtiles_import_worker\.js|../../../../../mbtiles_import_worker.js|' | \
+  # Additional apologies for this, which is even more fragile. Context is that 
+  # Piscina introduces its own worker file (unfortunately with a very generic name of `worker.js`)
+  # so we need to update the path it references to point to the eventually relocated
+  # file within the nodejs-project directory.
+  sed "s|'worker\.js'|'../../../../piscina_worker.js'|" > \
```
achou11, line 65: somehow managed an even more fragile addition to this file compared to the import worker above 😄

scripts/build-backend.sh
```diff
@@ -46,28 +46,38 @@ echo -en "Minifying with noderify..."
...
-  --filter=worker_threads \
   node_modules/@mapeo/map-server/dist/lib/mbtiles_import_worker.js > \
-  ../nodejs-project/mbtiles_import_worker.js
+  ../nodejs-project/mbtiles_import_worker.js \
+
+# Piscina uses its own worker to initiate other workers, 
+# so we need to bundle it seperately
+"$(npm bin)/noderify" \
+  node_modules/piscina/dist/src/worker.js > \
+  ../nodejs-project/piscina_worker.js
```
achou11, line 80: not entirely sure if there are any bindings i need to replace here - how would i check? ran fine otherwise when testing some endpoints

src/backend/patches/hdr-histogram-js+2.0.3.patch
```diff
@@ -0,0 +1,19 @@
...
++++ b/node_modules/hdr-histogram-js/dist/formatters.js
+@@ -25,12 +25,7 @@ exports.keepSignificantDigits = (digits) => (value) => {
+     return value;
+ };
+ exports.floatFormatter = (size, fractionDigits) => {
+-    const numberFormatter = new Intl.NumberFormat("en-US", {
+-        maximumFractionDigits: fractionDigits,
+-        minimumFractionDigits: fractionDigits,
+-        useGrouping: false,
+-    });
```
achou11, line 13: for context, the v8 engine we use doesn't include the `Intl` apis, so this fails on app startup

scripts/build-backend.sh
gmaclennan: You don't need any replacements. You can check by looking at the source code https://github.com/piscinajs/piscina/tree/current/src and seeing what modules are required.

src/backend/patches/hdr-histogram-js+2.0.3.patch
gmaclennan: That's frustrating! Another solution would have been to define a polyfill somewhere like

```js
global.Intl = { NumberFormat: () => (num) => num + '' }
```

But I think your solution is probably the best in this case.

scripts/build-backend.sh
gmaclennan: yeah... it will do for now. It's not crazy to do this kind of thing with regex (see https://github.com/danielgindi/rollup-plugin-natives/blob/master/src/index.js) but it is fragile, and it's more robust to do this with an AST parser e.g. acorn - see https://github.com/playcode/rollup-plugin-node-globals/blob/master/src/inject/index.js

At some point we can write some rollup plugins to do an AST parse and extract any workers, but this will do for now.

src/backend/server.js
gmaclennan: This is because we're not bothering to migrate the previous one?

achou11: yeah but it's mostly because of a limitation with how the map server module executes migrations at runtime. i basically completely replaced the "initial" migration with something that uses a slightly different schema than the first integration's db. so when trying to use the same db and running that initial migration, a runtime error will occur because of the differing schema.

What i should've done in the map server module was create a new migration that would then be applied on top of the original initial migration at runtime so that there would be a seamless upgrade, but i didn't think far ahead enough when doing that work and considered the map server experimental enough where replacing the initial migration would be okay

achou11: this is mostly an issue because the old db has been created on devices that have installed the most recent published version of mapeo. if that weren't the case, this wouldn't be as much of a concern right now 😅

achou11: and i don't want to just leave the old db on these devices so that it coexists with the new db, hence why this code to remove it exists

achou11:
This is basically ready for merging but to make it easier to highlight concerns:

1. The "legacy" db solution I implemented. Not really a problem in the short-term but something worth reflecting on (maybe I just have to be more careful about map server releases moving forward 😄 )

2. I imagine the offline SDFs will add a decent amount to the APK size (hard to tell in a dev build but definitely a non-zero amount). Any concerns around that?

PR #1012, opened 2022-09-14
Kill passcode screens

ErikSin:
# Screens for Kill Passcode

This introduces the screen for Kill Passcode, telling the user what the kill passcode is, and allows them to turn on the kill passcode.

This does not have any functionality, and is simply the screen.

image
image


src/frontend/screens/KillPasscode.tsx
```diff
@@ -0,0 +1,108 @@
...
+});
+
+export const KillPasscode: NativeNavigationComponent<"KillPasscode"> = () => {
+  const [killPasscodeEnabled, setKillPasscodeEnabled] = React.useState(false);
+
+  return (
+    <ScrollView style={styles.container}>
+      <Text style={styles.title}>
+        <FormattedMessage {...m.title} />
+      </Text>
```
gmaclennan: Do we want duplicate titles - e.g. in the header and in the main view?

package.json
```diff
@@ -11,6 +11,7 @@
     "start-otf": "FEATURE_ONBOARDING=true FEATURE_PASSCODE=true npm start -- --reset-cache",
     "start-experiments": "FEATURE_DEV_EXPERIMENTS=true npm start -- --reset-cache",
     "start-icca": "RN_SRC_EXT=icca react-native start",
+    "start-passcode": "FEATURE_PASSCODE=true npm start -- --reset-cache",
```
gmaclennan, line 14: I don't see where this FEATURE_PASSCODE is read / used, or is that coming in a later PR?

gmaclennan (commented):
Looks ok to me. Just UI question about the duplicate title, and dev question about the feature flag.

gmaclennan:
Also can we change the language here? I know we have an open question about copy for the app, but I think there is universal agreement that "kill passcode" is a bad name so I don't think we should use it in the code to name things either. Perhaps "Obscurity Passcode"?

ErikSin:
> Also can we change the language here? I know we have an open question about copy for the app, but I think there is universal agreement that "kill passcode" is a bad name so I don't think we should use it in the code to name things either. Perhaps "Obscurity Passcode"?

Yes, that makes sense. I am going to keep it as `kill passcode` in this PR, and then change it in the next PR that is stacked on this. Changing file names on stacked PR just introduces both files, so it will be much easier to do on the pr that is stacked on top

package.json
ErikSin: https://github.com/digidem/mapeo-mobile/blob/33df94085dff0537c339cdca67cb8b02aa7a2930/src/frontend/screens/Settings/Settings.tsx#L177

src/frontend/screens/KillPasscode.tsx
ErikSin: This was in the design that Sabella Built, but It made more sense when we had the paper key title as well. Ill take it out for now

ErikSin: This was in Sabella's Design...Im not sure why this was the case. Ill ask Sabella

PR #1013, opened 2022-09-21
App passcode state management

ErikSin:
# State Managment
This Pr introduces:
1. `SecurityContext` which deals with state management of app passcode and kill passcode
2. Turn off App Passcode Screen, which allows user to turn off/or change their app passcode

This branch is based on #1012 as a stacked PR

## Features of Security Context
- When App Passcode is set, App passcode is stored in async storage
- When Kill Passcsode is set, Kill Passcode is store in async storage
- On initialization of the app, async storage is checked in order set the values to be available to the whole app

### What this PR does NOT do
- Passcode and kill passcode can be set, but there is still no step to authenticate. In other words, the authentication screens and corresponding navigation are not built yet.

### To Do
- [x] Merge #1012 
- [ ] Rebase from develop once 1012 is merged

image


gmaclennan (commented):
I think this needs a more detailed review than I have time for this evening, but a couple of initial thoughts on the context part, which I was able to review:

1. I think whether an app passcode is enabled/required (e.g. true/false) and whether an obscurity passcode is enabled should be stored in the settings context, since these are options on the settings screen. It does make sense to store the actual passcodes in a different context.
2. A pattern we have used so far for context, when loading from initial values from async storage, is to simply hold off rendering children until the values are loaded. That's has so far felt like an acceptable delay in rendering, rather than having to build all the child components with a "loading" state.
3. We have the `usePersistedState` abstraction, which avoids a lot of the boilerplate code here dealing with async storage.
4. See if you can have a single "state" for auth. Right now there are two states" AppMode and AuthStatus. The combinations of these might be difficult to debug. To me it seems like states are "unauthenticated", "authenticated", "obscure". With possibly "loading", but re. (2) maybe we can avoid that.

ErikSin:
>     1. I think whether an app passcode is enabled/required (e.g. true/false) and whether an obscurity passcode is enabled should be stored in the settings context, since these are options on the settings screen. It does make sense to store the actual passcodes in a different context.
> 
>     2. A pattern we have used so far for context, when loading from initial values from async storage, is to simply hold off rendering children until the values are loaded. That's has so far felt like an acceptable delay in rendering, rather than having to build all the child components with a "loading" state.
> 
>     3. We have the `usePersistedState` abstraction, which avoids a lot of the boilerplate code here dealing with async storage.
> 
>     4. See if you can have a single "state" for auth. Right now there are two states" AppMode and AuthStatus. The combinations of these might be difficult to debug. To me it seems like states are "unauthenticated", "authenticated", "obscure". With possibly "loading", but re. (2) maybe we can avoid that.

1. I moved obscurePasscode to setting. AppPasscode state is derived from whether app passcode is set to null or not. We could explicitly create this state, and use a `useEffect` to attach this state to the settings context, but feels like an unnecessary step. 
2. switched to this loading architecture
3. Moved away from async storage into usePersistedState
4. I created 4 state: "authenticated", "unautheticated", "notRequired", "obscureMode", and took away the app mode state.

gmaclennan (commented):
Saw the changes. Some more questions, thoughts and suggestions about how to make this more maintainable / easier to reason with:

1. What is the difference between Auth states `authenticated` and `notRequired`? Maybe consider merging them? The initial auth state can be derived based on if a passcode is set (e.g. passcode === null -> state = 'authenticated')
2. I don't think the SettingsContext should expose the passcode. Logic for authenticating should stay within the context, e.g. exposing an `authenticate(passcode)` method. We should probably implement hashing the passcode before general release, but not necessary for this PR.
3. Do we want to support custom obscurity passcodes in the future? If so then maybe keep that state in the context the same as the passcode state.
4. Context should probably then expose two booleans on whether an obscurity passcode or regular passcode is set.

We can then render the screens based on authState with this pattern: https://reactnavigation.org/docs/auth-flow (no manual navigation when entering a passcode to access the app).

Possible pattern to avoid needing an auth loading state (or could just accept having a loading state):

```js
const SecurityProviderOuter = ({ children }) => {
  const [passcode, status, setPasscode] = usePersistedPasscodeState()
  if (status === 'loading') return null
  return <SecurityProviderInner {{ passcode, setPasscode }}>{children}</SecurityProviderInner>
}

const SecurityProviderInner = ({ children, passcode, setPasscode }) => {
  // Now we know that passcode is either `null` or `string`
  // Do the other logic to set the contextValue
}
```

ErikSin:
> Saw the changes. Some more questions, thoughts and suggestions about how to make this more maintainable / easier to reason with:
> 
>     1. What is the difference between Auth states `authenticated` and `notRequired`? Maybe consider merging them? The initial auth state can be derived based on if a passcode is set (e.g. passcode === null -> state = 'authenticated')
> 
>     2. I don't think the SettingsContext should expose the passcode. Logic for authenticating should stay within the context, e.g. exposing an `authenticate(passcode)` method. We should probably implement hashing the passcode before general release, but not necessary for this PR.
> 
>     3. Do we want to support custom obscurity passcodes in the future? If so then maybe keep that state in the context the same as the passcode state.
> 
>     4. Context should probably then expose two booleans on whether an obscurity passcode or regular passcode is set.
> 

I took out `notRequired` and created useEffect where `if passcode===null => authState is set to `authenticated`

Also changed the architecture so that authentication only happens in the security context (aka passcode is not exposed to the children).

There are a couple things to note. passcode, and obscure passcode are set in security context. I also created a state in the security context, using the following type, to track whether pass/obscure code is being used or not (since we are not exposing the pass/obscure code) :
```ts
type AuthValuesSet = {
  passcodeSet: boolean;
  obscureSet: boolean;
};
```
this state is derived from the pass/obscure code set. In other words if `pass/obscure code ===null` then `pass/obscure codeSet =false`. Since the pass/obscure code is set in this context, and we are not exposing it to settings, i cannot set `AuthValuesSet`in the settings context, unless I passed the pass/obscure code to the settings context....which would defeat the whole purpose of these boolean values.

I also set it up that custom obscure codes can be used in the future, but right now it defaults to the one set.

> 
> We can then render the screens based on authState with this pattern: https://reactnavigation.org/docs/auth-flow (no manual navigation when entering a passcode to access the app).
> 

Yes, this is used in my next PR. right now, I am just setting the passcode and holding it state (there is no actual authentication happening) => in the spirit of smaller PRs.

> Possible pattern to avoid needing an auth loading state (or could just accept having a loading state):
> 
> ```js
> const SecurityProviderOuter = ({ children }) => {
>   const [passcode, status, setPasscode] = usePersistedPasscodeState()
>   if (status === 'loading') return null
>   return <SecurityProviderInner {{ passcode, setPasscode }}>{children}</SecurityProviderInner>
> }
> 
> const SecurityProviderInner = ({ children, passcode, setPasscode }) => {
>   // Now we know that passcode is either `null` or `string`
>   // Do the other logic to set the contextValue
> }
> ```

I think i don't fully understand why this architecture doesn't work(without having a parent): 

``` ts
const [passcode, status, setPasscode] = usePersistedPasscodeState()
<SecurityContext.Provider value={contextValue}>
      {status === "loading" ? null : children}
 </SecurityContext.Provider>
```

Can you elaborate on why this wouldn't work?

src/frontend/context/AppProvider.js
```diff
@@ -1,29 +1,28 @@
...
 
 // This is a convenience wrapper for providing all App contexts to the tree,
 // apart from the Permissions Provider which is needed separately.
 const AppProvider = ({ children }: { children: React.Node }) => (
   <LocationProvider>
-    <ObservationsProvider>
-      <ConfigProvider>
-        <SettingsProvider>
-          <SecurityProvider>
+    <SecurityProvider>
```
ErikSin, line 15: I moved the `SecurityProvider` outside the `ObservationProvider` because in obscure mode, no observations will be seen, so the `observationProvider` needs to know if obscure mode is on

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,155 @@
...
+        throw new Error("passcode not valid");
+      }
+
+      setPasscode(passcodeValue);
+    },
+    [OBSCURE_PASSCODE]
+  );
+
+  const setObscureCodeWithValidation = React.useCallback(
+    (newObscureVal: string | null | undefined) => {
```
ErikSin, line 98: when using this function, currently nowhere in the app passes `newObscureValue` because when the value is undefined it sets it to the default. There is no UI to set the obscureCode, but this architecture will make it easy to set obscureCode if we choose to do that.

src/frontend/hooks/useAuthState.ts
```diff
@@ -0,0 +1,4 @@
+import * as React from "react";
```
ErikSin, line 1: Eventually the navigation will just need access to auth state, so abstraction will simplify that

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,155 @@
...
+  React.useEffect(() => {
+    if (passcode === null) {
+      setobscureCode(null);
+      setAuthState("authenticated");
+    }
+  }, [passcode]);
+
+  const setPasscodeWithValidation = React.useCallback(
+    (passcodeValue: string | null) => {
+      if (passcodeValue === OBSCURE_PASSCODE) {
```
gmaclennan, line 75:
```suggestion
      if (passcodeValue === obscureCode) {
```
Future-proof in case we forget later

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,155 @@
...
+        throw new Error("passcode is reserved");
+      }
+
+      if (!validPasscode(passcodeValue)) {
+        throw new Error("passcode not valid");
+      }
+
+      setPasscode(passcodeValue);
+    },
+    [OBSCURE_PASSCODE]
```
gmaclennan:
```suggestion
    [obscureCore]
```

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,155 @@
...
+        return;
+      }
+
+      if (!validPasscode(newObscureVal) || newObscureVal === passcode) {
+        throw new Error("passcode not valid");
+      }
+
+      setobscureCode(newObscureVal);
+    },
+    [passcode, OBSCURE_PASSCODE]
```
gmaclennan: `OBSCURE_PASSCODE` not needed as a dependency since it doesn't change.

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+      throw new Error("Incorrect Passcode");
+    },
+    [passcode, obscureCode]
+  );
+
+  const setAuthValues = React.useCallback(
+    ({ type, value }: AuthSetters) => {
+      if (type === "passcode") setPasscodeWithValidation(value);
+      setObscureCodeWithValidation(value);
```
gmaclennan:
```suggestion
      else setObscureCodeWithValidation(value);
```

src/frontend/screens/AppPasscode/InputPasscode.tsx
```diff
@@ -67,28 +70,40 @@ interface SetPasscodeProps {
...
+    if (screenState === "confirmSetPasscode" && !initialPassword.current) {
+      setScreenState("intro");
+    }
+  }, [screenState]);
 
   React.useEffect(() => {
-    if (error && inputtedPass.length > 0) setError(false);
-  }, [inputtedPass, error]);
+    if (inputtedPass.length > 0) setError(false);
+  }, [inputtedPass]);
```
gmaclennan: You should avoid useEffect for this. Instead do this in a handler for `onChangeTextWithValidation`.

src/frontend/screens/AppPasscode/InputPasscode.tsx
```diff
@@ -67,28 +70,40 @@ interface SetPasscodeProps {
   setScreenState: React.Dispatch<React.SetStateAction<PasscodeScreens>>;
 }
 
-/** This screen is used when the user is setting a password, confiriming their password, and will be used when the user is entering their password to enter the app */
+/** This screen is used when the user is setting a password and confiriming their password */
```
gmaclennan: I think you're maybe trying to do too much with this screen? I don't fully understand what is happening here, but it seems like you are using refs to enable this screen to be used in different places? Perhaps the UI part (without state) can be the part that you re-used, and we just write a different screen  / state logic for each place we use it? I feel like the code seems like it is doing a lot to avoid having 2 or 3 separate components.

gmaclennan (commented):
Sorry I'm taking so long with these reviews. I feel like this state management is an important part of this feature, and it's important to get it right so that the code is maintainable in the future and does not become a source of bugs. I hope the comments are helpful.

gmaclennan:
> > Saw the changes. Some more questions, thoughts and suggestions about how to make this more maintainable / easier to reason with:
> > ```
> > 1. What is the difference between Auth states `authenticated` and `notRequired`? Maybe consider merging them? The initial auth state can be derived based on if a passcode is set (e.g. passcode === null -> state = 'authenticated')
> > 
> > 2. I don't think the SettingsContext should expose the passcode. Logic for authenticating should stay within the context, e.g. exposing an `authenticate(passcode)` method. We should probably implement hashing the passcode before general release, but not necessary for this PR.
> > 
> > 3. Do we want to support custom obscurity passcodes in the future? If so then maybe keep that state in the context the same as the passcode state.
> > 
> > 4. Context should probably then expose two booleans on whether an obscurity passcode or regular passcode is set.
> > ```
> 
> I took out `notRequired` and created useEffect where `if passcode===null => authState is set to `authenticated`
> 
> Also changed the architecture so that authentication only happens in the security context (aka passcode is not exposed to the children).
> 
> There are a couple things to note. passcode, and obscure passcode are set in security context. I also created a state in the security context, using the following type, to track whether pass/obscure code is being used or not (since we are not exposing the pass/obscure code) :
> 
> ```ts
> type AuthValuesSet = {
>   passcodeSet: boolean;
>   obscureSet: boolean;
> };
> ```
> 
> this state is derived from the pass/obscure code set. In other words if `pass/obscure code ===null` then `pass/obscure codeSet =false`. Since the pass/obscure code is set in this context, and we are not exposing it to settings, i cannot set `AuthValuesSet`in the settings context, unless I passed the pass/obscure code to the settings context....which would defeat the whole purpose of these boolean values.
> 
> I also set it up that custom obscure codes can be used in the future, but right now it defaults to the one set.
> 
> > We can then render the screens based on authState with this pattern: https://reactnavigation.org/docs/auth-flow (no manual navigation when entering a passcode to access the app).
> 
> Yes, this is used in my next PR. right now, I am just setting the passcode and holding it state (there is no actual authentication happening) => in the spirit of smaller PRs.
> 
> > Possible pattern to avoid needing an auth loading state (or could just accept having a loading state):
> > ```js
> > const SecurityProviderOuter = ({ children }) => {
> >   const [passcode, status, setPasscode] = usePersistedPasscodeState()
> >   if (status === 'loading') return null
> >   return <SecurityProviderInner {{ passcode, setPasscode }}>{children}</SecurityProviderInner>
> > }
> > 
> > const SecurityProviderInner = ({ children, passcode, setPasscode }) => {
> >   // Now we know that passcode is either `null` or `string`
> >   // Do the other logic to set the contextValue
> > }
> > ```
> 
> I think i don't fully understand why this architecture doesn't work(without having a parent):
> 
> ```ts
> const [passcode, status, setPasscode] = usePersistedPasscodeState()
> <SecurityContext.Provider value={contextValue}>
>       {status === "loading" ? null : children}
>  </SecurityContext.Provider>
> ```
> 
> Can you elaborate on why this wouldn't work?

The challenge will be a "flashing passcode screen". `status` will be `loaded`, and children will render, with `authState` as `unauthenticated`, but then the `useEffect` will fire, and since passcode = null, then `authState` will become `authenticated`. The result will be the passcode code entry screen will render once, then be removed on the next render.

The way around this is to either implement a "loading" `authState`, or do what I suggested which avoids rendering any children until everything is loaded. My code is basically a work around from needing to do an initial render then a useEffect to change the state, because the inner context is first rendered when we already know if `passcode` is `null` or not, so the `authState` can start with an initial value of  `authenticated` if needed, rather than needing initialize as `unauthenticated` and then `useEffect` to change it.

gmaclennan:
(all this could be solved of course if we switch to [mmkv](https://github.com/mrousavy/react-native-mmkv) for storage, which has synchronous methods so we wouldn't have any loading state for persisted values)

src/frontend/context/SecurityContext.tsx
ErikSin: Obscure code will ALWAYS be set after the passcode. This is only a check that for when we have the default one. I will add a to do comment here

src/frontend/screens/AppPasscode/InputPasscode.tsx
ErikSin: I can give that a try. The pages are identical, with different text. But I agree that it is a little confusing given that even though the ui is the same, the logic for each page is slightly different.

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,172 @@
...
+  children,
+  passcode,
+  setPasscode,
+}: {
+  children: React.ReactNode;
+  passcode: string | null;
+  setPasscode: React.Dispatch<React.SetStateAction<null | string>>;
+}) => {
+  const [authState, setAuthState] = React.useState<AuthState>(
+    "unauthenticated"
```
gmaclennan:
```suggestion
    passcode === null ? "authenticated" : "unauthenticated"
```

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,172 @@
...
+  >(null);
+
+  // if passcode is unset, we want to makes sure that obscure code is unset, and that the user is 'authenticated' aka does not need a passcode to enter the app.
+  React.useEffect(() => {
+    if (passcode === null) {
+      setobscureCode(null);
+      setAuthState("authenticated");
+    }
+  }, [passcode]);
+
```
gmaclennan:
```suggestion
```

src/frontend/context/SecurityContext.tsx
gmaclennan: What about the following flow?

1. User sets a passcode
2. User sets an obscureCode
3. User changes their passcode

gmaclennan (commented):
ok looking good. The `useEffect` (which would cause the flash of passcode screen when implemented) was still in there so I have suggested to remove it.

Now the AppPasscode screen state is re-organized I can better understand what the state logic is doing there. However, the `screenState` does just seem to be replicating navigation state, is there a reason for doing this rather than having them each as separate screens and using `navigate()` rather than `setScreenState()` to move between screens?

src/frontend/context/SecurityContext.tsx
ErikSin: That is what this [useEffect here](https://github.com/digidem/mapeo-mobile/pull/1013/files/6599a58d695a8f81a137cf915016026f37dbdce9#diff-476d432624050bbf2aca57f2cd9260583394f372946e24474c4994df203c8d16) was for, to tie the state of the obscure passcode to the passcode. In otherwords, when the passcode is set to null, the obscure code is set to null.

ErikSin:
> ok looking good. The `useEffect` (which would cause the flash of passcode screen when implemented) was still in there so I have suggested to remove it.
> 
> Now the AppPasscode screen state is re-organized I can better understand what the state logic is doing there. However, the `screenState` does just seem to be replicating navigation state, is there a reason for doing this rather than having them each as separate screens and using `navigate()` rather than `setScreenState()` to move between screens?

I can't find the original comment, but I believe that you had asked me to stay away from adding several pages to navigation since they were similar/doing one thing (setting password). Also the nice thing about this, is that a user/developer can't just navigate towards one of these pages when they are not in the right state. For example they would not be able to navigate to the turnOffPassword Screen if they do not have a password set. This allows the state of the security context to be tied to the state of the password screen

ErikSin:
@gmaclennan I just co-located the `obcure passcode` and `auth state` to also be set to null and 'authenticated' anytime `passcode` is set to null. This will make sure that the user is always in the right state AND that obscure passcode needs to be always set again when passcode is set

ErikSin:
> What about the following flow?
>
>    User sets a passcode
>    User sets an obscureCode
>    User changes their passcode

updated the flow to account for that

src/frontend/context/SecurityContext.tsx
ErikSin: Updated here: https://github.com/digidem/mapeo-mobile/blob/77f121f4c65557fa11e41276af0a14d928df0f68/src/frontend/context/SecurityContext.tsx#L79

gmaclennan:
> I can't find the original comment, but I believe that you had asked me to stay away from adding several pages to navigation since they were similar/doing one thing (setting password). Also the nice thing about this, is that a user/developer can't just navigate towards one of these pages when they are not in the right state. For example they would not be able to navigate to the turnOffPassword Screen if they do not have a password set. This allows the state of the security context to be tied to the state of the password screen

Maybe [this comment](https://github.com/digidem/mapeo-mobile/pull/1013#discussion_r982687843) which I had intended to mean the opposite (split into separate screens rather than do many things with the same screen). Your explanation makes sense though, and I can see the reasoning for doing it this way. What you loose though is the screen transitions and the back button working as you might expect, and extra work to keep the page header/title in sync, but I guess it's fine. If you are really worried about the user reaching a screen when not in the right state you can always add a `useEffect` to navigate away, but it seems like you are trying to add code to work around potential future bugs? However I'm ok with this pattern for now for the sake of getting this merged, I don't think it will have a major detrimental effect.

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,173 @@
...
+      // If user is able to set their own obscure passcode get rid of this if statement.
+      if (passcodeValue === OBSCURE_PASSCODE) {
+        throw new Error("passcode is reserved");
+      }
+
+      if (passcode !== null && passcodeValue === obscureCode) {
+        throw new Error("passcode is already being used obscure code");
+      }
+
+      if (!validPasscode(passcodeValue)) {
```
gmaclennan, line 83: This is a non-blocking comment:

A pattern that can work well for this is to change this to a `validatePasscode` function that throws an error, rather than something that returns a boolean. That way you can do other checks like the one above (`passcodeValue === obscureCode`), and you can throw a more meaningful error message than just "passcode not valid", e.g. it can be "passcode must be at least 5 characters". For all these errors though you will probably need to define an error code, so that the text can be translated in the frontend. At some point we will need some code to put all the errors in one place, but that is not a priority right now I think.

src/frontend/context/SecurityContext.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const setObscureCodeWithValidation = React.useCallback(
+    (newObscureVal: string | null | undefined) => {
+      if (passcode === null && newObscureVal !== null) {
+        throw new Error("Cannot set obscure mode if passcode not set");
+      }
+
+      if (newObscureVal === undefined) {
+        setObscureCode(OBSCURE_PASSCODE);
+        return;
+      }
```
gmaclennan, line 106:
```suggestion
    (newObscureVal: string | null = OBSCURE_PASSCODE) => {
      if (passcode === null && newObscureVal !== null) {
        throw new Error("Cannot set obscure mode if passcode not set");
      }

```

Non-blocking, just wanted to suggest this pattern as an option in the future.

PR #1015, opened 2022-09-22
chore: use node 12.19.0 to match nodejs-mobile

achou11:
The nodejs-mobile version we use runs 12.19.0, so probably makes sense to update the node version we're using to match


achou11:
Ah interesting, may have been a known issue but seems like this change breaks certain backend tests (looks like p2p upgrade?). @gmaclennan does this ring a bell and is it worth investigating right now/soon?

PR #1021, opened 2022-10-03
feat: app authentication and obscure code

ErikSin:
## App Authentication
features include:
- User can set an App Passcode
- When app passcode is set, a listener is turned on that listens for when the app goes out of focus. after the app goes out of focus, and back into focus, the app is locked.
- User can enable obscure mode
- Hides all observations if obscure mode is activated
- Hides Security Setting if user in obscure mode

## Other features this PR introduces:
- function added to security context to set auth state to unathenticated
- navigation screen groups are separated by file
- `appPasscode` was taken out of dev experiments
- `AppStack.tsx` returns `ScreenWithAuth` which conditionally renders the auth screen if auth state is set to `unauthenticated`

## To Do:
- [ ] Screen Shots/Gifs
- [ ] Persist Nav State after user types in passcode
- [ ] Update `AppNavigator.icca.tsx`


src/frontend/Navigation/AppStack.tsx
```diff
@@ -152,248 +39,19 @@ export const NavigatorScreenOptions: NativeStackNavigationOptions = {
...
-    />
-    <RootStack.Screen
-      name="ObscurePasscode"
-      component={ObscurePasscode}
-      options={{ headerTitle: intl(ObscurePasscode.navTitle) }}
-    />
-  </RootStack.Group>
-);
+export const ScreensWithAuth = (intl: (title: MessageDescriptor) => string) => {
+  const authState = useAuthState();
```
ErikSin: Can I use this hook here?

src/frontend/context/SecurityContext.tsx
```diff
@@ -69,6 +71,14 @@ const SecurityProviderInner = ({
...
   >(null);
 
+  const setToUnauthenticated = React.useCallback(() => {
+    if (passcode === null) {
+      throw new Error("Cannot set to unauthenticated without a password set");
+    }
+
+    setAuthState("unauthenticated");
+  }, [passcode]);
+
```
ErikSin: used by the listener to set to `unauthenticated`, which renders the passcode screen

src/frontend/AppContainerWrapper.tsx
```diff
@@ -34,6 +36,29 @@ const AppContainerWrapper = () => {
...
+          ) {
+            setToUnauthenticated();
+          }
+        }
+      }
+    );
+
+    return () => appStateListener.remove();
+  }, [authValuesSet.passcodeSet, setToUnauthenticated]);
+
```
gmaclennan: In my mind this belongs in the Security Context - I don't see why we need to split it out and have it here.

src/frontend/Navigation/AppStack.tsx
gmaclennan: you can't use state here, since this is not a component.

src/frontend/Navigation/AppNavigator.tsx
```diff
@@ -33,10 +26,7 @@ export const AppNavigator = () => {
       initialRouteName="Home"
       screenOptions={NavigatorScreenOptions}
     >
-      {createDefaultScreenGroup(formatMessage)}
-      {devExperiments.onboarding && createOnboardingScreenGroup(formatMessage)}
-      {devExperiments.appPasscode &&
-        createAppPasscodeScreenGroup(formatMessage)}
+      {ScreensWithAuth(formatMessage)}
```
gmaclennan: I'm not sure this navigation approach will work for the passcode? I haven't tested this on my device, but will this not reset the navigation state every time you return to the app? (because the screens are all removed from the stack then re-added).

I think what is needed is to navigate to the passcode screen when the appState changes to background? And to maybe set the passcode screen as the first screen when first mounting the app? I'm not sure though, I can think this through with you if helpful.

src/frontend/context/ObservationsContext.tsx
```diff
@@ -112,6 +115,15 @@ export const ObservationsProvider = ({
...
 
+  React.useEffect(() => {
+    if (authState === "obscured") {
+      dispatch({ type: "reload_success", value: [] });
+      return;
+    }
+
+    dispatch({ type: "reload" });
+  }, [authState]);
+
```
gmaclennan: I think rather than this approach, better to maybe modify the contextValue instead, to avoid the need to reloads and useEffect, which adds complexity to this already complicated context. E.g.

```js
  const contextValue: ObservationsContextType = React.useMemo(() => {
    const derivedState = { ...state }
    if (authState === 'obscured') derivedState.observations = {}
    return [derivedState, dispatch]
  },
     [state, authState]
   );

File: src/frontend/context/SecurityContext.tsx [hunk 6]
```diff
@@ -20,13 +20,15 @@ type SecurityContextType = {
   setAuthValues: (val: AuthSetters) => void;
   authenticate: (val: string | null, validateOnly?: boolean) => boolean;
   authState: AuthState;
+  setToUnauthenticated: () => void;
```
gmaclennan: See comment in AppContainerWrapper, I think all this logic can be contained in this context, without needing to expose this API.

gmaclennan (changes requested):
I'm not sure the navigation logic will work — it looks like it will reset navigation state when switching away from the app — but I have not tested this on a device.

src/frontend/Navigation/AppNavigator.tsx
ErikSin: Yea, i think if we do that, then we can change `authState` into `obscureCodeSet:boolean`, since we won't have to conditionally render anything. And then we would have to keep the `AppStateListener` inside the app container so it has access to navigation. But it should simplify things. I am going to try implementing it now

ErikSin: I updated the nav structure. It has the desired outcome of saving nav state, and I was able to get rid of authState.

Also when the user closes the app, they have to veer away from the app. when they veer away from the app, they are navigated to auth screen. Since nav state is saved, when they open the app again they are navigated to the auth screen.

gmaclennan: I'm not sure why we need to change this file and `AppStack` now? I think this would work fine without any changes to these files? The reason I raise it is concern about maintenance — this is all quite confusing because we are using functions that return React elements, which are not components, and this just adds a second layer to that.

src/frontend/Navigation/AppNavigator.tsx
```diff
@@ -33,10 +26,7 @@ export const AppNavigator = () => {
       initialRouteName="Home"
```
gmaclennan: Shouldn't this be set based on the value of authState?

src/frontend/screens/AppPasscode/index.tsx
```diff
@@ -24,14 +24,20 @@ export type PasscodeScreens =
...
-    authenticationValuesSet.passcodeSet ? "enterPasscode" : "intro"
+    authValuesSet.passcodeSet ? "enterPasscode" : "intro"
   );
 
+  React.useEffect(() => {
+    if (obscureModeOn) {
+      navigation.navigate("Settings");
+    }
+  }, [navigation, obscureModeOn]);
+
```
gmaclennan: What is this doing? As I understand it, if the user navigates here from the settings screen, and obscure mode is on, then they will immediately be navigated back? Or is this to catch when the user returns to the app from being on one of these screens and they turn the obscure mode on when returning? It would be good to document.

src/frontend/Navigation/AppNavigator.tsx
gmaclennan: Reading through the rest of the implementation I guess probably not, because we need the Home screen on the stack so that we can "goBack" to it when we auth.

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,87 @@
...
+    defaultMessage: "Incorrect passcode, please try again ",
+  },
+});
+
+export const AuthScreen = () => {
+  const [error, setError] = React.useState(false);
+  const { authenticate } = React.useContext(SecurityContext);
+  const [inputtedPass, setInputtedPass] = React.useState("");
+  const { goBack } = useNavigationFromRoot();
+
```
gmaclennan: We need to turn off the back button here.

gmaclennan (changes requested):
I'm concerned that the current approach will be hard to maintain and could lead to bugs, since authentication is now event based, and based on navigation, and it would be easy for a bug to slip through where we either forget to authenticate or accidentally send the user to the auth screen. Also switching the auth state to a boolean (obscureModeOn) increases the need to refactor later — I think it's always better to stick with string finite states. In the future we might also decide for extra security we don't load any observations in the unauthenticated mode (like we do in obscureMode). The current implementation makes it very hard to do something like only showing the passcode screen after the user is away from the app for a certain period of time — something that we might decide we need in the future (I can imagine with testing that the passcode screen coming up all the time could be frustrating).

I would suggest the following implementation:

1. The security context is responsible for authenticating and logging out, and we keep the original auth state in there. In the same security context, listen for the AppState and going into the background, and de-auth when that happens.
2. Either in the AppNavigator, or in the security context, a useEffect that fires when authState changes to unauthenticated, and navigate to the passcode screen.
3. In the AuthScreen, a useEffect that fires when authState changes to !== 'authenticated', and navigator.goBack().

Sorry for all the back-and-forth on this, but I think it's important that we get this code right and we don't open up security holes and potential future bugs — it's a dangerous piece of code because if we get it wrong we block someone's access to the app.

src/frontend/screens/AppPasscode/index.tsx
ErikSin: If the user is in obscure mode, we don't want any of the security features to be visible. In the setting, the security features are hidden when in obscure mode, so this just guarantees that the user does not navigate to those screen

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,110 @@
...
+        }
+      );
+      console.log(subscription);
+      return () => {
+        if (true) {
+          subscription.remove();
+        }
+      };
+    }, [])
+  );
```
ErikSin: @gmaclennan this should prevent the user from going back....but it doesn't seem to be working. Andrew and I debugged it for a bit, but maybe a 3rd set of eyes might be helpful

gmaclennan: Why are you wrapping in a useCallback? I think that is breaking things, e.g. you are returning within the useCallback, not the useFocusEffect, and why the `if (true)`?

achou11: @gmaclennan that's unfortunately the usage that's presented in the react-native-navigation docs. It's a pretty weird api

https://reactnavigation.org/docs/use-focus-effect/
https://reactnavigation.org/docs/custom-android-back-button-handling/

the `if (true)` is just an artifact of debugging 😅

gmaclennan: Ah, I see you are copying from the React Native docs. I'm not sure why they are implementing it like that? Have you also seen https://reactnavigation.org/docs/preventing-going-back/ ? What are you testing also? This will only override the hardware back button, not other ways of going back.

achou11: yeah we didn't look at that, that may be what's needed!

ErikSin: im only trying to prevent the hardware back button due (there is no header) AND the other approach is causing problems. The password is authenticated, and we try to go back() into the app, but the back is being prevented because the goBack() is being called before the next render

gmaclennan: If trying the other approach, then you need a check for `authState === 'authenticated'` in the `beforeRemove` handler, to bail early before calling `e.preventDefault`.

ErikSin: I have tried that. The problem is that `navigate.goBack()` and `setAuthState(authenticated)` are happening in the same function. Essentially, `goBack()` is always being called before the authState has changed. I can pass the remove listener function to the authenticate function though...kind of round about, but should work

gmaclennan: Yes, that is one of the reasons for my [suggestion in (3)](https://github.com/digidem/mapeo-mobile/pull/1021#pullrequestreview-1139183487) to goBack() as an effect as a result of auth state, rather than trying to do both in the handler.

src/frontend/screens/AuthScreen.tsx
```diff
@@ -0,0 +1,113 @@
...
+  React.useEffect(() => {
+    if (authState === "unauthenticated") {
+      navigation.addListener("beforeRemove", disableBack);
+    }
+  }, [authState, navigation, disableBack]);
+
+  function removeListenerAndGoBack() {
+    removeListener();
+    navigation.goBack();
+  }
```
ErikSin: I am passing the `removeListenerAndGoBack()` directly to the `authenticate` function in the `securityContext` @gmaclennan

src/frontend/screens/AuthScreen.tsx
ErikSin: I tried that initially. The problem is that navigation is only accessible in the `appContainerWrapper` and not in the security context. This mean, the `goBack` had to be handled in this wrapper. So having a `useEffect` that fired `goBack()` when the `authState === 'authenticated` has the undesired effect of `goBack()` being fired when the user was not using a passcode (because when they are not using a passcode, they are `authenticated`).

gmaclennan:
**Bug report:**

The error message when entering an incorrect passcode is obscured by the keyboard (at least on my phone).

https://user-images.githubusercontent.com/290457/196411020-ebfe42cb-eaa7-43b6-922d-21a783330109.mp4

gmaclennan:
**UX issue report**:

The "DEVICE SECURITY" heading, implemented as a list item (which shows an effect when tapped but does not do anything) does not match any of the other UI patterns in the app, and is a repeat of the screen heading anyway.

gmaclennan:
**UX Issue**:

The "App Passcode" screen seems to use a different font and layout than the other screens in our settings:

gmaclennan:
**UX Issue**:

The Kill Passcode screen seems to use another font and layout for what looks like a list item for "Use Kill Passcode". The text and checkbox are not aligned.

gmaclennan:
**UX Issue**:

In the passcode entry screen, the cursor and the asterisks that appear to hide the passcode are not centered in the input boxes.

https://user-images.githubusercontent.com/290457/196413334-6575323c-2d83-428a-8c8f-10f2922c471f.mp4

gmaclennan:
**Bug report**:

On the "Kill Passcode" screen you need to tap exactly on the checkbox in order to activate it. On the "App Passcode" screen the whole list item is a touch target (e.g. you don't need to tap precisely on the checkbox). I realize this is a pattern from the app experiments screens, but I wonder if we should change this? Perhaps a separate issue to raise outside this PR.

macandrow:
> **Bug report:**
> 
> The error message when entering an incorrect passcode is obscured by the keyboard (at least on my phone).
> auth-error-not-showing.mp4

cc/ @ErikSin 

For fix, if we could implement Keyboardavoidingview which should auto add the padding needed to bump the content up to account for the error message, or manually add the space to bump the content up. We're also missing the error state for the inputs  (wasn't sure notion links would work here but can be found in the UX space > Design system > Components) which is a red border and exclaim icon. Since the input is a set of inputs it would be adding the icon inside the last input. That together should fix. 

Preferable to moving the error message because this issue is sure to appear elsewhere in the app and we won't have the real estate to move the messages elsewhere, nor be able to build a consistent pattern for where users should look to see them. Another option is scaling down the design by a percentage for devices < x size, but I'm not sure what component would help with this and we'd have to see which breakpoint to apply it.

macandrow:
> 

I had no specific design guidelines on this, I assumed the cursor + content would be in the mid.

macandrow:
> **UX Issue**:
> 
> In the passcode entry screen, the cursor and the asterisks that appear to hide the passcode are not centered in the input boxes.
> 
>  passcode-entry.mp4

I had no specific design guidelines on this, I assumed the cursor + content would be in the mid.

macandrow:
> **UX issue report**:
> 
> The "DEVICE SECURITY" heading, implemented as a list item (which shows an effect when tapped but does not do anything) does not match any of the other UI patterns in the app, and is a repeat of the screen heading anyway.
> 
> 

This design followed the Android settings style at the time. Looking at my device now ( I think I'm 1 upgrade behind) that style has changed. If we no longer want to mimic the Android template the labels can be the same style and size as our observation labels (bold, not sure what the size in app). 

Fine to remove now: the design shows a divider and headers because of the device backup settings. But if we aren't adding that section then sure let's remove the dupe heading.

ErikSin:
> **UX Issue**:
> 
> The Kill Passcode screen seems to use another font and layout for what looks like a list item for "Use Kill Passcode". The text and checkbox are not aligned.
> 

your ability to pick up that they were not center aligned is quite impressive. I couldn't see it, but `alignItem` was not set to center!

gmaclennan:
> your ability to pick up that they were not center aligned is quite impressive. I couldn't see it, but `alignItem` was not set to center!

That kind of thing leaps out at me. Unaligned text is like fingernails on a chalk board for me!

ErikSin:
> **UX Issue**:
> 
> The Kill Passcode screen seems to use another font and layout for what looks like a list item for "Use Kill Passcode". The text and checkbox are not aligned.
> 

@gmaclennan this will change once the text is not hardcoded (aka, once the copy deck is completed)

ErikSin:
> **UX Issue**:
> 
> In the passcode entry screen, the cursor and the asterisks that appear to hide the passcode are not centered in the input boxes.

This issue is not happening on my phone, but similarly was happening on Andrew's Phone. The fix is to increase line height...but then this was causing problems on my device. I think we should leave this for now and flag it as a potential issue and log how many devices this is happening on (and fix it for the majority of devices).

gmaclennan (commented):
I haven't tested on a phone, but fixes look good and ok to wait on alignment.

Since merging this enables the feature, should we wait for copy to be done so that we don't have something in `develop` that can't be released?

ErikSin:
> I haven't tested on a phone, but fixes look good and ok to wait on alignment.
> 
> Since merging this enables the feature, should we wait for copy to be done so that we don't have something in `develop` that can't be released?

Copy has been updated, and I will merge after CI passes

PR #1023, opened 2022-10-10
Clean up map manager implementation

ErikSin:
- updated Map style context to expose `mapServerReady:boolean` and `onlineMapState:"unknown" | "online" | "offline"`
- added hook `UseDefaulStyleUrl()` which exposes the online or offline style url based on online status
- added hook `useMapServerState()' which exposes `mapServerReady:boolean`
- Bottom sheet `BGMapSelector` takes list of styles from parent `MapScreen`. This means the list is updated every time the bottom sheet is opened
- `OfflineMapLayers.js` converted to typescript


src/frontend/sharedComponents/AddButton.tsx
```diff
@@ -28,7 +28,6 @@ export const AddButton = React.memo(AddButtonNoMemo);
 const styles = StyleSheet.create({
   container: {
     position: "absolute",
-    zIndex: 10,
```
ErikSin, line 31: Was showing up over the bg map bottom sheet

src/frontend/context/MapStyleContext.tsx
```diff
@@ -25,16 +47,49 @@ export const MapStyleContext: React.Context<MapStyleContextType> = React.createC
...
+      didCancel = true;
+    };
+  }, []);
+
+  React.useEffect(() => {
+    let subscription: Subscription | undefined;
+
+    if (backgroundMaps) {
+      subscription = api.maps.addServerStateListener(({ value }) => {
+        setMapServerReady(value === "started");
```
achou11: nit: not sure how frequently the listener fires, but maybe use the functional update here.

```suggestion
        setMapServerReady(prev => {
          const next = value === "started"
          return prev === next ? prev : next
        );
```

probably not too much of an issue for now

src/frontend/hooks/useMapStyle.ts
```diff
@@ -1,55 +1,26 @@
...
-      styleType: Exclude<MapTypes, "loading">;
-      setStyleId: SetStyleId;
-    };
-
-function useLegacyStyle(): MapStyleState {
-  const [onlineState, setOnlineState] = React.useState<OnlineState>("unknown");
+type MapStyleState = {
+  styleUrl: null | string;
+  styleType: MapTypes;
+  setStyleId: SetStyleId;
```
achou11: would just inline the callback type here instead of having an explicit `SetStyleId` type.

src/frontend/hooks/useMapStyle.ts
```diff
@@ -67,45 +38,60 @@ function useLegacyStyle(): MapStyleState {
...
+    return {
+      styleType: "mapServer",
+      styleUrl: api.maps.getStyleUrl(styleId) || null,
+      setStyleId,
+      styleId,
+    };
+  }, [styleId, setStyleId, mapServerReady, defaultStyleUrl]);
 }
 
 export function useMapStyle(styleId: string = "default"): MapStyleState {
```
achou11: looks like the parameter here is no longer used or needed?

src/frontend/screens/MapScreen/MapScreen.tsx
```diff
@@ -18,21 +18,53 @@ import { MEDIUM_GREY } from "../../lib/styles";
...
 import { NativeHomeTabsNavigationProps } from "../../sharedTypes";
+import api from "../../api";
+import { useMapServerState } from "../../hooks/useMapServerState";
 const log = debug("mapeo:MapScreen");
 
+export interface MapServerStyle {
+  id: string;
+  url: string;
+  name?: string;
+}
```
achou11: more for future reference, this is probably something to expose from the map server module itself, but for more complete type defs to use in the client, can do a couple of things. For example, for representing this type:

```ts
import {StylesApi} from '@mapeo/map-server/dist/api/styles'

// Snippet from https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html to extract the item type in a nested type (like an array)
type Unpacked<T> = T extends (infer U)[]
  ? U
  : T extends (...args: any[]) => infer U
  ? U
  : T extends Promise<infer U>
  ? U
  : T;

// Optional intermediate type, in case it's helpful to have the array type to use somewhere else
type StylesListResponse = ReturnType<StylesApi['listStyles']>

// This will be properly be inferred as: `{ bytesStored: number; name: string | null; url: string } & IdResource`
type MapStyle = Unpacked<StylesListResponse> 
```

achou11: aside from above, the full type of this (if needed) should be:

```ts
export interface MapServerStyle {
  id: string;
  url: string;
  name: string | null;
  bytesStored: number;
}
```

src/frontend/screens/Settings/Experiments/BGMaps.tsx
```diff
@@ -22,6 +22,11 @@ const m = defineMessages({
     id: "screens.Settings.Experiments.BGMaps.goTo",
     defaultMessage: "Go to",
   },
+  BGMapsDescription: {
+    id: "screens.Settings.Experiments.BGMaps.BGMapsDescription",
+    defaultMessage:
+      "Use the background maps features to change the map background seen in the mapview. This feature will allow you to upload a raster MBtiles file. Once the file is uploaded, you will be able to set the background map in the mapview and use it even if you are not connected to the internet.",
```
achou11, line 28: vector should technically work here too (just that the style json produced for it will be defined by the map server and not the user)

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -39,19 +43,44 @@ const m = defineMessages({
...
 
+type ModalContent = "import" | "loading" | "error";
+
 interface BackgroundMap {
-  size: number;
-  title: string;
-  mapId: string;
+  id: string;
+  name?: string;
+  url: string;
```
achou11: should this be identical to `MapServerStyle`?

src/frontend/sharedComponents/AddButton.tsx
achou11: slightly worried that removing this may break something else. would check a couple of things:

1. any conflicts with observations on the map for touch interactions
2. the camera screen

src/frontend/sharedComponents/Map/MapView.js
```diff
@@ -326,9 +326,12 @@ class MapView extends React.Component<Props, State> {
     const { zoom, coords, following } = this.state;
     const locationServicesEnabled =
       location.provider && location.provider.locationServicesEnabled;
+
+    console.log("MAPVIEW", { styleURL, styleType });
```
achou11: guessing for debugging purposes 😄

src/frontend/sharedComponents/Map/MapView.js
```diff
@@ -352,7 +355,7 @@ class MapView extends React.Component<Props, State> {
               bugsnag.leaveBreadcrumb("onDidFinishRenderingMap")
             }
             onDidFinishRenderingMapFully={() => {
-              if (styleType === "legacy" && styleType !== "fallback") {
```
achou11: what's the reasoning for removing this? based on the comment below, seems that we want to handle the fallback differently for the zoom level

src/frontend/sharedComponents/BGMapCard.tsx
```diff
@@ -20,57 +18,42 @@ const m = defineMessages({
     defaultMessage: "MB",
     description: "The abbreviation for megabyte",
   },
+  unamedStyle: {
+    id: "sharedComponents.BGMapCard.unamedStyle",
+    defaultMessage: "Unamed Style",
```
achou11:
```suggestion
  unnamedStyle: {
    id: "sharedComponents.BGMapCard.unnamedStyle",
    defaultMessage: "Unnamed Style",
```

src/frontend/screens/Settings/Experiments/BGMaps.tsx
achou11:
```suggestion
      "Use the background maps features to change the map background seen in the map view. This feature will allow you to upload a raster MBtiles file. Once the file is uploaded, you will be able to set the background map in the map view and use it even if you are not connected to the internet.",
```

achou11 (changes requested):
No major changes requested just yet! This is one of those tricky things where we should ensure that the original behavior is preserved while adding and refactoring to include the new stuff. Have some slight concerns about that, but probably easier to expand when testing on an actual device (which I have yet to do but want to)

src/frontend/hooks/useMapStyle.ts
```diff
@@ -67,45 +38,60 @@ function useLegacyStyle(): MapStyleState {
...
     }
-  }, [onlineState, customMapState]);
+
+    return { styleType: "fallback", styleUrl: defaultMap, setStyleId };
+  }, [defaultMap, customMapState]);
 }
 
-function useMapServerStyle(): MapStyleState {
-  const [styleId, setStyleId] = React.useContext(MapStyleContext);
+function useMapServerStyle(defaultStyleUrl: string): MapStyleState {
```
achou11, line 59: think this is missing the case where the `styleId` isn't set to a specific id (e.g. initially enabling map management).

End up with data looking like `{ styleId: "", styleUrl: "http://127.0.0.1:37916/styles/", ... }` and since that style url doesn't point to an actual style, passing it to a map will cause this error to show up:

image

in this case, should probably serve the default style (identical to the condition you check for on line 72). alternatively, wondering if the initial/fallback value for the `MapStyleContext` should be different? maybe not, but another potential approach

src/frontend/sharedComponents/AddButton.tsx
ErikSin: I checked those things, and as far as I know, there aren't any other things that it could be hidden under

src/frontend/sharedComponents/BGMapCard.tsx
ErikSin: Im going to keep the ID the same because some of this has been translated in crowdin already

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -77,43 +115,56 @@ export const BackgroundMaps: NativeNavigationComponent<"BackgroundMaps"> = ({
...
+      try {
+        await api.maps.importTileset(results.uri);
+        const list = await api.maps.getStyleList();
+        setBackgroundMapList(list);
+        setModalContent("import");
+        sheetRef.current?.close();
+      } catch (err) {
+        console.log("FAILED TO IMPORT", err);
+        setModalContent("error");
+      }
```
achou11: there are cases when the bottom sheet doesn't close after the import finishes successfully. not sure if the issue lies here but something about the modal closing/opening logic probably needs tweaking

achou11 (changes requested):
Observations after testing on physical device (mostly non-blocking):

- Initially enabling the maps management causes the current location marker to disappear, even when i select the default map afterwards. restarting the app fixes this but guessing it's related to the error that i highlighted in one of the comments for this review
  - image

- Styling
  - should add some bottom padding for the scroll view here:
    - image

  - should add some bottom padding to the list here: 
    - image

src/frontend/sharedComponents/Map/MapView.js
ErikSin: I don't know...this was one of those things that happened months ago, and Im sure I had a reason. I am testing it out now in the app

ErikSin:
> Initially enabling the maps management causes the current location marker to disappear, even when i select the default map afterwards. restarting the app fixes this but guessing it's related to the error that i highlighted in one of the comments for this review

I am having trouble recreating this. Can you try again, and give me your step by step process

src/frontend/hooks/useMapStyle.ts
ErikSin: Also having trouble re-creating this. Can you walk me through the steps you did to get here?

src/frontend/screens/Settings/Experiments/BGMaps.tsx
ErikSin: I'm going to leave this for now because the copy is being updated!

achou11 (approved):

src/frontend/context/MapStyleContext.tsx
```diff
@@ -25,16 +47,52 @@ export const MapStyleContext: React.Context<MapStyleContextType> = React.createC
...
+
+    ky.get(normalizeStyleURL(onlineStyleURL, config.mapboxAccessToken))
+      .json()
+      .then(() => didCancel || setOnlineState("online"))
+      .catch(() => didCancel || setOnlineState("offline"));
+
+    return () => {
+      didCancel = true;
+    };
+  }, []);
```
gmaclennan, line 71: I'm fixing this now, but just wanted to drop a note somewhere about it: previously this would run every time the user returned to the map screen, to check if the user was online. This change (moving it into context) means that it only runs when the app first opens, so if the user moves offline then this will never change. As mentioned, ideally we use a net info module to detect an online/offline state change, or we poll, so that we can detect a change to offline state even without navigating away and back to the map.

PR #1026, opened 2022-10-19
chore: add icca intro screen dependencies - not building

RainbowGiantSquid:


ErikSin:
I have updated the screen from `@digidem/wcmc-mapeo-mobile-intro` to `@unep-wcmc/icca-mapeo-mobile-intro-screens` in [this PR](https://github.com/digidem/mapeo-mobile/pull/1038). I think you may have been using a different version of `npm` which was causing unnecssary file/dependency changes. This new PR should fix that.

Flow is being taken out of our code slowly, so no need to update flow typings

PR #1029, opened 2022-10-26
feat: map manager delete map

ErikSin:
# Deletion of offline map
User is able to delete an offline map.

Flow:
Map Settings>Background Maps>Select a map > Delete Map > Confirm Deletion

## Gifs:
Kapture 2022-10-25 at 17 23 21


src/frontend/api.ts
```diff
@@ -346,6 +346,7 @@ function createMapServerApi() {
     // Return the url to a map style from the map server
     getStyleUrl: (id: string): string | undefined =>
       mapServerPort ? `${getBaseUrl(mapServerPort)}styles/${id}` : undefined,
+    getTileset: async (url: string) => await (await ky.get(url)).json(),
```
ErikSin: Is there a better way to do this?

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -0,0 +1,184 @@
...
+    setStyleId(id);
+    navigation.navigate("Home", { screen: "Map" });
+  }
+
+  React.useEffect(() => {
+    api.maps
+      .getStyle(id)
+      .then(style => {
+        // @ts-ignore
+        return api.maps.getTileset(style.sources["raster-source"].url);
```
ErikSin: This needs to be improved. I'm just not sure how to do this endpoint.

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -138,6 +141,7 @@ export const BackgroundMaps: NativeNavigationComponent<"BackgroundMaps"> = ({
             isSelected={styleUrl === defaultStyleUrl}
             styleUrl={defaultStyleUrl}
             mapTitle="Default Map"
+            bytesStored={0}
```
ErikSin: Is there a way to get the total bytesStored for the default map?

src/frontend/api.ts
achou11: yeah i'd update to this for now for the sake of consistency with how the other endpoints are set up:

```suggestion
    getTileset: async (id: string): Promise<TileJSON> =>
      (await guaranteeClient().get(`tilesets/${id}`)) as TileJSON,
```

i think we can make improvements to the overall setup of the API for this but not within the scope of this PR.

ErikSin: but the problem is that I can get the whole url from the `sources`. I suppose I could try to parse just the ID...but that seems to be a little circular given that there actually is no straightforward way to get the tileset Id

achou11: yeah i see what you mean (i'm responding to your other question acknowledging this). i think i'm okay with the circularness for the sake of API consistency :)

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
achou11: Given what we have to work with, I think I'd approach it like this:

1. Fetch the style (you already have this)
2. Although maybe not as applicable in the current state of the map server, but keep in mind that a style can refer to several tilesets. Due to this, I think it would be wise to get information from all of the sources from its `sources` field. Each source is an object as described here: https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/
3. In our case, each source will have a `url` field that points to a saved tileset on the map server, with the format along the lines of `http://127.0.0.1:PORT/tilesets/:tilesetId`. Combined with my other [suggestion](https://github.com/digidem/mapeo-mobile/pull/1029#discussion_r1007144544), you'd have to extract the `:tilesetId`s here which is a little roundabout but should be okay to do using `URL` and some string operations.
4. Once all tileset ids have been extracted, use something like `Promise.all` (or `Promise.allSettled`, depending on desired behavior) to fetch the tilesets from the map server in parallel. Once those have been retrieved, check for the existence of the desired properties of each
    - since a style can refer to several tilesets, not exactly sure what the best case is for the `description` that you're wanting to display e.g. what if a style has 2 tilesets that each have their own description?
    - keep in mind that based on the [TileJSON spec](https://github.com/mapbox/tilejson-spec) the `maxzoom` property is optional and if it's not specified, the default is `30`. for the displayed zoom information, guessing you'd take the maximum value that exists among the tilesets that the style references? (basically if any of them don't specify `maxzoom`, then `30` would be the max based on the spec i suppose)
  
  maybe @gmaclennan has additional thoughts on approaching this that may be fewer steps

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -0,0 +1,184 @@
...
+      })
+      .then(obj => {
+        setZoomAndDescription({
+          zoom: obj.maxzoom.toString() || undefined,
```
achou11: nit: does the zoom need to be a string in any case? where you're rendering it in the JSX doesn't require it to be so

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -0,0 +1,184 @@
...
+  return (
+    <React.Fragment>
+      <View style={{ flex: 1, backgroundColor: WHITE }}>
+        <MapboxGL.MapView
+          styleURL={styleUrl}
+          compassEnabled={false}
+          zoomEnabled={false}
+          logoEnabled={false}
+          scrollEnabled={false}
+          style={{ height: "60%" }}
```
achou11: wondering if a max height should also be specified here, to account for smaller devices

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -226,6 +231,10 @@ export const BackgroundMaps: NativeNavigationComponent<"BackgroundMaps"> = ({
 
 BackgroundMaps.navTitle = m.BackgroundMapTitle;
 
+export function convertBytesToMb(bytes: number) {
+  return (bytes / 1024 ** 2).toFixed(0);
```
achou11: i'd be careful about the including the `toFixed` here, since it then returns a string and also is prone to loss of precision. would remove that part and let the caller of this function decide what to do with the numerical result

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
achou11: not entirely sure but would probably be a hardcoded number. i _think_ that lives in https://github.com/digidem/mapeo-server but can't fully remember

src/frontend/screens/Settings/MapSettings/DeleteMapBottomSheet.tsx
```diff
@@ -0,0 +1,145 @@
...
+  function deleteMap() {
+    // Cannot delete Default Map
+    if (mapId === DEFAULT_MAP_ID) {
+      return;
+    }
+
+    // If user is deleting the map that is currently being used, we want to set the map to be the default map
+    if (styleId === mapId) {
+      setStyleId(DEFAULT_MAP_ID);
+    }
```
achou11: this should probably happen only if the deletion was successful i.e. `api.map.deleteStyle` resolves

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
achou11: i think what you have here is fine though, i'm doing the same thing in https://github.com/digidem/mapeo-mobile/pull/1031 and just not displaying the info if the amount is 0. since you can't delete this map anyways, i think it's not as important to include the storage amount for it

src/frontend/screens/Settings/MapSettings/DeleteMapBottomSheet.tsx
```diff
@@ -0,0 +1,145 @@
...
+          <Text style={[styles.deleteButton, { color: MAPEO_BLUE }]}>
+            {t(m.cancel)}
+          </Text>
+        </Button>
+      </View>
+    </BottomSheet>
+  );
+});
+
+const styles = StyleSheet.create({
```
achou11, line 108: Maybe more of a follow-up but we could potentially extract and update https://github.com/digidem/mapeo-mobile/blob/develop/src/frontend/sharedComponents/BottomSheetModal/Content.tsx as a shared component for any bottom sheet content we create and use it for situations like this. otherwise it's more prone to visual inconsistency with the [design system](https://www.notion.so/digidem/d2b2903e77fc4a8cae6d6a8467e25b59?v=def381515abe45578065598c05682c75&p=f28600ee2800443a8bf0980e011933ba&pm=c)

based on the screenshot you provided, the amount of space below the bottom most button is too large

src/frontend/sharedComponents/BGMapCard.tsx
```diff
@@ -34,23 +35,34 @@ interface BGMapCardProps {
...
   const { position } = React.useContext(LocationContext);
-
+  const { navigate } = useNavigationFromRoot();
   return (
     <View
       style={[
         { borderColor: MEDIUM_GREY, borderWidth: 1, borderRadius: 2 },
         style,
       ]}
+      onTouchStart={() =>
```
achou11: nit: weirdly this prop doesn't seem to be documented in the react native docs (although it works). generally, should use a [Touchable](https://reactnative.dev/docs/0.66/handling-touches#touchables) if a touch interaction is needed for something like this

achou11 (changes requested):
Overall looking good. Think all of this work is exposing some limitations of the map server api, which is great! Feel free to create issues in the map server repo if you have any feedback about useful endpoints or additions to existing ones

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
ErikSin: Right...Then i'll just not show it!

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
ErikSin: the ui specifically rounds to the nearest whole number, hence the toFixed. But the function name is a little misleading I suppose

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
ErikSin: Curious why typescipt is not picking this up as an error, given that the typing of `zoom` is `number`

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
achou11: think this helper is more useful without prescribing the precision, so i'd lean towards calling toFixed wherever this function is being used to offer the flexibility for other potential uses

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
achou11: maybe due to the `@ts-ignore` you have above?

src/frontend/sharedComponents/BGMapCard.tsx
ErikSin: Good catch, this was originally a touchable, but we disabled the action, so i changed to a view, and didn't change it back

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
ErikSin: oh no, its actually because obj.maxzoom is of type `any`. Aka there is no associated typing

achou11: it's type any because the api call in the line above resolves to `Promise<any>`. TS isn't complaining because you ts ignored it. if you remove that and update the return type for the api call, then the `obj` type will be properly annotated

ErikSin: right, but actually he reason i am doing that is because if this error.
image

ErikSin: With` style.sources`, i am doing something wrong (I havent updated this endpoint yet, but the error is the `.url`). Any idea what is going on here

achou11: ah yeah you need to check for the existence of the property first and then access it. It's kind of a TS-specific thing related to discriminate unions, but i think something along the lines of this should work:


```ts
const source = style.sources["raster-source"]

// First check if the property exists, and if it does, ensure that it's not just undefined
if ('url' in source && source.url) {
  // Do stuff with `source.url`
}
```

EDIT: the `&& source.url` actually isn't necessary for TS since it'll infer the type to be `url: string` at that point, but helpful for runtime purposes

ErikSin: That error message is misleading...didn't realize that this was a simple optional/undefined type error

achou11: yeah it could be better but it's slightly different than what you're used to seeing because of how discriminate unions work, so a different kind of narrowing is needed

looks like this piece explains it well enough to provide context: https://css-tricks.com/typescript-discriminated-unions/ 

EDIT: Sorry i keep saying discriminate unions but really this is more generally applicable to union types in TS

ErikSin: interesting way of narrowing that is done in that article. Perhaps something we can adopt for some of our complicated types

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -0,0 +1,330 @@
...
+  route,
+  navigation,
+}: NativeRootNavigationProps<"BackgroundMapInfo">) => {
+  const { formatMessage: t } = useIntl();
+  const { bytesStored, id, styleUrl, name } = route.params;
+  const [zoom, setZoom] = React.useState<"loading" | number | null>("loading");
+
+  const sheetRef = React.useRef<BottomSheetMethods>(null);
+
+  const levelOfDetail = React.useMemo(() => {
```
achou11: does this have to be memoized? first thought is that this could be a function extracted outside of the component and then called like `const levelOfDetail = getDetailDescription(zoom)`

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -0,0 +1,330 @@
...
+            const url = source.url;
+            const searchWord = "tilesets/";
+            const startPosition = url.search(searchWord);
+            if (startPosition === -1) return;
+            const endPosition = url.length;
+            return url.substring(
+              startPosition + searchWord.length,
+              endPosition
+            );
+          }
```
achou11: seems like this will work but a little hard to parse. not sure if this is easier to read but would lean on using the `URL` object, which would also confirm that the url is a proper url:

```js
const url = new URL(source.url)

if (url.pathname.startsWith('/tilesets/')) {
  const splitPathname = url.pathname.split("/")
  return splitPathname[splitPathname.length - 1]
}
```

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -0,0 +1,330 @@
...
+        });
+
+        // Promise.allSettled does not work in React Native
+        // See https://github.com/facebook/react-native/issues/30236
+        return allSettled(
+          tileSetIdArray.map(id => {
+            if (id) {
+              return api.maps.getTileset(id);
+            }
+          })
```
achou11: think you can simplify this to not check if the id exists or not. if it doesn't, you're most likely dealing with an empty array `[]` as opposed to an array of undefined e.g. `[undefined, undefined]`. Mapping over an empty array will produce an empty array anyways

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -0,0 +1,330 @@
...
+  },
+  deleteButtonContainer: {
+    display: "flex",
+    flexDirection: "row",
+    justifyContent: "center",
+    alignItems: "center",
+  },
+});
+
+function allSettled<T>(promises: (Promise<T> | undefined)[]) {
```
achou11: don't think you need the `undefined` typing and associated code with it unless you know you're dealing with actual `undefined` values in the array. the standard `Promise.allSettled` only accepts an array of promises and breaks otherwise

src/frontend/screens/Settings/MapSettings/DeleteMapBottomSheet.tsx
achou11: have some of this done in https://github.com/digidem/mapeo-mobile/pull/1031 so feel free to ignore this, assuming the visual issues were resolved

src/frontend/screens/Settings/MapSettings/DeleteMapBottomSheet.tsx
```diff
@@ -0,0 +1,150 @@
...
+        }
+        navigate("BackgroundMaps");
+      })
+      .catch(err => {
+        console.log(err);
+      });
+  }
+
+  return (
+    <BottomSheet
```
achou11: not blocking but think we need to specify when to use bottom sheet vs bottom sheet modal. i think for the one spec'd in the design system, that will usually be a bottom sheet modal, whereas the one used in the main map screen should be a regular bottom sheet (as it's implemented right now). basically, the nature of the interaction for it will determine that (because the modal vs normal have different behaviors too), and maybe we need to consult Sabella on this too.

for where this component is used, I think this should technically be a modal too, but if it works okay then it's something we can revisit later.

started thinking about all of this stuff in https://github.com/digidem/mapeo-mobile/pull/1031 😅

achou11 (approved):
couple of relatively minor comments but otherwise think okay to merge

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
ErikSin: This is actually related to the [allSettled comment below](https://github.com/digidem/mapeo-mobile/pull/1029#discussion_r1012203260). When creating`tileSetIdArray`, typescript is recognizing `tileSetIdArray` as `(string|undefined)[]`.  Technically it should be a `string[]` || `[]`. Perhap I am missing something, but I believe I am creating `tileSetIdArray` to only be `string[]` || `[]` and typescript is equating `[]` to `undefined[]`.

achou11: ah yeah. instead of mapping over the object values i.e. `Object.values(style.sources).map...`, i would use a for loop i.e.:

```ts
const tilesetIds: string[] = []

for (const source of Object.values(style.sources)) {
  // do the url logic stuff and get the tileset id
  
  // push to the array
  tilesetIds.push(id)
}

return allSettled(tilesetIds.map(id => api.getTileset(id)))
```

this guarantees that the ids in `tilesetIds` are never undefined (you're only pushing string values to it)

doing map always returns a value for each item you iterate, so that `if` statement is causing TS to (properly) infer your array type as `(string | undefined)[]`, which is not what you want

src/frontend/screens/Settings/MapSettings/DeleteMapBottomSheet.tsx
ErikSin: I agree, we should come up with some standard, it seems like even `@gorham/bottom-sheet` doesn't have a definition. I think generally modal implies some sort of navigation and/or blocking the rest of the UI, which is what is happening here, so Ill change it to a modal

achou11: > I think generally modal implies some sort of navigation and/or blocking the rest of the UI

this is basically spot-on with what i was thinking. hopefully switching to our custom BottomSheetModal here will be less code too 😄

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -49,77 +61,77 @@ const m = defineMessages({
...
+  //   defaultMessage: "Some buildings, trees",
+  // },
+  // lvlOfDetail19: {
+  //   id: "screens.Settings.MapSettings.BackgroundMapInfo.lvlOfDetail19",
+  //   defaultMessage: "Local highway and crossing details",
+  // },
+  // lvlOfDetail20: {
+  //   id: "screens.Settings.MapSettings.BackgroundMapInfo.lvlOfDetail20",
+  //   defaultMessage: "A mid-sized building",
+  // },
```
achou11: is the plan for this to be restored at some point?

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -163,24 +173,43 @@ export const BackgroundMapInfo = ({
...
       .then(tileJson => {
-        const maxZoom = tileJson.reduce((highest, tilePromise) => {
-          if (!("value" in tilePromise)) return highest;
-          const tile = tilePromise.value;
-          if (!tile.maxzoom) return highest;
-          if (tile.maxzoom > highest) return tile.maxzoom;
-          return highest;
-        }, -1);
+        const initialVal: (undefined | number)[] = [undefined, undefined];
+        const zoomRanges: (undefined | number)[] = tileJson.reduce(
```
achou11: there's probably a simpler way of calculating this using `Math.min` and `Math.max`. Can't suggest something off the top of my head but basically collecting all of the min zooms and max zooms, and passing them to the math functions to get the full range

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -49,77 +61,77 @@ const m = defineMessages({
...
+  //   defaultMessage: "Local highway and crossing details",
+  // },
+  // lvlOfDetail20: {
+  //   id: "screens.Settings.MapSettings.BackgroundMapInfo.lvlOfDetail20",
+  //   defaultMessage: "A mid-sized building",
+  // },
 });
 
-type Zoom = "loading" | number | null;
+type Zoom = (number | undefined)[] | "loading";
```
achou11: think adding `"loading"` as a type here (which is then used in React state) is kind of an odd pattern. the loading can probably be derived based on a type that is more uniform, such as `[number, number] | null`

achou11, line 134: think adding `"loading"` as a type here (which is then used in React state) is kind of an odd pattern. the loading can probably be derived based on a value that is more uniform, such as `[number, number] | null`

achou11 (approved):
Have a few comments that are non-blocking but worth looking into as follow-ups

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
ErikSin: because it is an api call, I want to add a loading state AND the array can be `[undefined|undefined]` if there is no min/max zoom. if the array was [undefined, undefined], it would indicate the api was called and resolved (so no more loading state). 

I can replace `loading` with `null`. But the type would still have to be `(number | undefined)[] | null` to capture that undefined but non loading state

achou11: yeah i see what you mean, although I'm wondering if when `[undefined, undefined]` would come up given that the [TileJSON spec](https://github.com/mapbox/tilejson-spec/tree/master/3.0.0#312-maxzoom) states that if the minzoom and/or maxzoom fields aren't provided, then there are default values:

- minzoom default = 0
- maxzoom default = 30

given that, i guess `[undefined, undefined]` could be a result of an error in the API call? but if that's the case, then maybe a more explicit state variable for errors would be handy

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
ErikSin: yes, programs team would like to add this, just not for this release

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
ErikSin: Does optional with a default value imply that it can be undefined, but has a default value, or if it is undefined it will default to said value?

The typescript typing implies that minzoom and maxzoom can be undefined, so that would need to be addressed on the mapserver side. Given the vagueness of the `default value` and `undefined`, I wasn't treating it like an error, but I can see why this could be an error.

achou11: > The typescript typing implies that minzoom and maxzoom can be undefined, so that would need to be addressed on the mapserver side.

yeah this was something i considered adding but i left it as is because I didn't want the server to provide default values just yet. map clients will generally handle this so not sure how strictly enforced by the map server this has to be. it would be ideal to handle that via something like an OpenAPI spec, which we haven't gotten around to.

> Does optional with a default value imply that it can be undefined, but has a default value, or if it is undefined it will default to said value?

_think_ i understand the question but my answer will be the latter. I believe map clients will interpret the spec as in if there is no `minzoom` (or `maxzoom`) field in the JSON provided i.e. `undefined`, then assume it is the default value based on the **spec**, not some user-defined default

achou11: yeah the more im reading the spec, the more it's clear to me that the server DOES NOT need to provide the default value for optional fields (otherwise it's not an optional field by definition). the default values are for consumers/clients to adhere to

ErikSin: if thats the case, then I think keeping it at `(undefined|number)[]` makes the most sense. Do you think we should change `loading` to `null`? I personally like that it is explicitly a loading state, but I do not feel strongly about it either way

achou11: would lean towards using null in this case given that it's a more common pattern and general consistency (i don't think there are other places where we mix primitive types like strings and arrays).

for the array, if you're keeping the `[undefined, undefined]` possibility, i would suggest using tuple types i.e. `[number, number] | [undefined, undefined]`, since the range shouldn't be more or less than two elements in length

ErikSin: it could also be `[number, undefined] | [undefined, number] `. Aka max zoom could be present, but non min zoom and reverse. unless we want to define all 4 tuples, `(number|undefined)[]` feels like the most elegant way

achou11: that goes back to my point about the default values from the spec. if that's accounted for and used to replace undefined values, then the only combination is really `[number, number]`. `(number | undefined)[]` doesn't guarantee the number of elements in the array

ErikSin: perhaps i misunderstood, but the values can be undefined? Or are you saying that if it is undefined, we should change it to the default values?

achou11: > Or are you saying that if it is undefined, we should change it to the default values?

yeah this is what i was suggesting. so after traversing the tilesets to find out their min/max zooms, you may end up with one of the following, for example:

- `[undefined, 20]`. Update to `[0, 20]`
- `[10, undefined]`. Update to `[10, 30]`
- `[undefined, undefined]`. Update to `[0, 30]`

and set those converted values to the state, so that the rendering doesn't have to deal with undefined values at all

ErikSin: I don't think that makes sense in terms of what we are showing to the user. If I understand correctly, the min zoom and max zoom are used to indicate when to overzoom and when to stop fetching tiles. For example, if the zoom 12 is the max zoom, and the user tries to zoom to 13, it will just overzoom 12 instead of producing more tiles. therefore having default values of 0 and 30 make sense in telling the renderer when to overzoom/not fetch tiles, since those are pretty encompassing values on either side of the spectrum, and it would endlessly try to fetch tiles if there were no values.

In this case, we are simply letting the user know what zoom level they have downloaded. So in the case that it is undefined, we do not actually know if they have 0 or 30, so I think it would be misleading to default to those values.

achou11: yeah that's a good point! i was thinking that the displayed zoom levels would be communicating the _allowed_ (or attemptable) zoom levels for the map. given your point, is it better to only provide info on the end of the ranges that are known? e.g. for `[0, undefined]`, we'd only display `min zoom: 0` and nothing about the max zoom?

> In this case, we are simply letting the user know what zoom level they have downloaded.

from what i understand, not sure the fields from the tilejson are enough to know this definitively though, unless you analyze every _tile_ that is used for the tileset(s). for example, it could say that the allowed zoom range is `[0, 20]`, but you may only have downloaded tiles for `[0, 10]` zoom range, since it's still within the range.

ErikSin: Right, that is a good point about the range. Ill just show max zoom and min zoom, NOT as a range, and not show anything if it is undefined.

ErikSin: Oh, no, i just re-read your comment, and I misunderstood. That is a good point about the zoom ranges. I suppose for now, we will just have to work with what we have and we can reevaluate whether or not that is actually useful info to the user. Aka, we can see if that is misleading more often than not

achou11: yeah agreed, this is all experimental so let's get feedback (and maybe specifically ask for it?) and see what the best way to communicate that information is, if at all. all of this is a lot easier with fully online maps 😄

rudokemper: So, we were definitely thinking about the min and max zoom of the downloaded tiles. I don't think `minzoom` and `maxzoom` from the tilejson spec will be helpful for a Mapeo end user, even for online maps I don't think. Don't think there is anything in tilejson that specifies the extent of available tiles in a tileset. Not sure about the architecture of mbtiles in SQLite, but for an xyz tileset in directory format, you could count the number of directories in the {z} position. Could the same be possible for iterating through an mbtiles database to get these numbers?

achou11: got it. yeah in that case, much harder to figure out 😄 

> Could the same be possible for iterating through an mbtiles database to get these numbers?

yep! that's what we do [here](https://github.com/digidem/mapeo-map-server/blob/fa26e0a24e424fbc6f4fb5e739af8663902ddfb7/src/lib/mbtiles_import_worker.js#L145), so we could store that as some kind of metadata for the tileset and surface as a non-standardized field for the tilejson served for that tileset. this wouldn't be as sufficient for online-capable maps (because you can continuously download tiles after import), which we're not really dealing with for now so not as much of an immediate concern

ErikSin: @rudokemper thanks for the insight. So would that only expose the max zoom if we did that?

rudokemper: True. Rather than doing a count, guess you could also just take the lowest and highest number of the {z} directories. (Again with the same caveat that I'm not sure how the mbtiles architecture differs from xyz directory).

achou11: yeah we'd take the min and max values of `z` found when iterating, as opposed to counting anything

ErikSin: @achou11 is this something we can expose this release? Or would you rather we got rid of this for now, and add it to a subsequent release?

achou11: i'd lean towards adding in a subsequent release just based on time constraints. i also have https://github.com/digidem/mapeo-map-server/pull/82 which would be nice to include in a new release of the map server too

achou11: we may even be able to do this at runtime by extracting the quadkeys from the tiles that style references and taking the min+max of the result. Only concern would be potential performance penalties (i don't think quadkeys are sorted, but could be wrong??)

achou11: added an issue for the map server here: https://github.com/digidem/mapeo-map-server/issues/83

ErikSin: I have removed this, and I will add this as a ticket to do in the future

PR #1030, opened 2022-10-26
Chore: add bottom sheet to confirm passcode

ErikSin:
## Bottom Sheet to set Passcode

User is prompted with a bottom sheet before they are able to set passcode!

Pressing `Cancel`, will not set the passcode, and take user back to Security Screen.
Press `Save App Passcode` will set passcode and will take user back to Security Screen.

Kapture 2022-10-26 at 12 46 10


src/frontend/sharedComponents/BottomSheetModal/Content.tsx
```diff
@@ -108,9 +121,6 @@ const styles = StyleSheet.create({
     fontSize: 20,
     textAlign: "center",
   },
-  buttonsContainer: {
-    paddingHorizontal: 20,
-  },
```
achou11, line 113: this change will break other consumers of the bottom sheet modal component. based on what's described in the [design spec](https://www.notion.so/digidem/d2b2903e77fc4a8cae6d6a8467e25b59?v=def381515abe45578065598c05682c75&p=f28600ee2800443a8bf0980e011933ba&pm=c), this shouldn't be changed. if there's a spacing issue when using this, probably should update the consumer's styles instead of this one

achou11 (changes requested):
Overall looks good. One notable change that I contested

src/frontend/sharedComponents/BottomSheetModal/Content.tsx
ErikSin: Ok, ill override it on my component

ErikSin: Never mind, ill just leave the design spec. It is causing the text to be over 2 lines (but i have a zoomed in font on my phone). But i think that is better than overriding the design spec

achou11: ah interesting, wonder how to address that given translations and all, aside from using more terse language. think those are questions for the design spec in general though :)

achou11 (approved):

Issue #233, opened 2019-09-25
SIGABRT in MainActivity


gmaclennan:
## Error in Mapeo Mobile

**SIGABRT** in **MainActivity**
Abort program

[View on Bugsnag](https://app.bugsnag.com/digital-democracy/mapeo-mobile/errors/5d8bc21a4912350019d43043?event_id=5d8bc21a0050391191bb0000&i=gh&m=ci)

## Stacktrace

    libc.so:251736 - tgkill

[View full stacktrace](https://app.bugsnag.com/digital-democracy/mapeo-mobile/errors/5d8bc21a4912350019d43043?event_id=5d8bc21a0050391191bb0000&i=gh&m=ci)

*Created automatically via Bugsnag*

Issue #215, opened 2019-09-17
ReferenceError in /data/data/com.mapeo.qa/files/nodejs-project/index.js:109422


gmaclennan:
## Error in Mapeo Mobile

**ReferenceError** in **/data/data/com.mapeo.qa/files/nodejs-project/index.js:109422**
q is not defined

[View on Bugsnag](https://app.bugsnag.com/digital-democracy/mapeo-mobile/errors/5d80fbbc5a0721001cdc16b3?event_id=5d80fbbc00505217d76b0000&i=gh&m=ci)

## Stacktrace

    /data/data/com.mapeo.qa/files/nodejs-project/index.js:109422 - Api.syncLeave

[View full stacktrace](https://app.bugsnag.com/digital-democracy/mapeo-mobile/errors/5d80fbbc5a0721001cdc16b3?event_id=5d80fbbc00505217d76b0000&i=gh&m=ci)

*Created automatically via Bugsnag*

PR #1031, opened 2022-10-26
Implement map import progress

### LINKED ISSUE #1017: Map Server Import Progress
# Import Progress
When user adds a new map, the app should show the estimated amount of time for the new app to be ready to use.

This will use the map server endpoint [GET /imports/progress/:importId](https://github.com/digidem/mapeo-map-server/blob/master/API.md#get-importsprogressimportid)

The time remaining can be calculated using `sofar`/`total`

# Designs
Import progress bar should go where indicated by red arrow, and should replace the size of the map

Image

### LINKED ISSUE #1018: Map Server Imported Bytes
# Show offline map size
Displays to the user the total file size (in mb) of an offline map.

To Do:
- [ ] use [get styles endpoint](https://github.com/digidem/mapeo-map-server/blob/master/API.md#get-styles)
- [ ] for each style, map the bytes onto the list of styles
- [ ] convert bytes to KB or MB as approriate

### Figma:
https://www.figma.com/file/sxz2NvysPVab6ZTqs6f57K/Cobra-Collective?node-id=183%3A800 

Image

achou11:
Closes #1017
Closes #1018

Notes:

- in addition to adding map import progress, this became a massive refactor of the map server integration as a whole (thanks Gregor for leading that work!). 
- also makes various UI improvements for different map management screens

- Potential improvements:
  - think there are still some rough spots with the error handling, and I haven't been able to fully test various error scenarios just yet.

Preview:

- Successful import

  map-import-pt-1

- Switching maps
  
  map-import-pt-2

- General screenshot

  image


src/frontend/context/BackgroundedMapImportsContext.tsx
```diff
@@ -0,0 +1,70 @@
+import * as React from "react";
+
+type BackgroundedMapImportsState = Record<string, string>;
```
ErikSin: For Readability: `type BackgroundedMapImportsState = Record<MapServerStyleInfo["id"], string>;`

patches/react-native-fetch-api+3.0.0.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/node_modules/react-native-fetch-api/src/Fetch.js b/node_modules/react-native-fetch-api/src/Fetch.js
```
ErikSin, line 1: Can we document this patch somewhere?

src/frontend/sharedComponents/BGMapCard.tsx
```diff
@@ -23,86 +29,258 @@ const m = defineMessages({
...
+
+function shouldShowProgressBar(
+  importStatus: ImportStatus
+): importStatus is {
+  status: "progress" | "complete";
+  soFar: number;
+  total: number;
+} | null {
+  return (
+    importStatus === null ||
```
ErikSin: We want to show progress bar when null?

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
```diff
@@ -89,88 +130,121 @@ export const BackgroundMaps: NativeNavigationComponent<"BackgroundMaps"> = ({
...
-              mapId={bgMap.id}
-              style={{ marginTop: 20 }}
-              styleUrl={bgMap.url}
-              isSelected={styleUrl === bgMap.url}
-              mapTitle={bgMap.name}
-            />
-          ))
-        )}
-      </ScrollView>
+  const bottomSheetContentProps: React.ComponentProps<typeof BottomSheetContent> =
```
ErikSin: Too many dependencies to memoize this gracefully I assume?

ErikSin (commented):
Just did an code review, but I will try it on my phone later!

patches/react-native-fetch-api+3.0.0.patch
achou11: Does it help to document as a comment in the patch itself?

src/frontend/sharedComponents/BGMapCard.tsx
achou11: yeah it's kind of confusing and maybe there's a better way to represent the state using an explicit status string, but basically when the `importStatus` is null, it could mean that we've started an import but haven't received any events back yet. In that case, we want to show the progress bar because it will be in the indeterminate state.

the reason I'm using a null value at all is because if the map's already been imported some time before e.g. yesterday, then there's no need to have an importStatus associated for it, hence null.

would it be helpful to add an import status that represents this "waiting" situation? e.g. extends the `ImportStatus` type so that there's a variant like `{ status: 'waiting' }`?

src/frontend/screens/Settings/MapSettings/BackgroundMaps.tsx
achou11: yeah probably not worth memoizing. didn't observe any notable issues as is

patches/react-native-fetch-api+3.0.0.patch
ErikSin: I suppose so. Or we could start a readme for all the patches? What do you think?

gmaclennan: There [is a readme for all the patches](https://github.com/digidem/mapeo-mobile/blob/develop/patches/README.md) that I had been keeping up-to-date. Seems like it is a bit behind now, we should document everything there.

ErikSin: @gmaclennan nice! great, so lets do it there!

src/frontend/sharedComponents/BGMapCard.tsx
ErikSin: Yeah, it can't hurt to have that explicit status of `waiting`. And looking through the code, it would simplify some of your null check logic

src/frontend/context/BackgroundedMapImportsContext.tsx
ErikSin: I suppose we should actually do `Record<MapServerStyleInfo["id"], Imports["id"]>`

achou11: need to fix a type in the map server dep in order for `Imports["id"]` to work properly, so probably will leave that part as string for now. see 33024293f9b8a2803b5f16822674f34c818da19b

patches/react-native-fetch-api+3.0.0.patch
achou11: see 2b517c24023df1b814a4c77055f9268d93b6a412

src/frontend/sharedComponents/BGMapCard.tsx
achou11: see 4012d1da791f78ebcea8a7a0f74189ac938e6f37

achou11: did additional work on this via 8bcad51dd905de3f2a7c9e21ba007b9aeb7fe0a7 and think it's much easier to follow now. basically removes the `null` variant and updates the wording of `"waiting"` to `"idle"`

achou11:
Anyone who tries this out on a real device, there's currently an issue where if an import error occurs, while the bottom sheet appears, some infinite rendering or something occurs and locks up the UI, eventually crashing. So yeah, error handling still needs work (think it's related to the event source stuff that I'm not handling properly. not the easiest to reproduce either...)

ErikSin:
> Anyone who tries this out on a real device, there's currently an issue where if an import error occurs, while the bottom sheet appears, some infinite rendering or something occurs and locks up the UI, eventually crashing. So yeah, error handling still needs work (think it's related to the event source stuff that I'm not handling properly. not the easiest to reproduce either...)

Yeah, I have not been able to reproduce this error. Do you have any ideas of what things I should try to reproduce, so I can help debug?

achou11:
> Yeah, I have not been able to reproduce this error. Do you have any ideas of what things I should try to reproduce, so I can help debug?

using a bad mbtiles file would probably be the most effective approach. we have a few here that we should try using, since we know that they're bad 😄 

https://github.com/digidem/mapeo-map-server/tree/master/test/fixtures/bad-mbtiles

achou11:
I think i fixed the ui freezing issue with 38004d253168904657750350544c014f740f17eb

achou11:
as of 3ea59d2643978bcf23c2264cb8211c75d5420508, this is really a full-on refactor of the map server integration with mobile, to account for challenges presented by the import progress feature. don't have the energy to detail all of the changes but will try to do so tomorrow.

a couple of issues still remain, proper flows + updates when imports error or complete. there are some specific cases that aren't fully working yet

gmaclennan:
fix(server): don't add event listeners on start

The map-server::state event listeners are added on server initialization,
so they should not be added and removed on start and stop.
```diff
diff --git a/src/backend/server.js b/src/backend/server.js
index 2ef1f87d..6644d5e3 100644
--- a/src/backend/server.js
+++ b/src/backend/server.js
@@ -267,9 +267,6 @@ function createServer({
           return;
         }
 
-        handleGetState = createGetStateHandler(mapServer);
-        rnBridge.channel.on("map-server::get-state", handleGetState);
-
         mapServer
           .start()
           .then(() => {
@@ -283,20 +280,6 @@ function createServer({
       stop: () => {
         log("Stopping map server");
 
-        if (handleGetState) {
-          rnBridge.channel.removeListener(
-            "map-server::get-state",
-            handleGetState
-          );
-        }
-
-        if (handleReportErrorState) {
-          rnBridge.channel.removeListener(
-            "map-server::get-state",
-            handleReportErrorState
-          );
-        }
-
         // TODO: Handle this better
         if (!mapServer) {
           log("Map server instance does not exist");

```

achou11:
> fix(server): don't add event listeners on start
> 
> The map-server::state event listeners are added on server initialization, so they should not be added and removed on start and stop.
> 
> ```diff
> diff --git a/src/backend/server.js b/src/backend/server.js
> index 2ef1f87d..6644d5e3 100644
> --- a/src/backend/server.js
> +++ b/src/backend/server.js
> @@ -267,9 +267,6 @@ function createServer({
>            return;
>          }
>  
> -        handleGetState = createGetStateHandler(mapServer);
> -        rnBridge.channel.on("map-server::get-state", handleGetState);
> -
>          mapServer
>            .start()
>            .then(() => {
> @@ -283,20 +280,6 @@ function createServer({
>        stop: () => {
>          log("Stopping map server");
>  
> -        if (handleGetState) {
> -          rnBridge.channel.removeListener(
> -            "map-server::get-state",
> -            handleGetState
> -          );
> -        }
> -
> -        if (handleReportErrorState) {
> -          rnBridge.channel.removeListener(
> -            "map-server::get-state",
> -            handleReportErrorState
> -          );
> -        }
> -
>          // TODO: Handle this better
>          if (!mapServer) {
>            log("Map server instance does not exist");
> ```

little bit confused by this comment. the suggested removals are for `::get-state`, not `::state`. Is adding event listeners for that unnecessary?

gmaclennan:
> little bit confused by this comment. the suggested removals are for `::get-state`, not `::state`. Is adding event listeners for that unnecessary?

Sorry I meant to say `map-server::get-state`.

[Line #215](https://github.com/digidem/mapeo-mobile/blob/98de3b2c3f2ea0ed22849b827ec8ed6fb509b233/src/backend/server.js#L215) adds a `get-state` event listener during the `createMapServer` initialization, and [Line #233](https://github.com/digidem/mapeo-mobile/blob/98de3b2c3f2ea0ed22849b827ec8ed6fb509b233/src/backend/server.js#L233) adds the special error handler incase the map server did not initialize.

The code I deleted was:

A) During server start it was adding a second `get-state` listener, doing the same thing as the listener added during initialization.
B) During server stop it was removing the `get-state` listener, but since it was attached during initialization we don't need to add and remove it during start and stop.
C) The removal of the special error handler listener during stop was redundant because that code would never be reached (it was only attached if `mapServer === undefined`) but even if it was reached then we would not want this listener to be removed (imagine the user opens the app, map-server fails to start, so error listener attached, then user switches away (server stop) then switches back (server start), now the error listener is removed, so sending a `get-state` event from the client would not work as expected.

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
```diff
@@ -0,0 +1,326 @@
...
+};
+
+/** Custom backdrop that will not close sheet when pressed */
+const BackdropNonPressable = (props: BottomSheetBackdropProps) => {
+  return (
+    <BottomSheetBackdrop {...props} opacity={0.3} pressBehavior={"none"} />
+  );
+};
+
+const MapImportBottomSheet = React.forwardRef<MapImportBottomSheetMethods, {}>(
```
achou11, line 158: depending on the behavior we want, this could reuse our BottomSheetModal component. using the BottomSheet was to account for issues with collapsing/closing and re-opening the sheet in sequence, which was only necessary because of my initial implementation of the error-handling.

think with Gregor's refactor and the UX changes for errors on this screen, we could scrap the usage of the BottomSheet and stick with BottomSheetModal, since that more closely reflects the usage we want (sheet that has blocking actions)

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
```diff
@@ -0,0 +1,326 @@
...
+  (props, ref) => {
+    const [state, setState] = React.useState<BottomSheetStates>("select");
+    const bottomSheetRef = React.useRef<BottomSheetMethods>(null);
+    const { snapPoints, updateSheetHeight } = useSnapPointsCalculator();
+    const { add: addMapImport } = useMapImportsManager();
+
+    React.useImperativeHandle(ref, () => {
```
achou11, line 165: wow first time i've seen useImperativeHandle being used. that's pretty cool! maybe `BottomSheetModal` should be doing something similar (not in this PR's scope, obviously)

src/frontend/hooks/useMapStyles.ts
```diff
@@ -0,0 +1,125 @@
...
+          });
+          setStylesList(stylesWithImportState);
+          setStatus("success");
+        })
+        .catch(() => didCancel || setStatus("error"));
+
+      return () => {
+        didCancel = true;
+      };
+    }, [activeImports, backgroundMaps])
```
achou11, line 59: i added this deps array in some previous commit because it would otherwise potentially infinitely render. does this look right @gmaclennan ? I think where this falls a little short is when mutative operations from the client occur (e.g. deleting a style), this doesn't really trigger. Could potentially define those actions in this hook and use some dummy state token to manually trigger this, I guess 🤔

src/frontend/polyfills/fetch.ts
```diff
@@ -0,0 +1,34 @@
...
+  fetch,
+}));
+
+// Polyfill document object (needed for @microsoft/fetch-event-source)
+polyfillGlobal("document", () => ({
+  ...(global.document || {}),
+  hidden: false,
+  addEventListener: () => {},
+  removeEventListener: () => {},
+}));
```
achou11, line 34: TODO: probably should write type defs that extend the global

achou11:
Broke this PR up into 3 stacked ones. Hopefully makes it easier to review. Will keep this one open for now until someone says we should go with the stacked ones or this one 😄 

1. #1048
2. #1049
3. #1050

achou11:
closing this in favor of the partitioned ones mentioned in https://github.com/digidem/mapeo-mobile/pull/1031#issuecomment-1317395713

PR #1035, opened 2022-10-31
fix: fix build issues caused by jcenter shutdown

achou11:
Looks like jcenter has finally been shutdown and as a result, certain deps that we use don't work without some intervention. This PR does the following:

1. updates the build config to strip android dep references to the jcenter repository (see https://github.com/facebook/react-native/issues/35136 for context)
2. removes jcenter repo from our build config
3. adds necessary extra configuration for `@react-native-mapbox-gl/maps` in order to pull certain Mapbox SDK deps from their password-protected repo.
4. removes an unused dep (`react-native-build-config`)

re: 3, i'm a little hesitant about this because:

- we have to update our CIs to include an access token
- adds another barrier-to-entry for contributors (now they need get or create an access token, most likely requiring a mapbox account).

ideally, these would be a temporary measure and we'd update our usage of the library to use the MapLibre SDK, which i believe doesn't have these limitations (they publish to Maven Central).

for additional context, I think if https://github.com/mapbox/mapbox-gl-native-android/issues/650 and https://github.com/mapbox/mapbox-gl-native-android/issues/630 were addressed, then maybe we wouldn't need the additional config steps, but not entirely sure if the `accounts` dep is the only problematic one

TODO BEFORE MERGING:

- ~@gmaclennan we will probably need some intervention if we're okay with the changes here, mainly updating our CI to add an access key and making sure our build config reads it~ DONE


package.json
```diff
@@ -82,7 +82,6 @@
     "react-intl": "^5.4.5",
     "react-native": "0.66.4",
     "react-native-android-open-settings": "^1.3.0",
-    "react-native-build-config": "^0.3.2",
```
ErikSin, line 85: Was this dependency just not being used?

achou11: Yeah that's what it looked like. It publishes to jcenter so figured just worth removing to avoid unnecessary potential issues (although I don't think it caused any since it wasn't being used)

although, i'm starting to wonder if maybe i had asked about this before. vaguely remember @gmaclennan saying it was being used indirectly but could be misremembering.

ErikSin: Oh, nevermind I see that it wasn't being used in your PR description!

achou11:
Based on brief internal discussion, we should look into just switching to MapLibreGL.

achou11:
Update: decided that for now we'll just move forward with the secret token nonsense because we're not in a position to attempt to switch over to MapLibreGL just yet. It's something we're definitely interested in doing later on and we should open an issue that references this PR in order to fully understand the context for doing so

achou11 (commented):
@gmaclennan not the most familiar with git secret so hoping 36b043f9386de16508d241f90a0f0ef4cd70411c looks alright in terms of the files changed and the nature of the changes

android/build.gradle
```diff
@@ -1,5 +1,39 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
+// Load the .env file in the repo and set project environment variables if found
+def loadDotEnv() {
```
achou11, line 4: moved this from `android/app/build.gradle` since this file seems to run first and can set the properties read by the app build file

PR #1037, opened 2022-11-01
chore: fix default background color for navigation screen content

achou11:
Fixes a regression caused by https://github.com/digidem/mapeo-mobile/pull/992/files#diff-3fe14a68ca2c8f88a4ef1e4dcef629f5b7d0f42115bb9f5936fd766b802e7982L140-L142 where the default background color for the screen is a light grey instead of white.

Using `chore` for commit title since the regression isn't yet published, so in theory this fix should not show up in the changelog when we do so

Preview:

- Before:

  image

  image

- After:
  
  image

  image


ErikSin (approved):
Thanks

PR #1040, opened 2022-11-07
Fix: Bump react-native version to fix build failures

### LINKED ISSUE #1039: Failure to build due to Facebook/Meta issue
Since Nov 4th this will fail to build due to https://github.com/facebook/react-native/issues/35210.

Fix coming soon, just need to update to react-native@0.66.5

gmaclennan:
**DO NOT MERGE**

On Nov 4th 2022 Facebook pushed a new version of react-native that [broke react-native builds for everyone else](https://github.com/facebook/react-native/issues/35210). The suggested fix is to bump react-native from v0.66.4 to v0.66.5, which is what this PR does.

Fixes: #1039 

However, this patch also seems to break react-native-v8, since running the app with this PR results in the following crash:

```
11-07 13:14:19.023 17803 17862 F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x109 in tid 17862 (mqt_js), pid 17803 (com.mapeo.debug)
11-07 13:14:19.063 17872 17872 E crash_dump32: failed to interrupt 17873 to detach: No such process
11-07 13:14:19.080 17872 17872 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
11-07 13:14:19.080 17872 17872 F DEBUG   : Build fingerprint: 'Android/sdk_phone_x86/generic_x86:9/PSR1.180720.012/4923214:userdebug/test-keys'
11-07 13:14:19.080 17872 17872 F DEBUG   : Revision: '0'
11-07 13:14:19.080 17872 17872 F DEBUG   : ABI: 'x86'
11-07 13:14:19.080 17872 17872 F DEBUG   : pid: 17803, tid: 17862, name: mqt_js  >>> com.mapeo.debug <<<
11-07 13:14:19.080 17872 17872 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x109
11-07 13:14:19.080 17872 17872 F DEBUG   : Cause: null pointer dereference
11-07 13:14:19.080 17872 17872 F DEBUG   :     eax 00000104  ebx c787c788  ecx 0000003e  edx 47fc5481
11-07 13:14:19.080 17872 17872 F DEBUG   :     edi c787c838  esi c22d0ea0
11-07 13:14:19.080 17872 17872 F DEBUG   :     ebp c787c778  esp c787c730  eip c880101e
11-07 13:14:19.084 17872 17872 F DEBUG   :
11-07 13:14:19.084 17872 17872 F DEBUG   : backtrace:
11-07 13:14:19.084 17872 17872 F DEBUG   :     #00 pc 00ef501e  /data/app/com.mapeo.debug-ZGCcIhvtNt9b-qjZhWkcog==/lib/x86/libv8android.so (v8::internal::LookupIterator::State v8::internal::LookupIterator::LookupInRegularHolder<false>(v8::internal::Map, v8::internal::JSReceiver)+446)
11-07 13:14:19.084 17872 17872 F DEBUG   :     #01 pc 00ef295f  /data/app/com.mapeo.debug-ZGCcIhvtNt9b-qjZhWkcog==/lib/x86/libv8android.so (v8::internal::LookupIterator::State v8::internal::LookupIterator::LookupInSpecialHolder<false>(v8::internal::Map, v8::internal::JSReceiver)+351)
11-07 13:14:19.084 17872 17872 F DEBUG   :     #02 pc 00ef210e  /data/app/com.mapeo.debug-ZGCcIhvtNt9b-qjZhWkcog==/lib/x86/libv8android.so (void v8::internal::LookupIterator::Start<false>()+142)
11-07 13:14:19.084 17872 17872 F DEBUG   :     #03 pc 01033d96  /data/app/com.mapeo.debug-ZGCcIhvtNt9b-qjZhWkcog==/lib/x86/libv8android.so (v8::internal::Runtime::GetObjectProperty(v8::internal::Isolate*, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>, bool*)+470)
11-07 13:14:19.084 17872 17872 F DEBUG   :     #04 pc 0103a595  /data/app/com.mapeo.debug-ZGCcIhvtNt9b-qjZhWkcog==/lib/x86/libv8android.so
11-07 13:14:19.084 17872 17872 F DEBUG   :     #05 pc 00acd8f6  /data/app/com.mapeo.debug-ZGCcIhvtNt9b-qjZhWkcog==/lib/x86/libv8android.so
```


gmaclennan:
Update: using react-native-v8@1.4.8 seems to fix this and the app runs, @achou could you also test, and we'll see if CI works ok.

achou11:
@gmaclennan just tried this on my device and things seemed to work as expected. noticed CI is still failing though (not sure if due to flakiness or something more legitimate)

gmaclennan:
Looks like the app is not running/opening on CI at all, so need to fix that before we merge this.

achou11:
Summary of changes I added:

- update jest cli to match jest version. this was preventing me from running the e2e tests locally. generally, i think it's supposed to match the jest version being used
- upgrade detox to most recently published version (19.13.0). the [release notes for 19](https://github.com/wix/Detox/releases/tag/19.0.0) claim that it's not a breaking change for basic users, which i think we fall into. happy to revert if we're worried about this and want to minimize the reach of this PR
- replace v8 with jsc intl. we needed to upgrade v8 to account for the React Native upgrade, but doing so caused Detox tests to not run properly. After some internal discussion, decided that the headaches aren't worth it for now and we should just use JSC for the sake of reducing maintenance complexity

achou11:
New issue to arise: Intl support for relative time.

Example error:

```sh
[Error: [React Intl Error FORMAT_ERROR] Error formatting relative time. 
Locale: TypeError: undefined is not an object (evaluating 'RelativeTimeFormat.bind')
MessageID: undefined
Default Message: undefined
Description: undefined 

]
```

Caused by: https://github.com/digidem/mapeo-mobile/blob/edc7115573fc3b58d689ee15c932d285dbef1be6/src/frontend/sharedComponents/DateDistance.js

Affected components:
- PeerList
- FormattedData
  - ObservationListView, ObservationShare, ObservationView, QuestionLabel, ObservationEditView, ObservationListItem, GPSModal

Solution: Need to polyfill via https://formatjs.io/docs/polyfills/intl-relativetimeformat/

achou11:
for posterity, can't believe i didn't find this sooner but this is why detox wasn't working with the v8 version we tried to upgrade to:

https://github.com/Kudo/react-native-v8/pull/141

achou11:
Created a script to create the polyfill file for `@formatjs/intl-relativetimeformat` so i wouldn't have to write out the imports manually. not sure if it's entirely accurate so worth looking over

https://github.com/digidem/mapeo-mobile/pull/1040/commits/c4346c8c920ed1d5abb2b950fa8ce6028fd38bd9

https://github.com/digidem/mapeo-mobile/pull/1040/commits/0270bd08d2ba04acacd7f1b8868343b1149dd53e

Startup time is notably slower, and not sure about bundle size increase, but let's evaluate and discuss as needed

scripts/build-polyfills.js
```diff
@@ -0,0 +1,59 @@
...
+const fs = require("fs");
+const path = require("path");
+const {
+  shouldPolyfill,
+} = require("@formatjs/intl-relativetimeformat/should-polyfill");
+const languages = require("../src/frontend/languages.json");
+
+main();
+
+function main() {
```
achou11: figured this file could be used for other file generation related to polyfills, but happy to update so that it just focuses on the Intl stuff.

this doesn't need to be added to CI or anything, since we commit the output. just handy to have around in case something changes or we need to add other polyfills

src/frontend/polyfills/intl-relativetimeformat.ts
```diff
@@ -0,0 +1,125 @@
+import "@formatjs/intl-locale/polyfill";
```
achou11: worth checking out the outputs here to see if anything looks off. this is generated from the script I added. maybe I should add a comment to the top of this file 🤔

gmaclennan: We shouldn't check this in if it is generated. Generation should happen probably as an npm script on install maybe? Also the languages.json is a translation reference (translations of language names between english and native names). It is _not_ a list of languages that Mapeo supports. You can get supported languages from the `supportedLanguages` export of `frontend/context/IntlContext.js`.

achou11: > Also the languages.json is a translation reference (translations of language names between english and native names). It is not a list of languages that Mapeo supports

ahhh got it. that wasn't super clear to me but makes sense looking back at it. i'll update the implementation to mirror what's in that context

> We shouldn't check this in if it is generated. Generation should happen probably as an npm script on install maybe?

yeah i was thinking about but initially treated the list of languages to polyfill as static, which isn't really the case. happy to update to be based on an install hook

scripts/build-polyfills.js
achou11: renaming to `build-intl-polyfills` to scope it a little better

src/frontend/polyfills/intl-relativetimeformat.ts
gmaclennan: no it will change as we add new languages to crowdin.

scripts/build-intl-polyfills.js
```diff
@@ -0,0 +1,74 @@
...
+const path = require("path");
+const {
+  shouldPolyfill,
+} = require("@formatjs/intl-relativetimeformat/should-polyfill");
+
+const messages = require("../translations/messages.json");
+const languages = require("../src/frontend/languages.json");
+
+// For @formatjs/intl-relativetimeformat and @formatjs/intl-locale, which are needed for jsc-intl runtime
+buildIntlRelativeTimeFormat();
```
achou11, line 13: at the time of writing this, the output file looks like this:

```ts
// This file is automatically generated through scripts/build-intl-polyfills.js. Do not edit this directly!
// If you need to rebuild it, you can run `npm run build:intl-polyfills`
import "@formatjs/intl-locale/polyfill";
import "@formatjs/intl-relativetimeformat/polyfill";
import "@formatjs/intl-relativetimeformat/locale-data/de";
import "@formatjs/intl-relativetimeformat/locale-data/es";
import "@formatjs/intl-relativetimeformat/locale-data/fr";
import "@formatjs/intl-relativetimeformat/locale-data/id";
import "@formatjs/intl-relativetimeformat/locale-data/ja";
import "@formatjs/intl-relativetimeformat/locale-data/km";
import "@formatjs/intl-relativetimeformat/locale-data/my";
import "@formatjs/intl-relativetimeformat/locale-data/ne";
import "@formatjs/intl-relativetimeformat/locale-data/nl";
import "@formatjs/intl-relativetimeformat/locale-data/en";
import "@formatjs/intl-relativetimeformat/locale-data/pt";
import "@formatjs/intl-relativetimeformat/locale-data/si";
import "@formatjs/intl-relativetimeformat/locale-data/sw";
import "@formatjs/intl-relativetimeformat/locale-data/ta";
import "@formatjs/intl-relativetimeformat/locale-data/th";
import "@formatjs/intl-relativetimeformat/locale-data/vi";
```

package.json
```diff
@@ -35,15 +36,17 @@
     "storybook-native": "./node_modules/@storybook/react-native-server/bin/index.js -p 7007 -c ./storybook-native",
     "storybook-web": "./node_modules/@storybook/react/bin/index.js -c ./storybook-web -s ./public -p 5005 --ci",
     "upload-sourcemaps:release:android": "./scripts/upload-sourcemaps-release-android.sh",
-    "postinstall": "patch-package && jetify && cp ./node_modules/mapeo-offline-map/style.json ./android/app/src/main/assets/offline-style.json",
+    "postinstall": "patch-package && jetify && npm run build:intl-polyfills && cp ./node_modules/mapeo-offline-map/style.json ./android/app/src/main/assets/offline-style.json",
```
achou11: wondering if i also need to update the `build:translations` script to call `npm run build:intl-polyfills`, because otherwise on initial install, the polyfills script wouldn't be working with the appropriate data?

achou11: ah yeah just tried and because the `translations/` directory could be missing on initial install, my script doesn't work. maybe i'll move the call to this script to `build:translations` instead

achou11:
Not sure how accurate this is, but `index.android.bundle` size from last successful CI run here is `1.46MB` while v5.4.7 is `1.4MB` (diff of `56.83 KB`)

PR #1048, opened 2022-11-16
chore: improve maps management architecture

achou11:
This is the first part of a 3-part stacked PR based on the work from #1031 

This is mostly extracting @gmaclennan's work on refactoring the integration with the map server and how we use and represent state throughout the app.


src/frontend/api.ts
```diff
@@ -190,88 +197,43 @@ function createRequestClient({
...
 
+type MapServerClient = ReturnType<typeof createRequestClient> & {
+  port: number;
+};
+
 // TODO: Incorporate server status and making sure it's ready for requests?
 function createMapServerApi() {
-  let mapServerPort: number | undefined;
+  let deferred = pDefer<MapServerClient>();
   let client: ReturnType<typeof createRequestClient> | undefined;
```
ErikSin: Is `client` being used anywhere?

src/frontend/api.ts
```diff
@@ -190,88 +197,43 @@ function createRequestClient({
...
 
+  function getClient() {
+    return deferred.promise;
+  }
```
ErikSin, line 210: For my own knowledge/clarification. When `getClient()` is called, it returns a promise. The Promise is resolved when the event listener `map-server::start` is fired. And that promise resolves into an object, that has the API functions as its properties. Is that right?

ErikSin: If that is the case, should we worry about the error state? Aka, if there is a problem with the mapserver, `deferred.reject()` never gets called, so this function would infinitely hang

src/frontend/api.ts
```diff
@@ -281,8 +243,7 @@ function createMapServerApi() {
       );
 
       // Poke backend to send a state event
-      mapsApi
-        .ready()
+      ready()
         .then(() => nodejs.channel.post("map-server::get-state"))
         .catch(() => {});
```
ErikSin, line 247: If I understand correctly, When `MapServer` has called `createClient()` we are asking the backend to update the front end with the state. Is that because we have created the ports in the front end, but not yet initialized the map server in the backend?

src/frontend/hooks/useMapStyles.ts
```diff
@@ -0,0 +1,113 @@
...
+    }, [backgroundMaps])
+  );
+
+  // Keeping this in-scope so we can add translated name later
+  const defaultStyle = React.useMemo(() => {
+    return {
+      id: DEFAULT_MAP_ID,
+      url: onlineStyleURL,
+      bytesStored: 0,
+      name: "Default",
```
ErikSin: Why don't we just add this to crowdin now?

src/frontend/hooks/useMapStyles.ts
```diff
@@ -0,0 +1,113 @@
...
+    },
+    [setSettings]
+  );
+
+  let selectedStyleId = settings.mapStyleId;
+  // If the user selected style is not available, default to the first in the
+  // styles list.
+  if (!styles.find(style => style.id === selectedStyleId)) {
+    selectedStyleId = styles[0].id;
+  }
```
ErikSin, line 117: We should also probably throw an error, otherwise it could look like it is just not working

src/frontend/hooks/useSelectedMapStyle.ts
```diff
@@ -0,0 +1,63 @@
+import { useMapAvailability } from "./useMapAvailability";
+import { useMapStyles, DEFAULT_MAP_ID } from "./useMapStyles";
+
+// This map style is used if the user is offline and there is no custom offline
+// map installed. It contains basic land, country, river and lake data and is
+// generated by https://github.com/digidem/mapeo-offline-map
+const fallbackStyleURL = "asset://offline-style.json" as const;
```
ErikSin: Rename to `FALLBACK_STYLE_URL`

src/frontend/screens/MapScreen/BGMapSelector.tsx
```diff
@@ -157,56 +135,6 @@ export const BGMapSelector = React.forwardRef<
...
-      {title && (
-        <Text style={styles.thumbnailTitle} numberOfLines={1}>
-          {title}
-        </Text>
-      )}
-    </View>
-  );
-};
-
 const styles = StyleSheet.create({
```
ErikSin, line 138: Can remove `thumbnail` style

ErikSin (changes requested):

src/frontend/api.ts
achou11: ah this one is not. just forgot to remove. good catch 👍

src/frontend/api.ts
achou11: > For my own knowledge/clarification. When `getClient()` is called, it returns a promise. The Promise is resolved when the event listener `map-server::start` is fired. And that promise resolves into an object, that has the API functions as its properties. Is that right?

Yep!

> If that is the case, should we worry about the error state? Aka, if there is a problem with the mapserver, deferred.reject() never gets called, so this function would infinitely hang

We probably should. Not exactly sure where that would go. Maybe here (checking for the error `state`)?

```ts
  nodejs.channel.addListener("map-server::state", state => {
    if (state.value === "stopping") {
      deferred = pDefer<MapServerClient>();
    }
  });
```

src/frontend/api.ts
achou11: In this case, I believe it's to initiate/trigger the state listener that's created just above this line. The port is created on the backend and is sent via the `map-sever::start` event, which we listen for elsewhere in this file

src/frontend/hooks/useMapStyles.ts
achou11: l think it makes sense to keep this as is because the first item in `styles` is the default/fallback style and consumers of this hook may still be waiting for updates to that styles state from the server for some period of time, so we wouldn't just want to prematurely throw an error. that's what i gather from this, at least.

src/frontend/screens/MapScreen/BGMapSelector.tsx
achou11: the thumbnail style isn't defined here to begin with, weirdly 🤔

src/frontend/hooks/useSelectedMapStyle.ts
achou11: Addressed via https://github.com/digidem/mapeo-mobile/pull/1048/commits/0a2069ee25e76376468b4d8f4ed2c6cbe0266db5

src/frontend/hooks/useMapStyles.ts
achou11: addressed via https://github.com/digidem/mapeo-mobile/pull/1048/commits/5aa7bf884f1e46dcefc8ffabf90aa9eac7f3555e

src/frontend/api.ts
achou11: addressed via https://github.com/digidem/mapeo-mobile/pull/1048/commits/80f3b467f99c6f894c8bcdb60f0025aace49a4d3

src/frontend/screens/MapScreen/BGMapSelector.tsx
ErikSin: It is, its just not showing up in the diffs because you didn't touch it in your PR:

https://github.com/digidem/mapeo-mobile/blob/f2847cabb0b91beec53ee16438ce8ecae334c712/src/frontend/screens/MapScreen/BGMapSelector.tsx#L147

achou11: ah i was looking at the wrong component. are you saying that we don't need to apply the style? seems like we'd want it to guarantee the height?

ErikSin: yeah, its not being used, so we can just get rid of it. It looks like the style is now defined in the[ `mapThumbnail`](https://github.com/digidem/mapeo-mobile/blob/4a0f0f23a92abfa1e821ceee83206d1e61784f42/src/frontend/sharedComponents/MapThumbnail.tsx#L7)

achou11: either way, i update this component quite a bit in https://github.com/digidem/mapeo-mobile/pull/1050, so will let that handle any changes of that nature (it already removes that)

src/frontend/api.ts
ErikSin: @gmaclennan any opinions on this?

ErikSin:
One more thing, In the original PR, I had changed the titles of components and files from `BG` to `Background`, at your request. But it seems like I missed `BGMapSelector` and `BGMapCard`, if you want to change those files (or perhaps in the last of the stacked PRs so they don't just get re written)

achou11:
> One more thing, In the original PR, I had changed the titles of components and files from `BG` to `Background`, at your request. But it seems like I missed `BGMapSelector` and `BGMapCard`, if you want to change those files (or perhaps in the last of the stacked PRs so they don't just get re written)

ah yeah probably better to do that in one of the stacked PRs to avoid noise in this one. Thanks for pointing out 👍

src/frontend/api.ts
gmaclennan: Main opinion is: "this is hard". We're actually replicating a lot of quite complex code that already exists for managing the state of the Mapeo backend, this time using promises rather than event handlers, which in hindsight I'm not sure I like as much. What we are actually wanting to do here is mirror the state of the server on the client, which we kind of do in the promise. To do this properly I think we need a mapserver state in the frontend that replicates the server state. We _could_ use the promise for this, but it's kind of like an opaque state machine, especially as we replace it with a new promise on restart. We also need to be clear on the backend about the difference between actions that result in an error, and the server actually being in an error state that is not recoverable.

In summary, I think it would be good to clean this code up and improve it, but I think what we have is ok for now, and we can come back to this later, perhaps we can postpone to the Mapeo 6 refactor when we can improve how this backend-frontend integration is done.

src/frontend/api.ts
gmaclennan: Because the backend and frontend are starting in parallel, and the frontend could possibly reload without the backend reloading, we need to get the _current_ state of the backend when the frontend starts. We listen to the backend state via an event emitter, that emits a state event when the state changes, which means that the front end could be waiting a long time for a state event if the backend is already in a "ready" state. The workaround for this is to "poke" the backend with a "send me your current state" request that results in the backend emitting a state event even though the state has not changed.

I find we hit this need quite a lot with states via event emitters — how to get the current state as well as subscribe to the state change event. Would love to learn a neat pattern for doing this!

src/frontend/hooks/useMapStyles.ts
gmaclennan: The typing for `styles` ensures that it always contains at least one item in the array. You can rely on the map server either being in a "loading" state, or this to be loaded with at least one style available.

src/frontend/hooks/useMapStyles.ts
```diff
@@ -0,0 +1,129 @@
...
+        })
+        .catch(() => didCancel || setStatus("error"));
+
+      return () => {
+        didCancel = true;
+      };
+    }, [backgroundMaps])
+  );
+
+  const defaultStyle = React.useMemo(() => {
```
gmaclennan: `useMemo` isn't really needed here because these are not expensive calculations and we are not doing any equality tests on the resulting object (the `styles` array is a new object every render anyway).

src/frontend/hooks/useSelectedMapStyle.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  // We have a check in useMapStyles to ensure that selectedStyleId is in the
+  // styles array, but this is to make TS happy
+  const selectedStyle =
+    styles.find(style => style.id === selectedStyleId) || styles[0];
+
+  if (
+    onlineMapAvailability === "unavailable" &&
+    selectedStyle.id === DEFAULT_MAP_ID
```
gmaclennan, line 46: In the future we need something better here, since it assumes that the only "online" map is the default map. I have some ideas about how we might handle the fallback map in the mapserver rather than the client, it would be good to discuss at some point.

src/frontend/screens/Observation/ObservationView.js
```diff
@@ -86,14 +86,14 @@ const InsetMapView = React.memo<MapProps>(({ lon, lat }: MapProps) => {
...
+      styleURL={selectedMap.styleUrl}
     >
       <MapboxGL.Camera
         centerCoordinate={[lon, lat]}
-        zoomLevel={styleType === "fallback" ? 5 : 12}
+        zoomLevel={selectedMap.isOfflineFallback ? 5 : 12}
         animationMode="moveTo"
       />
-      {styleType === "fallback" ? <OfflineMapLayers /> : null}
+      {selectedMap.isOfflineFallback ? <OfflineMapLayers /> : null}{" "}
```
gmaclennan: What is the `{" "}` for? (curiosity)

gmaclennan (approved):
Looks good to me, no changes needed I don't think, just left some comments with some clarification and todos for later.

src/frontend/api.ts
achou11: > I find we hit this need quite a lot with states via event emitters — how to get the current state as well as subscribe to the state change event. Would love to learn a neat pattern for doing this!

In Rx land, a [BehaviorSubject](https://rxjs.dev/guide/subject#behaviorsubject) does this. I can imagine creating a custom event emitter class that does something similar, at least naively

gmaclennan: Yeah... Rx / observables is basically a solution to this, but personally I find the mental overhead of observables to be a lot, and while I might some day spend the time to understand them all, I worry the next person in the codebase will be in a similar situation.

src/frontend/screens/Observation/ObservationView.js
achou11: ah think it was mistakenly added while fumbling on my keyboard 😅 . removing

src/frontend/hooks/useMapStyles.ts
achou11: addressed via c25d4b8d5c4fb685cab2d3d81bb9bda9e9001ec9

src/frontend/screens/Observation/ObservationView.js
achou11: addressed via https://github.com/digidem/mapeo-mobile/pull/1048/commits/902826b8f93249dcac8bf15ef523c2d00fbb7d05

PR #1049, opened 2022-11-16
implement map import progress

achou11:
Addresses https://github.com/digidem/mapeo-mobile/issues/1017 and https://github.com/digidem/mapeo-mobile/issues/1018

~NOTE: this is stacked on top of #1048~ (1048 has been merged, so just points to develop now)

This is the better scoped version of the initial intention of #1031 , focusing solely on the map import progress logic in the map settings screen


src/frontend/hooks/useMapImportProgress.ts
```diff
@@ -0,0 +1,141 @@
...
+                dispatch({ type: "remove", styleId });
+              }
+            } catch (e) {
+              throw new FatalError("Unable to parse msg " + ev.data);
+            }
+          },
+          onerror(err) {
+            if (err instanceof FatalError) throw err;
+            // If not thrown, then fetch-event-source will retry
+          },
```
ErikSin, line 86: If an error get thrown by the server NOT on open, will `err` hit `validateResponse` function?

src/frontend/screens/Settings/MapSettings/BGMapCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+          <WithTopSeparation>
+            {showProgressBar && (
+              <Bar
+                indeterminate={importInfo.status === "idle"}
+                color={MAPEO_BLUE}
+                width={null}
+                progress={
+                  importInfo.status === "idle"
+                    ? undefined
+                    : // @ts-ignore - will come back to this!
```
ErikSin: For clarification, is this going to be addressed in the next stacked pr?

src/frontend/screens/Settings/MapSettings/BGMapCard.tsx
```diff
@@ -0,0 +1,169 @@
...
+}: BGMapCardProps) => {
+  const { navigate } = useNavigationFromRoot();
+  const { formatMessage: t } = useIntl();
+
+  const importInfo = useMapImportProgress(styleId);
+
+  const showBytesStored =
+    styleId !== DEFAULT_MAP_ID && styleId !== CUSTOM_MAP_ID && !isImporting;
+  const showProgressBar = importInfo && importInfo.status !== "error";
+  const showProgressBarMessage =
```
ErikSin, line 75: for clarity, perhaps call it `showIdleOrProgressMessage` since the message is only showing up for the 2 states

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
```diff
@@ -167,34 +165,38 @@ export const BackgroundMapInfo = ({
...
-            </Text>
-          )}
-
-          {id !== DEFAULT_MAP_ID && (
+        <View style={[styles.flex, styles.container]}>
+          <View>
+            {bytesStored ? (
+              <Text style={styles.detailsText}>
+                {`${bytesToMegabytes(bytesStored).toFixed(0)} ${t(m.mb)}`}
+              </Text>
```
ErikSin: Sorry, i think I added this complexity originally. We have `bytesStored` as a required field [here](https://github.com/digidem/mapeo-mobile/blob/7d50157f3bedfe76e2e945e6e88bad269f3e1e8c/src/frontend/screens/Settings/MapSettings/BGMapCard.tsx#L63). And then it is optional nav prop on this page. The problem is that for the places where we do not have bytesStored, I set it to 0...which is obviously not correct. So we should either make it optional everywhere, which would simplify [this logic](https://github.com/digidem/mapeo-mobile/blob/7d50157f3bedfe76e2e945e6e88bad269f3e1e8c/src/frontend/screens/Settings/MapSettings/BGMapCard.tsx#L72), OR make it required everywhere and check that it is not a `custom` or `default` on this line.

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
```diff
@@ -0,0 +1,322 @@
...
+  );
+};
+
+/**
+ * Bottom sheet content for when an imported file is processing, checking the
+ * format etc. is correct (shows until the import endpoint returns with the
+ * import id). User is unable to close this sheet.
+ */
+const ImportingContent = () => {
+  return <BottomSheetContent title="Processing file..." buttonConfigs={[]} />;
```
ErikSin: Translate title

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
```diff
@@ -0,0 +1,322 @@
...
+      // Set state on close so when it opens again this is what you see
+      setState("select");
+      bottomSheetRef.current?.close();
+    };
+
+    const handlePressImport = async () => {
+      setState("importing");
+      let results;
+      try {
+        results = await DocumentPicker.getDocumentAsync();
```
ErikSin, line 188: should we define the `{type}` here as `.mbtiles`?

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
```diff
@@ -0,0 +1,322 @@
...
+
+      try {
+        const {
+          import: { id: importId },
+          style,
+        } = await api.maps.importTileset(results.uri);
+
+        if (style) {
+          addMapImport({ styleId: style.id, importId });
+        }
```
ErikSin, line 207: Is there a case where there would be no style?

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
```diff
@@ -0,0 +1,322 @@
...
+          style={styles.bottomSheetView}
+        >
+          {bottomSheetContent}
+        </BottomSheetView>
+      </BottomSheet>
+    );
+  }
+);
+
+export default MapImportBottomSheet;
```
ErikSin: out of curiosity, why are you default exporting this?

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
```diff
@@ -0,0 +1,322 @@
...
+    const [state, setState] = React.useState<BottomSheetStates>("select");
+    const bottomSheetRef = React.useRef<BottomSheetMethods>(null);
+    const { snapPoints, updateSheetHeight } = useSnapPointsCalculator();
+    const { add: addMapImport } = useMapImportsManager();
+
+    React.useImperativeHandle(ref, () => {
+      return {
+        open: () => bottomSheetRef.current?.expand(),
+      };
+    });
```
ErikSin, line 176: Is there a reason you didn't opt to just use the `bottomSheetProvider` and associated hook?

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
```diff
@@ -0,0 +1,322 @@
...
+          {bottomSheetContent}
+        </BottomSheetView>
+      </BottomSheet>
+    );
+  }
+);
+
+export default MapImportBottomSheet;
+
+function useSnapPointsCalculator() {
```
ErikSin, line 266: Should we just use the `useSnapPointsCalculator` from `BottomSheetModal` in shared components?

ErikSin (commented):
Mostly clarification questions!

ErikSin:
Also, we should un-comment out [this line](https://github.com/digidem/mapeo-mobile/blob/7d50157f3bedfe76e2e945e6e88bad269f3e1e8c/src/frontend/screens/Settings/Experiments/index.tsx#L87) in the experiments, so user can turn on background maps

src/frontend/hooks/useMapImportProgress.ts
achou11: slightly confusing due to how the `fetch-event-source` api works, but `err` here comes from `validateResponse`. every response from the endpoint goes through `onopen: validateResponse` first, which then determines whether the `onmessage` or `onerror` handlers are called.

https://github.com/Azure/fetch-event-source/blob/6fb7115ebacb90600b54383ea32f799f6f647203/src/fetch.ts#L111

ErikSin:
And take out [this line](https://github.com/digidem/mapeo-mobile/blob/7d50157f3bedfe76e2e945e6e88bad269f3e1e8c/src/frontend/context/SettingsContext.tsx#L66) to enable bg maps

src/frontend/screens/Settings/MapSettings/BGMapCard.tsx
achou11: i don't have as much context (i think Gregor [worked on the file last](https://github.com/digidem/mapeo-mobile/blame/40e9ac3e8f6e335f7e099c14978f076f2e5d8e82/src/frontend/screens/Settings/MapSettings/BGMapCard.tsx#L112)), but no this isn't touched in the stacked PR. Can look into fixing although I _think_ it's mostly TS not being given enough info to exclude the case it's complaining about. will have to investigate a bit

src/frontend/screens/Settings/MapSettings/BGMapCard.tsx
achou11: ah yeah the naming was more referring to the entity rather than the state e.g. progress bar as a noun rather than a bar of progress type 😅 happy to update if that's still unclear

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
achou11: no specific reason, can update to named export

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
achou11: in practice, not currently. the TS typing has style as potentially null though so this is there to satisfy that. otherwise, i believe the imports we're dealing with will always generate a style

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
achou11: that provider is only applicable for the bottom sheet modal component, but this is using the normal bottom sheet component. reasoning for that is due to a really subtle bug when trying to use our custom bottom sheet modal implementation, and using the normal bottom sheet instead worked. not an ideal divergence but maybe something to revisit later rather than perfecting here

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
achou11: maybe? don't know off the top of my head if it's common to see mbtiles files that aren't suffixed with `.mbtiles`. since it's just a sqlite file i thought that there may be some really small edge case where it doesn't have the suffix when exported from elswhere. maybe better to just define the type like you're suggesting though

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
achou11: not sure if the proposed changes are necessary for this component because I think it makes sense to show the storage amount if it's a non-zero number, regardless of the type of style/map it is. it's possible that in the future, the custom or default map will have a definitive number (and not just explicitly passed `0`), which this will already handle.

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
achou11: for a similar reason as https://github.com/digidem/mapeo-mobile/pull/1049#discussion_r1066386207, there's a very slight difference in implementation due to using a bottom sheet and not a bottom sheet modal. again, not an ideal divergence but would rather have the separate implementations than try to make a unified one 😅

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
achou11: ah, `type` accepts a MIME type and unfortunately, there [doesn't seem to be a registered one for mbtiles](https://www.iana.org/assignments/media-types/media-types.xhtml). Using sqlite3's doesn't seem to work, so will just leave this as is. We have error handling code both server and client-side so at least running into file type errors are somewhat mitigated

src/frontend/screens/Settings/MapSettings/BGMapCard.tsx
achou11: yeah looks like it was a TS limitation (or maybe expected behavior?). 

I converted from this:

```ts
export type MapImportState =
  | { status: "idle" | "error" }
  | { status: "progress" | "complete" ; progress: number }
```

To this:

```ts
export type MapImportState =
  | { status: "idle" }
  | { status: "error" }
  | { status: "progress"; progress: number }
  | { status: "complete"; progress: number }
  | undefined;
```

and it seems to fix TS's complaint. Guess it has to do with TS's narrowing via discriminant unions. The checks that are done to reach this code didn't fully satisfy TS for it to know that if the `status` isn't `idle` here, it's `progress` or `complete`

src/frontend/screens/Settings/MapSettings/BGMapCard.tsx
achou11: actually, this is only used in one spot, so maybe just easier to remove the variable declaration and do this inline 😄

src/frontend/screens/Settings/MapSettings/BGMapCard.tsx
achou11: addressed via https://github.com/digidem/mapeo-mobile/pull/1049/commits/d660ce2c6d24b940c8da73391ca627caac013328

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
achou11: addressed via 07e2ff819527e12fea985c85bbae0ada1a2d3dc0

src/frontend/screens/Settings/MapSettings/MapImportBottomSheet.tsx
achou11: addressed via 2b74e59583da43afec5c681ecd733443d5e9bc64

achou11:
re-enabled the option to turn on background maps in settings via d919348c592d4f45e55ed52aeca06ed90334a754

src/frontend/screens/Settings/MapSettings/BGMapCard.tsx
ErikSin: Great, that would make it the clearest

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
ErikSin: That makes sense, but then we should make the `bytesStored` nav prop NOT optional to be consistent, and doing a check to see if it 0 (which i think technically this is doing...aka its a truth-y(?) check

ErikSin (approved):
Just need to address nav prop comment, but then can be merged

src/frontend/screens/Settings/MapSettings/BackgroundMapInfo.tsx
achou11: ah i see. i thought this was the BackgrdoundMaps screen, not the info screen for a background map. Updating based on your suggestion!

PR #1051, opened 2022-11-16
chore: remove background map

ErikSin:
# Removed background Map in Experiments

### DIFF
--- a/src/frontend/context/SettingsContext.tsx
+++ b/src/frontend/context/SettingsContext.tsx
@@ -54,6 +54,10 @@ export const SettingsProvider = ({ children }: React.PropsWithChildren<{}>) => {
     [setState]
   );
 
+  if (state.experiments.backgroundMaps) {
+    setSettings("experiments", { ...state.experiments, backgroundMaps: false });
+  }
+
   const contextValue: SettingsContextType = React.useMemo(() => {
     // If we add any new properties to the settings state, they will be
     // undefined in a users' persisted state, so we merge in the defaults
--- a/src/frontend/context/SettingsContext.tsx
+++ b/src/frontend/context/SettingsContext.tsx
@@ -54,14 +54,12 @@ export const SettingsProvider = ({ children }: React.PropsWithChildren<{}>) => {
     [setState]
   );
 
-  if (status !== "loading" && state.experiments.backgroundMaps) {
-    setSettings("experiments", { ...state.experiments, backgroundMaps: false });
-  }
-
   const contextValue: SettingsContextType = React.useMemo(() => {
     // If we add any new properties to the settings state, they will be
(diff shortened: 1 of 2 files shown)


src/frontend/context/SettingsContext.tsx
```diff
@@ -54,6 +54,10 @@ export const SettingsProvider = ({ children }: React.PropsWithChildren<{}>) => {
     [setState]
   );
 
+  if (status !== "loading" && state.experiments.backgroundMaps) {
+    setSettings("experiments", { ...state.experiments, backgroundMaps: false });
+  }
```
achou11: generally don't want to set state in the body of the render. instead this can be done when calculating `contextValue`:

```ts
  const contextValue: SettingsContextType = React.useMemo(() => {
    // If we add any new properties to the settings state, they will be
    // undefined in a users' persisted state, so we merge in the defaults
    const mergedState = merge({}, DEFAULT_SETTINGS, state, { experiments: { backgroundMaps: false }});
    return [mergedState, setSettings];
  }, [state, setSettings]);
```

may need to play around with ^. don't remember how `merge` works for nested objects

maybe also add a comment about why the last object is there and how it's temporary

achou11 (changes requested):

src/frontend/context/SettingsContext.tsx
achou11: i guess the goal with what you have is to actually forcibly write that into the persisted storage, but i don't think it's necessary in that case.

if you really wanted to do that, i'd put what you have in an effect anyways

ErikSin: yeah, that wouldnt write it into storage.

A use effect is not entirely necessary here. I following [this pattern](https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) to avoid a useEffect

ErikSin: Reading over your comment again, are you saying that it is not necessary to write it into storage?

achou11: > Reading over your comment again, are you saying that it is not necessary to write it into storage?

i don't think it is. my suggestion is basically the same as saying "we're ignoring whatever preference was saved", which i think is totally fine to do in this case because this is the only place that reads these settings from async storage directly

> A use effect is not entirely necessary here. I following [this pattern](https://beta.reactjs.org/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) to avoid a useEffect

that example is slightly different in that it's talking about updating local state based on _prop changes_, not changes in the local state. not sure if that recommendation also applies for the latter, but would avoid anyways since i don't think we need to call it for the above reason

achou11 (approved):

PR #1053, opened 2022-11-17
Release/v5.5.0

ErikSin:


ErikSin:
@reviewer: [c37adb4](https://github.com/digidem/mapeo-mobile/pull/1053/commits/c37adb458bc6751d7fdb017f511ae5d05206c8db) fixes passcode stack not being available in ICCA variant

ErikSin:
@Reviewer [c9ca0c1](https://github.com/digidem/mapeo-mobile/pull/1053/commits/c9ca0c1ca248fa41f1f9f25fc5676132630d4a67) fixes photo modal not correctly opening photo.

See [notion issue](https://www.notion.so/digidem/When-editing-saved-observations-Photo-view-screen-appears-all-black-and-includes-Delete-Image-butto-f655fe813d0e4e62a282dd37bf594a22) for reference

ErikSin:
@Reviewer [88ee1ee](https://github.com/digidem/mapeo-mobile/pull/1053/commits/88ee1eea5cdf497136132c5fa8d953561b441b1e) fixes auto zoom not working as intended

See [notion issue ](https://www.notion.so/digidem/Default-zoom-level-on-home-Map-screen-far-lower-than-standard-Mapeo-build-Intentional-31a84df6a17a4460b1215201d4cb7150)for reference

src/frontend/screens/AppPasscode/EnterPassToTurnOff.tsx
```diff
@@ -1,16 +1,16 @@
 import * as React from "react";
-import { defineMessage } from "react-intl";
+import { defineMessages } from "react-intl";
```
achou11, line 2: need to do the same thing for `LeaveProjectInitial.tsx`

ErikSin: OH, yes, its on my local device, I thought i pushed it with this

ErikSin: https://github.com/digidem/mapeo-mobile/pull/1053/commits/c8f9fc44cdfbf9ad9b949126a06a11cac8b5c8cf

ErikSin:
@Reviewer [bf1cf65](https://github.com/digidem/mapeo-mobile/pull/1053/commits/bf1cf65d9013c8afbd83afb12b29b4105b9da956) fixes observation detail going to the previous question when pressing back, instead of closing observation detail.

See [Notion issue](https://www.notion.so/digidem/button-in-Details-screens-returns-to-New-observation-or-Edit-observation-screen-rather-than-previou-0bc3421d2abb4430a386234d5895f1ca) for reference

ErikSin:
@Reviewer [9b0860b](https://github.com/digidem/mapeo-mobile/pull/1053/commits/9b0860bb33de58788a9d454d2a92b216a63367e8) fixes when completing the observation detail questions, while editing an observation, the edit view title appears as `edit observation` NOT `new observation`.

See [Notion issue](https://www.notion.so/digidem/When-editing-saved-observations-clicking-DONE-after-Details-screens-returns-to-New-observation-scre-c50a0dfc30ad449d9b5219b221249238) for reference.

src/frontend/screens/ObservationDetails/index.tsx
```diff
@@ -35,10 +37,24 @@ const ObservationDetails = ({
   route,
 }: NativeRootNavigationProps<"ObservationDetails">) => {
   const [{ preset }] = useDraftObservation();
-  const current: number = +route.params.question;
```
ErikSin: I believe this was a typo that made it into the code. Adding the current question number to itself does not make sense. the user is navigated to a NEW instance of this screen every time a new question appears, with the question param being updated.

achou11: i don't think this is a typo actually. it's a (unintentionally?) clever way of converting a potential string into a number 😅 e.g. `const foo = +"1"` results in `foo` being equal to `1`. probably not necessary given how we have react native navigation set up though

> Adding the current question number to itself does not make sense

That's not what this is doing. it would be written as `const current: number += route.params.question` if that were the case, which is invalid

ErikSin: ahh yes very clever. but yes, you are correct. With the new navigation we can pass all primitive types, and they will be extracted as the correct type

achou11 (approved):

PR #1059, opened 2022-11-30
fix: upgrade android target sdk to fulfill google play store requirements

### LINKED ISSUE #1058: Update Android Target SDK to meet Google Play requirement
Looks like as of Nov 2022, Google Play requires app updates to target Android 12 (i.e. SDK 31). **Google will block releases of apps that don't meet this requirement**.

The document outlines details about migration from 30 to 31 and React Native has already addressed this in [v0.68.0](https://github.com/facebook/react-native/blob/main/CHANGELOG.md#v0680). Not sure how reasonable it would be to upgrade all of React Native, but probably possible to just make a few file changes to include the new target sdk version.

https://developer.android.com/google/play/requirements/target-sdk

achou11:
Closes #1058

Notes:

- This is NOT an upgrade of React Native to a version that includes changes that fulfill the SDK requirements. This is an attempt to do the bare minimum to fulfill the requirement.
- The project now uses JDK 11 and not JDK 8. Instructions for how to install JDK 11 can be found in the **0.67** version of the React Native docs: https://reactnative.dev/docs/0.67/environment-setup
  - If you have multiple installations of JDK (check by running some variation of `/usr/libexec/java_home -V`), make sure you use 11 when working with this project.
  - I added a small change to the contributing docs to point to the React Native instructions for setting up the development environment
- Upgrading the target sdk caused an [issue with the Mapbox telemetry android module](https://github.com/rnmapbox/maps/issues/1677) when targeting 31 but using 8.5.0 of the mapbox-gl react native dep, sending me through a loop in order to find a fix for it. Basically, we have to manually specify the Android dep versions we want in order to properly fix that issue, which is a little hazardous. Most notably, we upgrade the Android SDK from 9.1.0 to 9.7.2, but based on the [changelog](https://github.com/mapbox/mapbox-gl-native-android/blob/main/CHANGELOG.md), it doesn't seem like there's anything too worrying (could be entirely wrong).
- i wrote a decently extensive comment about it, but we couldn't upgrade the `buildToolsVersion` to 31 for the following reasons:
  - non-blocking
    - requires upgrading to Gradle 7, gradle wrapper 7, and android gradle plugin 7
    - requires upgrading Bugsnag gradle plugin to 7 
  - blocking
    - upgrading to Gradle 7 requires upgrading every Expo module we rely on (see https://github.com/expo/expo/issues/12774). not a task I was ready to take on given the anticipated pre-requisite of moving away from unimodules
- overall, targeting 31 seems to require very minimal code changes. only things I had to update were a couple of android xml files.
- I was able to test this on my main phone (which runs Android 13) and my development phone (Android 11). both of them seemed to work fine for basic functionality.
- requires using [v2 of our docker image](https://github.com/digidem/docker-android/)

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -350,12 +350,6 @@ android {
         exclude 'META-INF/LICENSE.txt'
         exclude 'META-INF/NOTICE.txt'
     }
-
-    // Use Java 11, which is required when Android Target SDK is 31 (Android 12)
-    compileOptions {
-        sourceCompatibility JavaVersion.VERSION_11
-        targetCompatibility JavaVersion.VERSION_11
-    }
 }
 
 dependencies {
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -351,10 +351,10 @@ android {
         exclude 'META-INF/NOTICE.txt'
     }
 
-    // Use Java 1.8
+    // Use Java 11, which is required when Android Target SDK is 31 (Android 12)
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_11
+        targetCompatibility JavaVersion.VERSION_11
     }
 }
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -15,7 +15,7 @@ task:
     NODEJS_MOBILE_BUILD_NATIVE_MODULES: "0"
     MAPBOX_DOWNLOAD_TOKEN: ENCRYPTED[6c72ef68629d29a9a0d9aa396d1f55badda203ce2d2ce0e21c6f79d6714dc2966db4326efb73d0f6e2082cdf27943d34]
   container:
-    image: digidem/docker-android
+    image: digidem/docker-android:2
     cpu: 8
     memory: 24G
   node_modules_cache:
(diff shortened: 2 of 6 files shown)


android/app/build.gradle
```diff
@@ -350,12 +350,6 @@ android {
         exclude 'META-INF/LICENSE.txt'
         exclude 'META-INF/NOTICE.txt'
     }
-
-    // Use Java 1.8
-    compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
-    }
```
achou11, line 358: @gmaclennan do you remember the context for having this? I see that it was introduced in https://github.com/digidem/mapeo-mobile/commit/68a3136835836895190ad344d721a7f9b4d221bd but it causes the detox build command to break, and removing it entirely seems to have no other consequences.

achou11:
with https://github.com/digidem/mapeo-mobile/pull/1059/commits/41e2dfd142152ad6388c908d6ad1c427272003e7, can confirm that e2e tests work locally. only major issue is CI not working, which seems to be because the updated docker image isn't being pulled in by Cirrus 🤔

android/app/build.gradle
gmaclennan: This was the "recommendation" with the setup of expo-unimodules for the version we are currently using (v0.3.1): https://github.com/unimodules/react-native-unimodules/blob/v0.3.1/README.md?plain=1#L62

It is unclear what impact not having this will have, since the statement is just "we recommend", so if this works, then I think we are ok. You should check the camera works ok in the Release Candidate QA I think.

gmaclennan (approved):

PR #1081, opened 2023-02-12
chore: convert passcode screen to native nav

ErikSin:
# Converted passcode screens to utilize `React Navigation`

Passcode Screen were previously not using native navigation to go in between screen. This PR converts them to `Stack Nav Screens` using `React Navigation`

Also fixes a bottom modal flicker when user was navigated to turn off App Passcode.

Finally, takes away the next button in `EnterPassToTurnOff` Screen. Addresses [QA suggested improvement](https://www.notion.so/digidem/Tech-Priorities-a8bb127b4ba94322ae2a8cd75477e949?p=f53fa66c6a054fb4bb10c12d93ebd67f&pm=s)

### DIFF
--- a/src/frontend/screens/AppPasscode/InputPasscode.tsx
+++ b/src/frontend/screens/AppPasscode/InputPasscode.tsx
@@ -6,6 +6,7 @@ import {
 } from "react-intl";
 import { View, StyleSheet, Text, ScrollView } from "react-native";
 import { useBlurOnFulfill } from "react-native-confirmation-code-field";
+
 import { useNavigationFromRoot } from "../../hooks/useNavigationWithTypes";
 import { WHITE, RED, MAPEO_BLUE } from "../../lib/styles";
 import Button from "../../sharedComponents/Button";
@@ -30,11 +31,11 @@ interface InputPasscodeProps {
     subtitle: MessageDescriptor;
     errorMessage: MessageDescriptor;
   };
-  validate: () => void;
+  validate: (pass: string) => void;
   showPasscodeValues?: boolean;
   error: boolean;
-  inputValue: string;
-  setInputValue: (val: string) => void;
+  hideError: () => void;
+  showNext?: boolean;
 }
 
 export const InputPasscode = ({
@@ -42,9 +43,11 @@ export const InputPasscode = ({
   text,
   showPasscodeValues,
   error,
-  inputValue,
-  setInputValue,
+  hideError,
+  showNext = true,
 }: InputPasscodeProps) => {
+  const [inputValue, setInputValue] = React.useState("");
+
   const inputRef = useBlurOnFulfill({
     value: inputValue,
     cellCount: CELL_COUNT,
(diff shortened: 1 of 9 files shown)


src/frontend/screens/AppPasscode/TurnOffPasscode.tsx
```diff
@@ -122,24 +156,27 @@ export const TurnOffPasscode = ({ setScreenState }: TurnOffPasscodeProps) => {
...
+  ConfirmTurnOffPasswordModalProps
 >(({ turnOffPasscode, closeSheet }, sheetRef) => {
   const [snapPoints, setSnapPoints] = React.useState<(number | string)[]>([
     0,
     "40%",
   ]);
 
+  // This is crude, but it is suddenly opening and then closing if I don't do this...
+  React.useLayoutEffect(() => closeSheet(), [closeSheet]);
+
```
ErikSin: Without this, it is flickering open and closing right away when navigated to this screen. I originally thought it was because I wasn't using native navigation to navigate here, which lead me to add all these pages to native navigation....but that wasn't the problem (but we should probably use native navigation anyway....)

src/frontend/screens/AppPasscode/InputPasscode.tsx
```diff
@@ -84,18 +93,28 @@ export const InputPasscode = ({
...
-          <Button fullWidth onPress={validate}>
-            <Text style={[styles.buttonText, { color: WHITE }]}>
-              <FormattedMessage {...m.button} />
-            </Text>
-          </Button>
+          {showNext && (
+            <Button
+              fullWidth
+              onPress={() => {
+                console.log(inputValue);
```
achou11: remove

src/frontend/screens/AppPasscode/TurnOffPasscode.tsx
```diff
@@ -150,7 +184,7 @@ const ConfirmTurnOffPasswordModal = React.forwardRef<
       enableContentPanningGesture={false}
       enableHandlePanningGesture={false}
       handleHeight={0}
-      handleComponent={() => null}
+      index={-1}
```
achou11, line 188: out of curiosity, is this the change that addresses the flicker? just want that change to be highlighted for future reference :)

achou11 (approved):
one small removal needed but otherwise lgtm.

also, would point this to `develop` branch (see Slack)

src/frontend/screens/AppPasscode/TurnOffPasscode.tsx
ErikSin: I'm glad you asked...So I thought the flicker was caused by not using native navigation....which caused this whole refactor...only to find out that wasn't the cause. Upon some further research I found that setting the `index` to -1 stops it from opening initially. Also, the `handleComponent` doesn't make a difference. I tested it out and forgot to put it back, so Ill re-add it.

The refactor should have been done anyway, but to stop the flickering, it was unnecessary.

PR #1082, opened 2023-02-15
Es/fix back map details

ErikSin:
Fixes back button in map details


src/frontend/sharedComponents/BottomSheetModal/index.tsx
```diff
@@ -95,7 +98,7 @@ const useSnapPointsCalculator = () => {
 
 interface Props extends React.PropsWithChildren<{}> {
   onDismiss: () => void;
-  onHardwareBackPress?: () => void;
+  onHardwareBackPress?: () => void | boolean;
```
achou11, line 101: think this would need to be adjusted to this:

```suggestion
  onHardwareBackPress?: (() => void) | boolean;
```

generally would avoid making this a union of a boolean though and use a separate argument, as done in https://github.com/digidem/mapeo-mobile/pull/1083

achou11 (changes requested):

PR #1093, opened 2023-03-06
chore: Remove REQUEST_INSTALL_PACKAGE permission

ErikSin:
Removed REQUEST_INSTALL_PACKAGE permission

### DIFF
--- a/KNOWN_ISSUES.md
+++ b/KNOWN_ISSUES.md
@@ -41,3 +41,4 @@
 #### Misc
 
 - Switching quickly between two different installations of Mapeo on the same phone can cause an error. [#434](https://github.com/digidem/mapeo-mobile/issues/434)
+- Use of the p2p updater will not work for v5.6.0 and above if the app has been downloaded from the Google Play Store, as per [Google's policy](https://support.google.com/googleplay/android-developer/answer/12085295). Users will be able to update older version of Mapeo, up to v5.6.0, on other phones. But they will not be able to recieve an update themselves.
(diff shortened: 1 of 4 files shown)


achou11 (approved):
Just to confirm, we're okay with this feature not working for the non-Google Play builds for now? If so, we should create an issue to conditionally apply the permission based on build flavor, as something to work on in the future.

Relevant resources:
- https://stackoverflow.com/questions/28478110/android-studio-two-flavors-with-different-manifest-files/28480064#28480064
- https://developer.android.com/studio/build/manage-manifests#merge_rule_markers

achou11:
Are there any documentation changes in this repo that need to be made as a result of this change?

achou11:
also for extra context when looking back on this: Google Play isn't allowing us to publish a release with the permission in question due to policy stuff

ErikSin:
> Just to confirm, we're okay with this feature not working for the non-Google Play builds for now? If so, we should create an issue to conditionally apply the permission based on build flavor, as something to work on in the future.
> 
> Relevant resources:
> 
>     * https://stackoverflow.com/questions/28478110/android-studio-two-flavors-with-different-manifest-files/28480064#28480064
> 
>     * https://developer.android.com/studio/build/manage-manifests#merge_rule_markers

I think we should. It does feel like quite a lot of work, so I don't want to do it now as we are mid sprint and we haven't decided if it is our priority. Ill mark it as something to consider next sprint and Ximena and I will discuss :)

ErikSin:
> Are there any documentation changes in this repo that need to be made as a result of this change?

Yeah, actually ill wait to merge this until the programs team has approved of the copy that we are going to change in the app. I will also add it to `known problems` in our readme.

PR #1098, opened 2023-05-22
fix: fix sharing of observation on Android 12+

### LINKED ISSUE #1097: sharing observation containing attachment does not work on Android 12+
pressing the `Share` button on the observation screen does not do anything when the observation contains an attachment. in development, doing so surfaces the following error message:

```
Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.",
```

see https://github.com/react-native-share/react-native-share/issues/1073 for reporting of issue in `react-native-share`

achou11:
Fixes #1097 

Notes:

- Upgrades `react-native-share` to latest (`8.2.2`). Fix was introduced in [`7.0.1`](https://github.com/react-native-share/react-native-share/releases/tag/v7.0.1).
- Slight uncertainty about the permissions change noted here: https://react-native-share.github.io/react-native-share/docs/migrate-v4-to-v5#removal-of-write_external_storage-permission-request. Not sure if it applies to us in this case (we're sharing file uris e.g. `file:///data/user/0/com.mapeo.debug/files/media/preview/8a/8a6e62b150eae104f2cd59c78aac8948.jpg`). from general testing, doesn't seem like the permission is needed to keep this working cc @gmaclennan 
- Steps to verify fix (must use Android 12+ device):

1. Create observation that has an attachment (e.g. photo)
2. Open observation screen
3. Press `Share` button

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39338,9 +39338,9 @@
       }
     },
     "react-native-share": {
-      "version": "3.7.0",
-      "resolved": "https://registry.npmjs.org/react-native-share/-/react-native-share-3.7.0.tgz",
-      "integrity": "sha512-MCTGiPcBfelrnX/7n82N5eQAl31F6xljqFK2CWbPDTSt7bbJUZH3Znvy8KYePLYqAr7HVrk621eanQN6vSFMzA=="
+      "version": "8.2.2",
+      "resolved": "https://registry.npmjs.org/react-native-share/-/react-native-share-8.2.2.tgz",
+      "integrity": "sha512-kVCI/cT0GnuYUTXe6mAimrjrnt4VWoRfrWqJZjFeoYFqAyOEfos84RC4eZlZnOT5eVtmTXRIkor5vgSkKOlZhw=="
     },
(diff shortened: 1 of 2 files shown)


achou11:
~noticing that the preview isn't showing up in the share menu. can't remember if that's expected or not but guessing it's not because it shows up in the case where there's no attachments present.~

Screen Shot 2023-05-22 at 4 41 23 PM

~seeing the same thing as https://github.com/react-native-share/react-native-share/issues/613 in adb logs - wondering if that's related. not really sure how to fix, but this doesn't seem to actually break the sharing itself, so hopefully not a major blocker for this PR~

UPDATE: this is not a regression as this is the previously observed behavior for observations without attachments, so please ignore this comment 😄

achou11:
Closing in favor of #1099

PR #1099, opened 2023-05-25
Hotfix/v5.6.2

achou11:
Sharing observations containing attachments failed to work on Android 12+ due to an outdated dependency (react-native-share). This commit upgrades the dependency to its latest available version. See https://github.com/digidem/mapeo-mobile/pull/1098 for more details about the change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [5.6.2](https://github.com/digidem/mapeo-mobile/compare/v5.6.1...v5.6.2) (2023-05-25)
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


gmaclennan:
> Slight uncertainty about the permissions change noted here: https://react-native-share.github.io/react-native-share/docs/migrate-v4-to-v5#removal-of-write_external_storage-permission-request. Not sure if it applies to us in this case (we're sharing file uris e.g. file:///data/user/0/com.mapeo.debug/files/media/preview/8a/8a6e62b150eae104f2cd59c78aac8948.jpg). from general testing, doesn't seem like the permission is needed to keep this working cc @gmaclennan

This should not be needed. It is only required when sharing files as base64 strings, since as I understand it react-native-share writes the file to a temporary cache folder in order to share it. Since we share files directly, and they are in internal storage, not external, this permission should not be needed. I've read through the other changes in react-native-share and checked the installation instructions to make sure everything is up-to-date in Mapeo, and it looks good to me. Hopefully this will work without issue. Let's publish and then hotfix again if necessary in case edge cases on certain devices are detected.

gmaclennan (approved):

PR #1104, opened 2023-06-01
[DO NOT MERGE] feat: run node 16 on device

### LINKED ISSUE #1100: upgrade nodejs-mobile-react-native to v16.17.10
https://github.com/nodejs-mobile/nodejs-mobile-react-native

achou11:
Closes #1100 

Notes:

- there's a dev experience regression described in https://github.com/nodejs-mobile/nodejs-mobile-react-native/issues/45 that _should_ be fixable, but not entirely sure what the fix is just yet
  - we do have the option to patch the dep to remove the steps in question, since they don't apply to us right now
- in order to upgrade, we need to drop the `x86` architecture entirely (for now). `nodejs-mobile` v16.17.0 dropped support for this, although it may reintroduce it later. according to our download statistics, not many devices using mapeo have this architecture (~7 active installations as of this week) but this is a notable change that's being introduced, so should be approached carefully.
  - https://github.com/nodejs-mobile/nodejs-mobile-react-native/issues/34#issuecomment-1418941732
  - https://github.com/nodejs-mobile/nodejs-mobile/blob/main/doc_mobile/CHANGELOG.md#2023-01-26-version-16170
  - there's a small amount of app code that would be affected by this (p2p app updater). **do we need to update the app code to account for this change?** (given that we had to disable p2p app updater for now, maybe it's okay not to?)
- this PR also enforces upgrading the dev node version to v16.17. however, by default installing that version comes with npm v8, which does not work well with `nodejs-mobile-react-native` and our project due to issues introduced in npm v7+. basically, this updated nmrn version attempts to workaround the npm issue but it specifically doesn't work for `better-sqlite3` because of the way that project is structured (additional context [here](https://github.com/digidem/react-native-with-mapeo-core-next/issues/2#issuecomment-1567445144)). as a result, we upgrade to node 16 but developers need to explicitly use npm v6 for now.
  - https://github.com/npm/run-script/issues/23
  - https://github.com/digidem/react-native-with-mapeo-core-next/issues/2
- **This upgrade requires upgrading to NDK 24.0.8215888.**
- ~Due to the NDK version being upgraded, we need to update the docker image we use for Cirrus CI. work-in-progress PR is here: https://github.com/digidem/docker-android/pull/3. Ideally we'd update the docker image and make sure this PR successfully runs on Cirrus before merging it~ UPDATE: done and updated this PR - looks like it worked!


achou11:
looks like backend tests are failing due to upgrading Node version from 12 to 16. seems to be related to different streams behavior.

1. for one of the `beforeAfterStream` tests, this test is failing:

https://github.com/digidem/mapeo-mobile/blob/95beddfef3e71ca3522f1fd2b6b87dcd53678e3d/src/backend/test/utils/beforeAfterStream.test.js#L344

seems like the main reason is the result of this line:

https://github.com/digidem/mapeo-mobile/blob/95beddfef3e71ca3522f1fd2b6b87dcd53678e3d/src/backend/upgrade-manager/utils.js#L361

In node 12, `ok` is `true` but in node 16, `ok` is `false`. seems like 16 actually has the behavior I would expect for this test's case.

---

2. a couple of the `upgrade-storage` tests are failing:

https://github.com/digidem/mapeo-mobile/blob/95beddfef3e71ca3522f1fd2b6b87dcd53678e3d/src/backend/test/upgrade-storage.test.js#L193

https://github.com/digidem/mapeo-mobile/blob/95beddfef3e71ca3522f1fd2b6b87dcd53678e3d/src/backend/test/upgrade-storage.test.js#L261

I'm noticing that this listener in `beforeAfterStream` is not being triggered as expected in node 16, causing the tests to fail I guess:

https://github.com/digidem/mapeo-mobile/blob/95beddfef3e71ca3522f1fd2b6b87dcd53678e3d/src/backend/upgrade-manager/utils.js#L341

---

In both cases, I'm able to add a couple of lines to fix them, but they're very uninformed and while all of the backend tests pass as a result (locally), I still lack confidence.

Created a branch with these "fixes": https://github.com/digidem/mapeo-mobile/compare/1100/upgrade-nodejs-mobile-rn...ac/node-16-upgrade-backend-tests

cc @gmaclennan

achou11:
not going to merge this but keeping it as a draft for the foreseeable future. refer to https://github.com/digidem/mapeo-mobile/issues/1100#issuecomment-1593522398 for more details

PR #1148, opened 2024-03-08
Es/mapbox token

ErikSin:
Created script to generate env variables (to be used by js front end).

updated bitrise.yml to run script in order to access secret key in bitrise


achou11 (commented):
Think using something like https://github.com/goatandsheep/react-native-dotenv would notably simplify the implementation as it adheres to the [Twelve Factor App](https://12factor.net/) approach and also allows usage of shell environment variables (useful for CI like Github + Bitrise).

https://www.npmjs.com/package/react-native-config could also work, but not sure if it allows usage of shell environment variables at build time. not familiar with how to set up an env file for Github Actions and Bitrise but I know both support defining secrets that get injected as shell variables that can be used.

ErikSin:
@achou11 I actually have tried this library (and a few other config libraries). None of them work with our current archtecture. Dotenv, doesnt work with our version of babel, and upgrading babel, lead to a whole slew of other issues. I also tried [react-native-config](https://github.com/lugg/react-native-config) and that also didn't work.

After playing around with these libraries for several hours, I think the having a prebuild script is the path of least resistance

achou11:
> @achou11 I actually have tried this library (and a few other config libraries). None of them work with our current archtecture. Dotenv, doesnt work with our version of babel, and upgrading babel, lead to a whole slew of other issues. I also tried [react-native-config](https://github.com/lugg/react-native-config) and that also didn't work.
> 
> After playing around with these libraries for several hours, I think the having a prebuild script is the path of least resistance

ah damn - unfortunate but yeah I guess our react native version is quite behind at this point :/

achou11:
not sure if more convenient but another idea: we already make use of `transform-inline-environment-variables` (added to support feature flags) so maybe that can be used for this purpose too? although thinking about the various changes needed to get that working, that may be even harder to maintain 😅

achou11:
@ErikSin was able to get react-native-config working - can push up a separate branch if you're interested in the changes

ErikSin:
> @ErikSin was able to get react-native-config working - can push up a separate branch if you're interested in the changes

yesss please!!

PR #1150, opened 2024-03-11
fix: adjust mapbox access token usage

achou11:
- Uses `react-native-config` to read a `.env` file containing the token of interest
- Updated instruction in `contributing.md` to reflect use of env file
- Updated bitrise yml to create env file using the mapbox access token which is saved as a secrete key
- Updated Cirrus yml with encrypted mapbox token, and a script to create an env file with the encrypted token

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing
 
-Thank you so much for taking the time to contribute to Mapeo development! Mapeo Mobile is a mobile app written in react-native that currently runs on Android. 
+Thank you so much for taking the time to contribute to Mapeo development! Mapeo Mobile is a mobile app written in react-native that currently runs on Android.
 
 After doing the [Initial Install](#initial-install), for most development you will need Android Studio and the Android SDK installed on your computer as outlined in the [Full App Development](#full-app-development) section.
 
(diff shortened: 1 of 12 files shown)


android/app/build.gradle
```diff
@@ -227,7 +230,11 @@ android {
 
         // Bugsnag integration
         manifestPlaceholders = [bugsnagReleaseStage: getDefaultBugsnagReleaseStage(myVersionName)]
+
+        // React Native Config
+        resValue "string", "build_config_package", "com.mapeo"
```
achou11, line 235: following the step described here: https://github.com/lugg/react-native-config?tab=readme-ov-file#advanced-android-setup

Hoping I read it correctly and did the right thing, but not entirely sure tbh. Figured it was a relevant step because we do specify different suffixes depending on the build and command

ErikSin: this step i skipped when I tried this library earlier. So this must be necesssary...

ErikSin:
I did a manual build in bitrise to test the use of the secret key and everything seems to be working: https://app.bitrise.io/build/5ee6fbee-1cb2-4924-9eaa-8b06619035c4?tab=artifacts

ErikSin:
@gmaclennan Im currently using a token from my personal account. Were you thinking that we just use a free token from the Digidem account, or did you have something else in mind?

ErikSin:
Updated the token (to the digidem account token) in bitrise, and the reference to the encrypted token in the cirrus.yml

ErikSin (approved):

gmaclennan:
FYI `process.env` doesn't really exist in the react native runtime

achou11:
> FYI `process.env` doesn't really exist in the react native runtime

this was done at my suggestion to make it slightly easier for local development by taking advantage of our usage of https://babeljs.io/docs/babel-plugin-transform-inline-environment-variables (being used for feature flagging). can revert it and instead update the docs to say that an `env` file is now required if you think that's a better way to go

gmaclennan:
Ah great, didn't realise we used that in comapeo. We could have skipped adding react-native-config 😀

ErikSin:
> Ah great, didn't realise we used that in comapeo. We could have skipped adding react-native-config 😀

I don't think its working for .env files. It seems like it is only picking up shell env variables. So react-native-config is still needed. There was some work we could have done to update babel and make it work, but that lead to a whole slew of other updating issues

achou11:
> Ah great, didn't realise we used that in comapeo. We could have skipped adding react-native-config 😀

yeah unfortunately it would be a little inconvenient to only rely on this for local development as you have to specify/export the environment variable _in the shell_ for the command that eventually runs `react-native bundle`, so we'd still probably need some kind of env file to work with anyways, unless we hardcoded the access token in a few places 😄

achou11:
decided to revert the change introducing `process.env` usage  and just make it required to set up a `.env` file for local development. i updated the contributing docs to be a little clearer about what's needed for that part

CONTRIBUTING.md
```diff
@@ -23,28 +23,23 @@ npm install
...
 
-Due to an [issue](https://github.com/rnmapbox/maps/issues/1572) with installing some Mapbox SDK Android deps, you will also have to complete additional steps before getting the app to build:
+The `MAPBOX_DOWNLOAD_TOKEN` variable is necessary for installing components of Mapbox's native SDK for building the app (see this [issue](https://github.com/rnmapbox/maps/issues/1572) for more context). To set it up:
 
 1. Refer to the [`Configure Credentials`](https://docs.mapbox.com/android/maps/guides/install/#configure-credentials) section and follow the instructions for creating a **secret access token**. This requires creating a [Mapbox](https://mapbox.com) account.
 
-2. With the secret token, you'll need to define the `MAPBOX_DOWNLOAD_TOKEN` environment variable. Although not ideal, there are a couple of options to choose from:
+2. With the secret token, update your `.env` file so that it has line similar to the following: `MAPBOX_DOWNLOAD_TOKEN=<your_token_here>`
```
ErikSin, line 36: Will this work? Its currently being used in an android file. But android files have access to env variables?

achou11: yep! we define a helper in the gradle scripts to read the `.env` file already:

https://github.com/digidem/mapeo-mobile/blob/48b75344d89d6af5ce84a1ce09520d6902d28089/android/build.gradle#L4-L29

ErikSin: Great!

